Issue #107, opened 2019-08-07
Setting active reflections (rings) in materials pane


joelvbernier:
we would like to be able to have the user select the active rings they want from the displayed list by two methods:

1. Using the mouse (ctrl and shft click)
2. By specifying a maximum bragg angle
3. By specifying a minimum d-spacing

The PlaneData class has an "exclusions" attr that is a boolean mask to select a subset of the complete list of generated reflections (as specified currently by the max sum of h^2 + k^2 + l^2).  There is also a tThMax attr that is initialized as None, but if set to a float, it treats it as a cutoff on 2theta.  I can code up a dspMin attr to handle the 3rd case.  Setting PlaneData.exclusions = None resets to display all.

This functionality was in the old wx GUI under the "edit HKLs" button.

psavery:
As of #236, the materials now default to selecting the first 5 rings when they are modified. When a material is added, it has the same exclusions as the previously selected material (from which it was copied).

With regard to my comment above, #236 now also makes it so that the rings that are selected in the materials panel are the ones that are used during calibration.

PR #140, opened 2019-08-12
Add ring selection limits to materials panel

psavery:
This allows a user to specify either a max Bragg angle
or a minimum d spacing to be used for ring selection. Any rings
that are higher than the max Bragg angle or lower than the min
d-spacing will not be displayed in the GUI.

It still does not yet update exclusions to be used for the powder
calibration. Please let me know if that is desired.

Addresses part of #107

See the image below for an example. The two outer-most rings are not displayed because they are higher than the max Bragg angle and lower than the min d-spacing.
image

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -70,20 +70,23 @@ class InstrumentViewer:
         rbnds = []
         rbnd_indices = []
 
-        selected_rings = HexrdConfig().selected_rings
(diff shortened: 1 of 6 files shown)


cryos:
@joelvbernier we should look at this before starting the new ring work you mentioned - this is draft because you don't think it is complete @psavery?

psavery:
@cryos I can't fully remember unfortunately. I'll merge master into this branch and play around with it a little.

Issue #209, opened 2019-10-31
User picked calibration option


joelvbernier:
Ok -- I have some skeleton code for a simple, fast powder calibration mode based on user-picked points.  Here we go:

`fig` and `ax` are the polar view plot

First my line-based picker class:
```
class LineBuilder:
    def __init__(self, line):
        self.line = line
        self.xs = list(line.get_xdata())
        self.ys = list(line.get_ydata())
        self.cid = line.figure.canvas.mpl_connect('button_press_event', self)

    def __call__(self, event):
        """
        Picker callback
        ??? maybe handle button 1 and 3 differently?  1 - build, 3 - finish
        """"
        print('%s click: button=%d, x=%d, y=%d, xdata=%f, ydata=%f' %
          ('double' if event.dblclick else 'single', event.button,
           event.x, event.y, event.xdata, event.ydata))
        if event.inaxes!=self.line.axes: return
        self.xs.append(event.xdata)
        self.ys.append(event.ydata)
        self.line.set_data(self.xs, self.ys)
        self.line.figure.canvas.draw()
```
Now for the code that needs to be called when we start the picker process, *e.g.* from a button signal

```
# list for set of rings 'picked'
ring_data = []

# fire up the cursor for this tool
cursor = Cursor(ax, useblit=True, color='red', linewidth=1)
```
and this is what adding a ring looks like:
```
# =============================================================================
# the following call stack should be attached to a button slot for "add ring" 
# =============================================================================
# as we want to append the selected points to ring_data
lspec_str = 'c' + '.'    # !!! cycle through colors for the line object
line, = ax.plot([], [], lspec_str)  # empty line
linebuilder = LineBuilder(line)
plt.show()

# User clicks points on a line...
# upon "finish" (need button, or perhaps catch a right click?), append to ring_data
ring_data.append(np.vstack([linebuilder.xs, linebuilder.ys]).T)
```
for example, first line:
![Screen Shot 2019-10-31 at 10 42 28 AM](https://user-images.githubusercontent.com/1154130/67972243-86cc9c80-fbcb-11e9-9c25-78e1fbe39e69.png)

second line:
![Screen Shot 2019-10-31 at 10 43 19 AM](https://user-images.githubusercontent.com/1154130/67972334-b1b6f080-fbcb-11e9-97a6-55966bb346a9.png)

third line:
[...]

Issue #195, opened 2019-10-22
Add ability to draw mask to zero out bright areas


cjh1:
this would be in the polar view.

PR #210, opened 2019-11-05
Add calibration line picker for polar view

psavery:
If the user chooses "Run"->"Calibration Line Picker" (which is
only enabled in polar mode), they enter a calibration line picking
mode, where they can left-click to add points to a line, and 
right-click to start a new line. 

While the user is picking points, a dialog is shown that includes
a zoomed-in portion of the image. The user can modify the zoom
box width and height as well.

Once the user is done picking points, they click "Ok" on the dialog 
box, and it sends the data they picked to the main window for a
calibration.

It currently does not do any calibrating, but just printing the data.

If it is a little slow, we might be able to increase the speed by saving
the AxesImage and the Line2D in the zoom canvas, rather than
clearing it every time.

image

Let me know what you think, @joelvbernier 

Fixes: #209 
Fixes: #195


psavery:
So I think this PR is actually in pretty good shape. It is functional, and you can calibrate multiple detectors at the same time. 

The only thing I might would change is that the mouse movements are kind of slow when the calibration line picker is up. This is for two reasons I think: 1. The ROI zoom and intensity plots need to update with every mouse movement, and 2. The main canvas has to be re-rendered with every mouse movement because of the zoom box that is being drawn.

However, speeding up the mouse movements for the calibration line picker is something that doesn't necessarily need to be done on this PR, either.

I think if someone else tries this out and approves, I think it's ready for merging.

PR #218, opened 2020-01-21
Update conda dev instructions

cjh1:


cjh1:
@cryos Can you try this out when you get a chance?

Issue #219, opened 2020-01-22
Detector selector doesn't update upon instrument config (re)load


joelvbernier:
The detector id list dosn't refresh upon loading a new instrument config.  Also, I'm confused about the usage; does it auto-detect the detector association using pattern matching (like the raw image load), or do you select each imageseries file for each detector?
![Screenshot 2020-01-21 23 11 28](https://user-images.githubusercontent.com/1154130/72873162-f373a680-3ca3-11ea-9896-a8932ce90bfc.png)

joelvbernier:
update:  it does refresh upon quitting and restarting the app, just not when loading a new instrument config file.

bnmajor:
#220 addresses the first part of this problem. When a new configuration file is uploaded, the detector names are updated in the load panel and the file table is reset.

Should the currently loaded images (if any) be cleared as well to avoid confusion? Also the table is always cleared after a new configuration file is loaded - is this correct, or should this only happen if the name and/or number of detectors change with the new config?

bnmajor:
To address usage: 
To select an `imageseries` where all files are located in one directory you only need to select one of each of the files and it will use pattern matching to find the equivalent files for each remaining detector. When using sub-directories for each detector, you will select all of the needed files for a single detector and again, pattern matching will be used to look in the remaining sub-directories to find the equivalent files for the rest of the detectors. In either case the appropriate parent directory must be set using the `Change Image Folder` dialog.

Going over this again there were assumptions in the code made about the order of detectors in order to associate them with the correct files. Changes made in #221 make sure that associations between file and detector are made using the actual detector name instead of assuming they will always be listed in any particular order.

cjh1:
Lets confirm this is fixed

bnmajor:
Fixed in #220. Tested again and the detector list is updated on new configuration load.

PR #220, opened 2020-01-23
Clear load panel when a new configuration is loaded

bnmajor:
If a new configuration file is loaded detectors are updated and file table is reset.


bnmajor:
Addresses part of #219

cryos:
This has conflicts we need to resolve @bnmajor

bnmajor:
Conflicts have been resolved

PR #221, opened 2020-01-23
Associate selected files with detectors using name

bnmajor:
Previous logic assumed detectors were sorted in ascending order in order to
associate selected files with the appropriate detectors. Files are now matched
to their detector using the detector name in the file name or subdirectory. This
assures that any detector name in any order will always have the correct files
associated with it.


bnmajor:
Addresses part of #219

Issue #92, opened 2019-08-02
Overlay manager


joelvbernier:
Can we promote the panel for the ring overlays to a separate manager ("overlay manager") that will provide the ability to add overlays from multiple materials?  Looking forward, there will be 2 additional overlay types I can envision using:

- single crystal Laue patterns; and 
- single crystal rotation series.

in addition to the polycrysal ("powder") rings.  The single crystal overlay types will require a material _plus_ a calibration crystal spec.  They will be point markers instead of lines.  Laue patterns are 2-d for single images like the powder patterns with a set of (x, y) detector coordinates specified for each image, while the rotation series overlays are 3-d with (x, y, omega) coordinates, but could be displayed on a omega-projected imageseries (_e.g._ max over all frames)

It would be ideal to be able to build an overlay set from combinations of these three components, which would likely require hooks to line/marker spec for the user to differentiate them.

psavery:
@joelvbernier Hi Joel. For some clarification here, do we want to keep the materials options inside the panel, or move it somewhere else?

I was thinking that if we keep it in the panel, we could perhaps add another table above the hkl table (see below), and the new table would contain the list of materials. The user could add, modify, or delete materials. The hkl table and space group info would be populated with whichever material is currently selected. What do you think about this?

![image](https://user-images.githubusercontent.com/9558430/72812968-81ae4500-3c30-11ea-86f5-18a6df833971.png)

psavery:
I think another potentially simpler option that I might go with is to just add a checkbox at the bottom that says `Show All Materials`, and when it is checked, it will show overlays of all the materials in hexrd. The user will be able to edit the styles of each material individually. And, in the future, we will add options for the different materials to have a type (options for types would be something like `powder`, `Laue`, and `rotation series`).

joelvbernier:
The current master is a big step in the right direction.  After using it, one simple enhancement  would be to add a button to "hide all" instead of having to thumb through the material list to toggle each "visible" checkbox.  The next step is to add the hooks for different types of overlays.  Currently we only have powder rings, but I would also like to plot Laue patterns; this requires both a material selection, then the 12 additional crystal parameters.  Code snippet to follow...

psavery:
@joelvbernier Okay. I kept around the "Calibration Crystal Editor" in a pop-up dialog (it used to be a part of the instrument config). It's currently under "Edit"->"Calibration Crystal", but it doesn't actually do anything. I think we can hopefully re-use this dialog.

[...]

PR #222, opened 2020-01-24
Allow multiple materials to be displayed simultaneously

psavery:
This adds a checkbox to the materials panel called "Visible", which can be used to show/hide each individual material.

Additionally, a button was added to the materials panel called "Edit Style...". When clicked, this pops up a dialog box that allows the user to edit the overlay style for the current material. The line color, line style, and line width can be edited, both for the "rings" and for the "ranges".

Updated materials panel:
===================
image

The overlay style editor:
==================
image

One thing that might be confusing is that "Show Rings", "Show Ranges", and the range spinbox are all global settings for the whole program. But everything else in the panel are local settings for the individual material. Maybe we should make those global settings into local settings as well, and the "Visible" checkbox will be combined with the "Show Rings" checkbox?

What do you think, @joelvbernier?

This addresses part of #92

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -17,7 +17,6 @@ from .display_plane import DisplayPlane
 
 def cartesian_viewer():
     images_dict = HexrdConfig().current_images_dict()
-    plane_data = HexrdConfig().active_material.planeData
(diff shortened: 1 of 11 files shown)


psavery:
@cryos Merge conflict fixed. I did a quick test, and it *looks* like it is working okay. I'll merge it and test a little more after...

Issue #223, opened 2020-01-27
Load data interface doesn't allow for different detector-specific options


darrencpagan:
In general the image transforms and dark mode options for each detector can be different.  The current interface doesn't allow for this.

joelvbernier:
PS -- I submitted this logged in as Darren.

cryos:
Is there a real example we can work on where there are different detectors? It will complicate the code, and we will have to look at where we made these assumptions but I am sure we can work to generalize things.

cryos:
This is working now.

PR #224, opened 2020-01-29
Allow transform and dark mode to be set per detector

bnmajor:
This adds a checkbox that allows the user to choose whether to apply the selected
transformation and dark mode to all detectors, or to set them per detector. The
default state is per detector.

When selecting per detector, each detector must be changed by selecting the
detector from the drop down menu and then updating the tranformation and/or dark
mode. If left unchanged, that detector will receive the default value. Switching
between detectors will update values accordingly and make it clear which values
will be used for that detector when the files are read in.

When 'Apply Selections to All Detectors' has been selected, the current settings
will be used for all detectors, regardless of which detector is currently
visible. Switching between detectors will display the same transformation and
dark mode selections for all.

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -75,6 +75,9 @@ class HexrdConfig(QObject, metaclass=Singleton):
     """
     update_status_bar = Signal(str)
 
+    """Emitted when a new instrument configuration file has been loaded"""
+    instrument_config_loaded = Signal()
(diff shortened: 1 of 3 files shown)


bnmajor:
The state is remembered even when `Apply Selections to All Detectors` has been selected, so toggling the switch off again will reload the previous selections.

This does not currently allow the aggregation to be set per detector.

Addresses #223.

Issue #225, opened 2020-01-29
Current index in the series should be reset to zero on file load


cryos:
There may be other places too, but the index is retained. I.e. we load a series, look at image 20, then load a file containing two frames, and we get a failure shown in the console. We should be resetting to 0 and setting the range to 0. We probably need to assure the application is in a reasonable state when loading a new file as we don't know how many images are in the sequence until they are loaded.

cryos:
This is now resolved.

PR #226, opened 2020-01-30
Reset index to zero when new files are loaded

bnmajor:
Reset the current index in the series to zero when new files are loaded.


bnmajor:
Addresses #225

PR #233, opened 2020-01-31
Use hexrdgui as the main entry point

cryos:
The main hexrd package is going to use hexrd as their entry point, so we
will use hexrdgui as our entry point. Updated the URL, name for
consistency.


cjh1:
If we are changing the name of the executable we should do it in the packages as well. For example here: https://github.com/HEXRD/hexrdgui/blob/master/packaging/package.py#L39

cjh1:
We will need to rename: https://github.com/HEXRD/hexrdgui/blob/master/packaging/darwin/hexrd and https://github.com/HEXRD/hexrdgui/blob/master/packaging/linux/hexrd

Issue #235, opened 2020-01-31
Update PlaneData's tThMax with materials panel limits


psavery:
@joelvbernier Do we want the max Bragg angle and the min d spacing options in the materials panel to update the tThMax for all materials, or just the currently selected material?

psavery:
I set it up so that each material has it's own tThMax. Let me know if I should change it.

PR #236, opened 2020-01-31
Sync up materials' plane data with the materials panel options

psavery:
Previously, the materials panel only stored its configuration internally to hexrdgui, and it did not change options on the materials' plane data.

This update intends to change that, so that modifying things in the materials panel will produce correct calibrations.

So far, this PR only syncs up the maxtTh, which is why it is a draft. One more commit is intended, which will modify materials' exclusions when the selections in the materials panel change.

Fixes: #107
Fixes: #235
Fixes: #237

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -6,7 +6,6 @@ from hexrd import instrument
 from hexrd.gridutil import cellIndices
 
 from hexrd.ui.hexrd_config import HexrdConfig
-from hexrd.ui.utils import select_merged_rings
 
 from skimage import transform as tf
 from skimage.exposure import equalize_adapthist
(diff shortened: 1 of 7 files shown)


psavery:
My latest commit message:

Rather than having a separate internal variable for the ring selections,
use the material's plane data exclusions instead.

This means that selecting rows now modifies the exclusions, and when
materials are loaded, rows are programmatically selected to match
the exclusions. The row selections are also now preserved when
changing materials. Thus, the row selections *should* always match
the exclusions of the currently selected material.

Modifying the exclusions also makes it so that, when a calibration is
performed, the rings that are used in the calibration are the selected
ones (the other rings get ignored).

Now, nearly all of the individual materials' properties on the materials
panel can be saved and loaded in the "materials.hexrd", including ring
selections, and the max bragg angle. The only properties that are not
being saved in the "materials.hexrd" file are:

1. Whether the material is visible or not
2. The overlay styles of the material
3. The range width (this is a global property for all materials
   currently)

Most of these are saved in the local hexrd config file in between runs.

psavery:
I think this is working and is ready for review.

I tested lots of things. But there are also lots of things to test, and I may have missed something. But it seems to be working in my tests.

psavery:
As a side note: calibration currently only uses the currently selected material in the materials panel (and it only uses the selected rings in the table). In the future (but not yet), we will allow multiple materials to be used in the calibration.

Issue #163, opened 2019-08-28
Image borders for cartesian and polar views


joelvbernier:
Need to export detector boundary separately for these views with a toggle for display.  Requires additional output from rendering code.

cryos:
This is working, I want @joelvbernier to verify it works as he wants before closing the issue out.

psavery:
Yeah, so I think this is indeed working. However, it currently works by editing the image, rather than drawing an extra plot on top of the image (as is done for the overlays). This makes it so that the azimuthal integration in the polar plot is affected by the borders. I'm wondering if we should get it to start drawing an extra plot on top of the image instead of modifying the image.

cryos:
Yes, that is what must happen at this point as we start to rely on that plot more. Let's prioritize this

PR #247, opened 2020-02-05
Plot detector borders on top of the image

psavery:
Previously, if "View" -> "Show Detector Borders" was checked, the
cartesian and polar images would be modified, with the detector borders
drawn in as the max pixel values. This meant that the borders would
affect the azimuthal integration plot in the polar plot.

Rather than drawing the detector borders into the images, plot them as
separate lines on top of the images. This way, the images and the
azimuthal integration plot are not affected by their presence.

This seems to work well, although drawing the detector borders for the
polar plot is time consuming. We are using
[hexrd.instrument.PlanarDetector.pixel_angles()](https://github.com/HEXRD/hexrd/blob/79ebd31e43ca4b96a7cad60ff61ff2cc4fb14fba/hexrd/instrument.py#L1636)
for this, and it appears to be slowing down the generation of the polar
plot by a lot. It would be good to figure out a way to get the polar
detector borders faster. The `pixel_angles()` function generates and
returns a lot of values we don't need (we only extract the first and
last rows/columns from the output). Maybe we can make a new function
that only generates the detector border values.

The drawing of the Cartesian borders seems to be fast, and it remains
very interactive with the slider widgets.

Some images of the new detector borders in action are here:
cartesian
polar

Fixes: #163

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -43,6 +43,7 @@ class InstrumentViewer:
         self.images_dict = images_dict
         self.pixel_size = pixel_size
         self.warp_dict = {}
+        self.detector_corners = {}
 
         dist = HexrdConfig().cartesian_virtual_plane_distance
         dplane_tvec = np.array([0., 0., -dist])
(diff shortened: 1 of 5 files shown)


psavery:
Polar border drawing is now much faster as of the latest commit.

joelvbernier:
Just verifying: the borders are just an overlay on the image axes and don't contribute to the lineout sum of the image, correct?

psavery:
@joelvbernier Yep, that should be right. The borders are now overlays just like the materials overlays.

joelvbernier:
Awesome!  Probably good to merge at this point then.

joelvbernier:
One thing -- there is a future enhancement to this functionality, but it isn't relevant until we have masks working.  This is the case where the "active" detective area is a region inside the rectangular boundary.  We would need to find the border using a shrink wrap of the mask array (or the like).  Scipy seems to have this functionality.  Perhaps you've done something similar in previous work?

Issue #216, opened 2020-01-14
Open Imageseries does not recognize all imageseries input format


joelvbernier:
Need to add adaptors for "image-files" yaml spec and "frame-cache" (i.e. .npz) imageseries in addition to HDF5 format.  Refer to Don's [wiki](https://github.com/donald-e-boyce/hexrd/wiki/ims-load-options) for usage

cryos:
Can you confirm that this works @bnmajor? Thanks.

bnmajor:
At this point the load panel seems to handle all file types for me. I was able to load `hdf5`, `yaml`, `tiff`, and `npz` images. I did also document the load panel in more detail in the original PR #156 in case that is useful. 

Currently the `APS Imageseries` option for opening files allows the user to select a directory and then all images within that directory are used to create an `image-files` type image series, discussed in #112. The load panel does currently offer this functionality, although it could certainly be added in.

cryos:
> The load panel does currently offer this functionality, although it could certainly be added in.

I think you mean does not here

bnmajor:
@cryos You are correct, thank you for catching that.

joelvbernier:
yes, confirmed.  I do think, however, that the APS ImageSeries directory munging import should somehow be worked into the directory selection ("change image folder") button on the load dialog.

bnmajor:
To the best of my knowledge there is no way to edit the actual file dialog to request directory data munging, but this could be handled with a 'Use APS Imageseries' (or differently worded) checkbox that would disable the `Select Image Files`. This could use existing logic to look for sibling directories like `Use Subdirectories` does, assuming each directory would have the name of the relevant detector. Alternatively, the dialog could be automatically re-opened to manually select the directory for each detector, as is the case now.

bnmajor:
@joelvbernier I think I've got this working, just need to know if selecting directories should be manual for each detector, follow the sub-directories pattern of assuming the directories are siblings and named for their respective detector, or both (manually select unless `Use Subdirectories` is checked, then assume location and directory names).

joelvbernier:
I think we can use the detector names and forget about having to specify manually.  Thanks for this!

bnmajor:
Perfect! The changes for this are available in #248.

cryos:
I think we are done here, closing. Please reopen if any further issues remain.

PR #248, opened 2020-02-07
Load aps imageseries

bnmajor:
Simply moves the existing `APS ImageSeries` functionality from the menu to the load panel. To use, select the `Load APS ImageSeries` checkbox, select the directory to use, then select `Select Image Files`. The files from that directory will be displayed in the table. 

This currently follows the sub-directories pattern: directories are expected to be siblings with names that match the detector names. Only one directory is selected, the rest are inferred.


bnmajor:
Addresses #216

Issue #244, opened 2020-02-03
Modify number of images when loading a new config


psavery:
Currently, if the user tries to load a new config while there are already images loaded, and the number of detectors in the new config doesn't match the number of images already loaded, an error is produced.

Rather than producing an error, we should modify the number of images so that it matches. Perhaps just removing all images will also be a good solution (since, the mismatch in the number of detectors and the number of images most likely means the config isn't for those images).

bnmajor:
#256 removes any loaded images when a new configuration file is loaded. Currently this happens for every new config file. Should this only happen when the new config has a different number of detectors and/or different detector names than the currently loaded config file?

psavery:
I think it might be good to check the detector names before/after to see if they match (so something like `prev_detectors = HexrdConfig().get_detector_names()`, then after the load, `new_detectors = HexrdConfig().get_detector_names()`, then check):

```python
    if prev_detectors != new_detectors:
        # Clear the images
```

Just because someone might be comparing different configs of the same detector, and this would allow them to not have to re-load the images each time. But if that's not a likely scenario, then maybe it doesn't matter.

bnmajor:
I've added in a check to see if the new detectors differ from the old before clearing the images. Additionally, the loaded images are now also cleared if a detector is manually added or removed using the `Instrument Calibration` panel, under the `Form View` tab.

psavery:
Fixed by #256
#260 Will also make it so dummy images are loaded when a new config is loaded.

PR #256, opened 2020-02-17
Clear images when a new configuration is loaded

bnmajor:
When a new configuration file is loaded, remove any loaded images and reset the
load panel editing options.

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -309,6 +309,11 @@ class HexrdConfig(QObject, metaclass=Singleton):
         hexrd.imageseries.save.write(ims, write_file, selected_format,
                                      **kwargs)
 
+    def clear_images(self):
+        self.imageseries_dict = {}
+        self.hdf5_path = None
+        self.load_panel_state = {}
+
     def load_instrument_config(self, yml_file):
         with open(yml_file, 'r') as f:
             self.config['instrument'] = yaml.load(f, Loader=yaml.FullLoader)
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -315,6 +315,7 @@ class HexrdConfig(QObject, metaclass=Singleton):
         self.load_panel_state = {}
 
     def load_instrument_config(self, yml_file):
+        old_detectors = self.get_detector_names()
         with open(yml_file, 'r') as f:
             self.config['instrument'] = yaml.load(f, Loader=yaml.FullLoader)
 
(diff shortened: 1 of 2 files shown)


bnmajor:
Addresses #244

hexrd/ui/hexrd_config.py
```diff
@@ -328,7 +334,11 @@ def load_instrument_config(self, yml_file):
         self.backup_instrument_config()
 
         self.update_visible_material_energies()
-        self.instrument_config_loaded.emit()
+
+        new_detectors = self.get_detector_names()
+        if old_detectors != new_detectors:
+            self.instrument_config_loaded.emit()
```
psavery: I think that if we are going to use the signal like this, we should probably just change it to call `self.detectors_changed.emit()` instead. We can then probably remove the `instrument_config_loaded` signal from hexrdgui, since it's only emitted here. 

I also see [here](https://github.com/HEXRD/hexrdgui/pull/256/files#diff-73b9467134adf696570c35cf87a32d14R92), `HexrdConfig().detectors_changed` is now connected to `config_changed`, so we can probably then also remove [the line after](https://github.com/HEXRD/hexrdgui/pull/256/files#diff-73b9467134adf696570c35cf87a32d14R93).

hexrd/ui/load_panel.py
```diff
@@ -147,8 +147,9 @@ def config_changed(self):
         self.ui.file_options.setRowCount(0)
         self.reset_data()
         self.enable_read()
-        HexrdConfig().load_panel_state = {}
-        self.state = self.setup_processing_options()
+        HexrdConfig().clear_images()
+        self.setup_gui()
+        self.parent().image_tab_widget.load_images()
```
psavery: I *think* that the main window will already be updating the images [here](https://github.com/HEXRD/hexrdgui/blob/378bc1fe62feb8220f2360e2bd673c513b30a280/hexrd/ui/main_window.py#L171) (which should get called after this block). So this might not be needed. I'm going to be making sure the dummy images get loaded on new config in #260 as well.

hexrd/ui/hexrd_config.py
bnmajor: This makes sense, no need for redundant signals. Fixed to use the `detectors_changed` signal instead.

hexrd/ui/load_panel.py
bnmajor: Removed.

hexrd/ui/hexrd_config.py
psavery: Thanks Brianna!

hexrd/ui/load_panel.py
psavery: Thanks Brianna!

PR #258, opened 2020-02-17
Convert borders to degrees in a better location

psavery:
For the dexela instrument, I would be getting this error when
trying to draw the polar borders:

```
AttributeError: 'numpy.ndarray' object has no attribute 'degrees'
```

I'm not sure what was causing the error, but it was happening at
`np.degrees(borders)` where borders was a list of tuples of
np.ndarray objects.

It seems that the error does not occur if we convert to degrees
earlier, so let's just do that instead.


deboyce-consult:
I was curious about this, so I looked into it. 

If you apply np.degrees to a tuple of arrays, the arrays have to be the same size, or else you get that message. The help says the inputs have to be "array-like", so I am guessing that if they are not the same shape, it looks for a degrees attribute as a last ditch effort.

joelvbernier:
Thanks Don!

Issue #240, opened 2020-02-03
Display dummy images when no images are loaded


psavery:
These dummy images will have shapes matching the specification in the current instrument config.

PR #260, opened 2020-02-17
Display dummy images when no images are loaded

psavery:
When there are no images loaded, which happens either at the
start of the program or when changing to a different instrument
config, show dummy images, which are just arrays that have the
shape defined in the instrument config and all values of 1.

These dummy images get replaced with the real images when a user
loads real images.

Fixes: #240 
Fixes: #261

raw_images
cartesian
polar

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -89,8 +89,12 @@ class MainWindow(QObject):
 
         ImageFileManager().load_dummy_images()
 
-        # The dummy images will be drawn in MainWindow.show()
-        # See MainWindow.show() for more details.
+        # In order to avoid both a not very nice looking black window,
+        # and a bug with the tabbed view
+        # (see https://github.com/HEXRD/hexrdgui/issues/261),
+        # do not draw the images before the first paint event has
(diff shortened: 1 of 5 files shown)


hexrd/ui/main_window.py
```diff
@@ -165,10 +168,18 @@ def on_action_open_config_triggered(self):
...
-            self.update_all(clear_canvases=True)
+            self.update_config_gui()
+
+            new_detectors = HexrdConfig().get_detector_names()
+            if new_detectors != prev_detectors:
+                # Load the dummy images. The new config probably isn't
+                # for the current images.
+                self.load_dummy_images()
+            else:
+                self.update_all(clear_canvases=True)
```
psavery: This section overlaps a little bit with the updates in #256, since we also want to load new dummy images when the config changes.

I'll wait until #256 is merged, then I'll re-work this, and then this PR will be ready to go.

psavery: It is merged, and this is hopefully resolved.

psavery: Resolved.

Issue #237, opened 2020-02-03
Use plane data exclusions for ring selections


psavery:
Have the ring selections in the materials panel use and be synced up with the plane data exclusions. This makes it so that ring selection also affects the rings used in calibration.

psavery:
To do: make sure the tth ranges get set on the material's plane data.

psavery:
@joelvbernier @cryos 
So, I just realized from the code, that the plane data's `tThWidth` does get set, but all of the materials' `tThWidth`'s are forced to be the same, and they are set based upon what is in the "Show Ranges" spinbox (0.125 degrees by default).

This also means that the loaded materials' tthwidth do not get preserved, but get forced to be 0.125 degrees.

Maybe we want each material to have it's own `tThWidth`, and use that spin box to set the `tThWidth` for each material, rather than forcing all materials to have the same?

psavery:
I allowed each material to have it's own `tThWidth` in #262. I think this should fix any issues we were facing.

joelvbernier:
There is also the minor issue in the GUI that the number is not ±, but the full width.  We should update the formatting in the spin controller (or have it display half the tThWidth value with the ±).

When instantiating an material object, the tThWidth is set to None; perhaps having a default value of ±0.125° is reasonable.

psavery:
@joelvbernier  I modified this in #262, which is now merged.

Here's what it looks like now: 
![Screenshot from 2020-02-21 15-44-19](https://user-images.githubusercontent.com/9558430/75070367-0e168680-54c1-11ea-8ca0-1efe6a5677b3.png)

So I removed the ± and renamed it to "width".

PR #262, opened 2020-02-18
Allow each material its own tThWidth

psavery:
Rather than forcing all materials to have the same tThWidth,
as was done previously, allow each material to have their own
tThWidths. The "Show Ranges" checkbox was renamed to "Enable Width",
and this toggles setting `None` on the material's tThWidth. The
spinbox displays the tThWidth on the material. And, when the active
material in the materials panel is changed, the checkbox and spinbox
update to the newly selected material.

This also changes a little naming in the materials panel, including
renaming "Show Rings" to a more general "Show Overlays".

Additionally, the plus or minus prefix on the width was removed.

Finally, this also enables saving/loading of the `tThWidth`s via the `materials.hexrd` file.

Fixes: #237

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -71,16 +71,15 @@ class InstrumentViewer:
         rbnd_indices = []
 
         # If there are no rings, there is nothing to do
-        if len(plane_data.getTTh()) == 0:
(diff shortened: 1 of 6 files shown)


psavery:
This PR changes some of the looks and naming of the options in the materials panel.
Screenshot from 2020-02-18 11-10-29

With these changes, now nearly all of the settings are specific to the currently selected material. The only global setting is `Show Overlays`. I think this will make things less confusing.

Issue #265, opened 2020-02-20
Image aren't removed when a detector is deleted


cjh1:
if you remove a detector the image associated with the detector remains.

PR #267, opened 2020-02-21
Update images when detectors are added/removed

psavery:
This causes the dummy images to be updated when detectors
are removed/added. It also adds a new signal in the hexrd
config that causes a deep re-render to occur.

Currently, there are many things that can cause a deep re-render
to occur, but most of them occur due to internal checks in the
ImageCanvas class (such as "_polar_reset_needed()"). This adds
the ability for other parts of the code to force a deep re-render
as well. It is needed, for instance, if the user switches to a
config that has the same number of detectors (because the ImageCanvas
class will not automatically perform a deep re-render in this case).

Fixes: #265

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -70,6 +70,12 @@ class HexrdConfig(QObject, metaclass=Singleton):
     """
     rerender_needed = Signal()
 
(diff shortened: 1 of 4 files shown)


psavery:
I double-checked the changes to the load panel with @bnmajor, and they are fine.

Issue #241, opened 2020-02-03
Export polar image data to HDF5 file


psavery:
Add an option to export the numpy array data of the polar images to an HDF5 file. We can also write out the axis values (intensity and 2-theta) to the file.

joelvbernier:
all of these attributes are hanging off of the PolarView class.

psavery:
@joelvbernier Actually, I just realized that "Export"->"Polar Plot" exports the polar image data as npz, with all of these keys:
```
tth_coordinates
eta_coordinates
intensities
extent
```

Do we want to modify this menu item to export HDF5 instead of npz? Or, make it so that this menu item can export to both NPZ and HDF5? Or just leave the NPZ exporter and skip the HDF5 export?

joelvbernier:
Maybe add the HDF5 output as an option -- for generality's sake?

I have a little recursive function for unwrapping a dict (of dicts) to HDF5 if that is useful to you.

psavery:
@joelvbernier Yeah I think adding an HDF5 option sounds good. 

I think that one option we might be able to do is to have one menu item "Export->Polar Plot", but if the user names the file with a ".h5" extension, write it out as HDF5, and for a ".npz" extension, write it out as npz.

PR #268, opened 2020-02-21
Add ability to export polar plot to HDF5

psavery:
When the polar plot is being displayed, the user can export the
polar plot data via "File"->"Export"->"Polar Plot". Now, the user may
choose between HDF5 (default) or npz for the export. If the file
name ends with ".npz", a ".npz" file will be written. Otherwise,
an HDF5 file will be written.

The data that is written out is as follows:
```
tth_coordinates
eta_coordinates
intensities
extent
```

They are all written out as datasets in the root in the HDF5 file.

Currently, no material overlays are written out and no borders are
written out. But that is potentially changeable.

Fixes: #241

### DIFF
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -1,3 +1,5 @@
+import h5py
+import os
 import numpy as np
 
 from hexrd import instrument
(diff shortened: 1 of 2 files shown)


joelvbernier:
Thanks @psavery, looks like this should be ready to merge up.

PR #273, opened 2020-02-25
Release updates

cjh1:


joelvbernier:
I will mention that we will eventually include the entry point for the GUI under the hexrd console script (this will be a hexrd 0.8 milestone) to be called as `hexrd gui`.  In the meantime, it is good to have the unique executable to allow for testing.

PR #276, opened 2020-02-28
Slider on config change

### LINKED ISSUE #275: Sliders incorrect on tab view or config change
The slider tool for scrolling through the frame of an ImageSeries is not being created/displayed properly when in tab view, or when a new configuration with different detectors is loaded. Switching into tab view (or loading images in tab view) displays the slider widget for all detectors, rather than just the current detector. Switching back out of the tab view keeps all of the slider widgets visible.

double_slider
double_slider_2

Additionally, loading a new configuration file with different detector names displays no sliders for any detector.
no_slider

bnmajor:
Corrects toolbar visibility when switching to and from tabbed view and loading new configuration files.

This fixes #275


cjh1:
So, I have play with this a little and it looks like it fixes the problems described in the issue. I think there may still be some issues with the behavior. It seems that the image is not reset to the slider value or the slider is not reset ( depending on the desired behavior ) in the following sequence, see the following screencast. 
Screencast-from-03-02-2020-02_02_42-PM

bnmajor:
This seems to be working correctly now - the appropriate slider is displayed for the current view/tab, and switching between tabbed/un-tabbed view no longer resets the current image.

Issue #281, opened 2020-03-11
Load Data parsing failure


joelvbernier:
The following filenames give a load error:

```
mruby-0129_0004-ff1_0_0.npz
mruby-0129_0004-ff1_0_1.npz
mruby-0129_0004-ff1_1_0.npz
mruby-0129_0004-ff1_1_1.npz
mruby-0129_0004-ff2_0_0.npz
mruby-0129_0004-ff2_0_1.npz
mruby-0129_0004-ff2_1_0.npz
mruby-0129_0004-ff2_1_1.npz
```
for the detector keys:
```
ff1_0_0
ff1_0_1
...
ff2_1_1
``` 
This is for the current master.

joelvbernier:
Looks like we should in general just be able to use the `find()` builtin method for strings, no?

```
imsname = 'mruby-0129_0004.ff1_0_1.npz'

imsname.find('ff1_0_1')
Out[102]: 16

imsname.find('ff2_0_1')
Out[103]: -1
```

bnmajor:
Closing for now as this seems to be fixed with #282, we can always re-open if there are any issues.

PR #282, opened 2020-03-11
Match files to detectors with special characters

bnmajor:
Do not make assumptions about detector names, names may contain special
characters as well as letters or numbers.

### DIFF
--- a/hexrd/ui/load_panel.py
+++ b/hexrd/ui/load_panel.py
@@ -241,14 +241,6 @@ class LoadPanel(QObject):
         for img in selected_files:
             f = os.path.split(img)[1]
             name = os.path.splitext(f)[0]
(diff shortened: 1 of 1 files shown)


bnmajor:
Tested with and without sub-directories, this works for the existing detectors and data that I had (`ge1`, `ge2`, `ff1`, `ff2`), as well as with detectors named with special characters (`ff1_0_0`, `ff_0_1`). Correctly differentiates between similarly named files (will select `mruby-0129_0004-ff1_0_0.npz` for detector `ff_0_0`, but not `mruby-0129_0004-ff1_0_0_0.npz`).

Fixes #281

joelvbernier:
Has this one been tested?  Need this funcitonality ASAP.

Issue #283, opened 2020-03-11
Loaded images not updated if in cartesian view


joelvbernier:
I just encountered this problem: if you have loaded an imageseries (from the "File➜Open➜Images" dialog), if you try to load another while in the cartesian view, it does not update.  If you toggle between raw and cartesian, however, it does.  I would suspect the same bug is also present for the polar view.

One of the obvious signs something was off is that I went form imageseries to single tiff files for an 8-panel instrument, and the frame slider did not disappear until I toggled to the raw view.

psavery:
I will take a look.

bnmajor:
@joelvbernier The option to load images will be removed from the file menu eventually, but when I followed the same steps using the load panel I did still find a bug. Loading a new `ImageSeries` updates the images (and the slider), but it displays the `Raw` images while the `Image Mode` tab does not update (it remains in `Cartestian` or `Polar` mode, whichever was last selected).

Would the preferred behavior be to simply update the `Image Mode` tab to be `Raw`, or to update the view to reflect the currently selected `Image Mode`?

joelvbernier:
FWIW I went from imageseries to single tiffs per detector.

joelvbernier:
@bnmajor we definitely want to preserve the ability to "quick load" images -- in fact, I would like to add the same basic aggregation options to the simple load that we have for the imageseries "Load Data" panel (flipping, aggregation, dark subtraction).  I was unaware that menu item was slated for removal.

joelvbernier:
@bnmajor To further elaborate, I would imagine there is raw image data in memory which is updated from the loading callback; the raw view displays those data, while the cartesian and polar views are calculated from that data.  I'd say the behavior should be that any time the active image data is updated, all the views should be updated accordingly.

I am also working on the Mac, which might be part of the problem?  I noticed that my "home" buton in the matplotlib navigation toolbar spontaneously stopped working as well...

joelvbernier:
Update:

Ok -- if I toggle between loading single image files for each detector and imageseries while in
- cartesian mode: the canvas does update, but the appearance of the scrolling bar is "sticky" in that it does not appear if I go from single images to multiframe imageseries (or disappear in the opposite workflow) unless I toggle back to the raw mode view.
- polar mode: same behavior as above.
- raw mode: this seems to be working as intended.

cryos:
[...]

PR #284, opened 2020-03-13
Load the correct image mode for new files

bnmajor:
Whether using the load panel or the file menu, load the new images in the
currently select image mode (raw, cartesian, polar). Make sure that the slider
widget is updated in all modes.


bnmajor:
This does not create a new load utility class as @cryos has suggested, but this could provide a short-term fix the issues discussed in #283 .

bnmajor:
Closing in favor of the more complete solution in #286

PR #286, opened 2020-03-16
Load utility

### LINKED ISSUE #283: Loaded images not updated if in cartesian view

bnmajor:
As discussed in #283, we should keep the load functionality for both the menu and the load panel. To minimize duplicate code, an `ImageLoadManager` class has been created for both pathways to use. 

This should fix #283, as well as matching files selected via the file menu to their detector (which the load panel was already doing).

### DIFF
--- a/hexrd/ui/image_tab_widget.py
+++ b/hexrd/ui/image_tab_widget.py
@@ -92,13 +92,12 @@ class ImageTabWidget(QTabWidget):
 
     def load_images(self):
         self.update_image_names()
+        self.switch_toolbar(self.currentIndex())
 
         if HexrdConfig().tab_images:
             self.load_images_tabbed()
         else:
             self.load_images_untabbed()
-
-        self.switch_toolbar(self.currentIndex())
 
     def change_ims_image(self, pos):
         HexrdConfig().current_imageseries_idx = pos
@@ -154,6 +153,7 @@ class ImageTabWidget(QTabWidget):
 
     def show_cartesian(self):
         self.update_image_names()
+        self.switch_toolbar(self.currentIndex())
 
         # Make sure we actually have images
         if len(self.image_names) == 0:
@@ -168,6 +168,7 @@ class ImageTabWidget(QTabWidget):
 
     def show_polar(self):
         self.update_image_names()
+        self.switch_toolbar(self.currentIndex())
 
         # Make sure we actually have images
         if len(self.image_names) == 0:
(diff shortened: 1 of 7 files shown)


cjh1:
@bnmajor This is working well for me. Although it looks like a big change, its mainly moving code around? Looks like we have a conflict to resolve. Do we still need #284?

bnmajor:
@cjh1 It is mainly moving code around, largely copy and paste into a single class and updating function calls accordingly, rather than duplicating the code for both paths. The conflict is from the new progress bar updates, which modified some of the code that has been refactored. If this gets merged in we will not need #284, I originally made that PR in case we needed an immediate fix for packaging before this was complete.

cjh1:
@bnmajor Ok, sounds good, when you have resolve the conflict @psavery can take a look. We can go ahead and close the other PR then.

bnmajor:
@cjh1 @psavery This seems to still work as expected after fixing merge conflicts if you want to test it out as well

psavery:
I'm gonna try it out.

psavery:
@bnmajor FYI, the first time you switch to tabbed view, you get multiple toolbars at the bottom:

Screenshot from 2020-03-16 14-02-53

bnmajor:
Thank you, that is a problem... I'll look to see what I mixed up.

psavery:
So, when switching views (e.g., from raw to cartesian), the previous behavior was that the old canvas wouldn't be cleared until the new canvas was ready, and then it would draw the new canvas.

The new behavior is that the old canvas is cleared (so the canvas turns white immediately), and then when it finishes loading, the new image is drawn.

I personally prefer the old behavior, because I think it is easier on my eyes. But it is not necessarily a big issue, and probably wouldn't matter unless it bothered others also.

psavery:
Also, the progress dialog pops up for a second for the "quick load" of images when there is more than 1 image in a file (such as loading a `.npz` file). But this isn't necessarily a problem.

I think this PR is looking pretty good overall!

hexrd/ui/image_tab_widget.py
```diff
@@ -165,9 +171,11 @@ def show_cartesian(self):
         self.image_canvases[0].show_cartesian()
         self.addTab(self.image_canvases[0], '')
         self.tabBar().hide()
+        self.switch_toolbar(self.currentIndex())
```
psavery: By the way, in case we want to look at the canvas clearing sometime in the future, it looks like it is caused by this `self.switch_toolbar()` function.

PR #289, opened 2020-03-17
Restrict reasons slider visibility can be toggled on

bnmajor:
Visibility was being toggled on, only to be toggled off again, causing unnecessary clearing of the image canvas. This was causing a blank canvas to flash before loading the current image mode, which didn't look very nice.

### DIFF
--- a/hexrd/ui/image_series_toolbar.py
+++ b/hexrd/ui/image_series_toolbar.py
@@ -75,7 +75,8 @@ class ImageSeriesToolbar(QWidget):
             self.name = name
 
     def set_visible(self, b=False):
-        self.widget.setVisible(b)
+        size = len(self.ims) - 1 if self.ims else 0
+        self.widget.setVisible(b and size)
 
     def val_changed(self, pos):
         self.parent().change_ims_image(pos)
--- a/hexrd/ui/image_series_toolbar.py
+++ b/hexrd/ui/image_series_toolbar.py
@@ -75,8 +75,7 @@ class ImageSeriesToolbar(QWidget):
             self.name = name
 
     def set_visible(self, b=False):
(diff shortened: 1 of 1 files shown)


hexrd/ui/image_series_toolbar.py
```diff
@@ -75,7 +75,8 @@ def update_name(self, name):
             self.name = name
 
     def set_visible(self, b=False):
-        self.widget.setVisible(b)
+        size = len(self.ims) - 1 if self.ims else 0
```
cjh1: Is this test checking if we have images? If so I think we could do `self.widget.setVisible(b and bool(self.ims))`, may be I am miss understand what this check is doing ...

psavery: I think it's checking to see if we have more than one image in the imageseries.

bnmajor: The check is making sure we have more than one frame to view. Currently even `.tiff` images are loaded as an ImageSeries so that they can be rotated, etc as needed.

psavery: It might be a little clearer if it is something like `self.widget.setVisible(b and self.ims and len(self.ims) > 1)`, or if `self.ims` is guaranteed to be a dict, `self.widget.setVisible(b and len(self.ims) > 1)`.

bnmajor: Since everything is now being loaded through a shared pathway `self.ims` should always be a dict, so I've updated the check accordingly:
`self.widget.setVisible(b and len(self.ims) > 1)`.

cjh1: That looks better, thanks

Issue #264, opened 2020-02-20
Perform transformations on raw images


bnmajor:
Often at least one detector will need to be rotated when loading raw images, allow for transformations on load.

bnmajor:
It was mentioned in the meeting last week that it should be possible to transform raw images as well as an `ImageSeries`. This should all be possible through the load panel, as I believe that the intention was to eventually altogether remove the `File > Open > Images` menu option (although this may very well have changed).

@joelvbernier I am currently able to load `.tiff` files with the panel, and therefore perform transformations on them that way. Were you looking for this as an additional feature because you want to keep loading raw images separate from the load panel? Or are you looking for different functionality in the panel when loading raw images?

bnmajor:
As discussed in #283 and fixed in #286, we will continue to use both menu and load panel loading options, so this feature is still needed.

cjh1:
@bnmajor Is this now fixed? If so can we close this?

bnmajor:
Yes, I believe #290 solved this

PR #290, opened 2020-03-17
Transform menu images

bnmajor:
Add the option to transform images that have been selected through the file menu, as detectors may need to perform a simple rotation but not need all of the options or steps of the load panel.

This addresses #264 

transform


bnmajor:
Selections are remembered between sessions, but the state is currently shared with the load panel for reasons of easily re-using the existing image processing code. This means that whichever settings were changed last (load dialog or load panel) will be saved in the state and will be what is remembered and used in both cases the next time the program is run. These can be separated from each other if that is preferred.

This PR also fixes a bug that existed that was clearing the saved load panel state when the dummy images are loaded. While that is the behavior we want most of the time, on the initial load we want to remember the previous state so that we can set both the load panel and the load dialog correctly.

Issue #302, opened 2020-03-20
Use more decimals for tree view


psavery:
It currently rounds to two decimal places.

PR #308, opened 2020-04-06
Increase the precision of floats in the tree view

psavery:
Previously, when editing floating values in the tree view, the 
floats would immediately be truncated to 2 decimal places when editing
began, and it would not allow you to specify more than 2 decimal places
of precision.

To fix this, we are using our ScientificDoubleSpinBox class for floats,
which allows for scientific notation. Additionally, the previous commit
increased the precision of the scientific spinbox itself. These changes
will hopefully provide enough precision for the tree view.

Fixes: #302


psavery:
@cjh1 You might want to make sure that the behavior of the scientific spinbox didn't change in a problematic way for the form view. But when I tested it, I didn't see any noticeable differences, other than that we can now specify more decimal places (which is good).

Issue #300, opened 2020-03-20
Make table on materials tab popout from panel


cryos:
The table needs more space, doesn't need to be there all the time,
![image](https://user-images.githubusercontent.com/42214/77200141-617bf480-6ac0-11ea-823a-6ab1efdb3674.png)
Make that something in a dialog and we can open the dialog on demand.

psavery:
Should this always be in a dialog? Or should there be an option to move it back and forth between the dialog and the panel?

joelvbernier:
I'd say a dialog is a good default -- not familiar with other options.  Are you suggesting a dialog that is dockable inside the panel if desired?

psavery:
Although I'm not 100% sure, I think we could make it into something like a dock widget. And you can add buttons to show/hide dock widgets.

But I can just do the dialog for now since it will be easier.

PR #311, opened 2020-04-28
Move materials table into a dialog

psavery:
This moves the materials table into a dialog, which is not shown
by default. It can be shown by clicking the new "Show Materials Table"
button in the materials panel. The materials table can be hidden by
clicking on the "x" button in the top right corner.

All of the functionality should be the same. Rows are highlighted if
they are to not be excluded. The dialog will also automatically
update when settings are modified in the materials panel.

These changes were done to free up some space in the materials panel.

materials_panel

with_table

I am happy to move the button if anyone thinks it would look nicer somewhere else.

Fixes: #300

### DIFF
--- a/hexrd/ui/materials_panel.py
+++ b/hexrd/ui/materials_panel.py
@@ -2,10 +2,11 @@ import copy
 import math
 import numpy as np
 
-from PySide2.QtCore import QItemSelectionModel, QObject, Qt
-from PySide2.QtWidgets import QMenu, QMessageBox, QTableWidgetItem
+from PySide2.QtCore import QObject, QSignalBlocker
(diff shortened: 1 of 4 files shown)


psavery:
I rebased this on master

joelvbernier:
Patrick -- I just noticed that @donald-e-boyce did a commit from a clone of the `instrument-package` branch, leading to both an instrument package _and_ module in the repo.  This causes import errors, as the package is not ready for prime time.  I just deleted the package from my master to do testing related to the overlays (about to submit a PR) and it seems all good.  I will make a PR to fix the issue from my master rather than trying to play commit-undo-jung-fu.  The badness appears isolated to the presence of the instrument package.

PR #316, opened 2020-05-26
Create new HEDMInstrument class from config dict

psavery:
A [new HEDMInstrument class](https://github.com/HEXRD/hexrd/blob/eb2f5ffd288287df9c331755b240d90a9bb7ea25/hexrd/instrument/instrument.py#L167) was added as a part of hexrd v0.8.
This HEDMInstrument class does not currently accept an instrument
config for initializing it, and this causes errors in hexrdgui
for the cartesian and polar viewers.

To adjust our code for this, add a new function for creating an
HEDMInstrument from the config, and use this when we need to create
an instrument.

hexrd/hexrd#19 is also needed for the Cartesian viewer to work with
hexrd v0.8.

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -2,9 +2,9 @@ import warnings
 
 import numpy as np
 
-from hexrd import instrument
 from hexrd.gridutil import cellIndices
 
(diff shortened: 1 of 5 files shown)


psavery:
I think this is working and in good condition, and it is ready for review.

Note: this needs to be merged for hexrdgui to work with the master branch of hexrd.

Issue #310, opened 2020-04-17
Multiprocessed/multithreaded data loading


joelvbernier:
In the course of using the GUI for an 8-detector setup, it has become clear that we desperately need to multiprocess or multithread the load data actions across the individual panels.  Not sure which one is better from a performance point of view, but it would be simple to throw to a process pool as they run independently.

psavery:
I think I have some good work towards this. I have one example that is running much faster due to multiprocessing.

It looks to me that multiprocessing might be better for us than multithreading because for multithreading, the GIL will only allow one thread of python code to execute at a time. The [multiprocessing library](https://docs.python.org/3/library/multiprocessing.html) also makes multiprocessing easy.

However, for multiprocessing, the objects that are passed to the processes need to be pickle-able. It looks like the framecache (.npz files) imageseries is pickle-able, but hdf5 imageseries is not (since it holds onto an h5py object, which is not pickle-able, but we may be able to change that).

@joelvbernier Is there a way we can get the data for the 8 detector setup so we can test with that as well?

psavery:
Actually, since the numpy functions are implemented in C/Fortran, it might also be faster to do multithreading instead of multiprocessing because it avoids the pickling. Maybe the best option is to benchmark.

psavery:
As a side note on this issue as well, I noticed that, for my examples, if you perform dark background subtraction, it takes significantly longer (5-8x longer) to aggregate the images. It looks like this is because, in my examples, the original data is uint16, and after we perform the background subtraction, the output data is float64. This increases the size of our data by 4x.

Can we truncate that back down to a uint16 before aggregating in order to significantly improve performance?

donald-e-boyce:
That's a good question and the issue should be clarified.  

The imageseries.process module (line 57) preserves the datatype on dark subtraction. That made most sense to me.
```
def _subtract_dark(self, img, dark):
    # need to check for values below zero
    # !!! careful, truncation going on here;necessary to promote dtype?
    return np.where(img > dark, img - dark, 0)
```

psavery:
@donald-e-boyce Yeah, so it looks like when we create a dark image using one of the aggregation functions (like `imageseries.stats.median`), the input is `np.uint16`, and the output is `np.float64`, which probably makes sense, because the stats output can be fractions.

[...]

PR #319, opened 2020-05-28
Multiprocess/multithread data aggregations

psavery:
TMP: multiprocess and multithread timing tests

This is a temporary commit with timing tests for using multiprocess
and multithread for aggregating images.

For: #310

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -1,5 +1,9 @@
 import copy
+import functools
+import multiprocessing
 import os
+import threading
+import time
 
 from hexrd import imageseries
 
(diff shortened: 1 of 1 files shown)


cjh1:
I think I can see what is going on here, trying to work out the best fix ...

joelvbernier:
Is this the 8-detector case?

cjh1:
Yes, I have few more changes that I will push soon.  I see better speedups on my machine, as I have a physical core per detector/thread.

joelvbernier:
Question: what about multithreaded load?

cjh1:
I think we decided for now that is wasn't the largest bottleneck and would revisit later.

Issue #298, opened 2020-03-20
Updates to conda install instructions


psavery:
~~It looks like `psutil` is not being installed. I had to install it myself with `conda install -c conda-forge psutil`. We should probably add that (it is required for hexrd master).~~

The above is now fixed in hexrd master

~~I had to **not** use `-e` when installing hexrd and hexrdgui, like this:~~
```
pip install --no-deps -U ./hexrd
pip install --no-deps -U ./hexrdgui
```

~~Because, if I used `-e`, the program would complain when I tried to load it that `PySide2` is missing. I think it checks for the dependencies in the `setup.py` upon running if it is installed via an egg, or something like that.~~

I believe I was only having the above issues because I was not including the `CONDA_BUILD` environment variable that the README mentions to use.

I will test this and confirm later, but I think this is not an issue.

psavery:
We may also want the conda install instruction to be:

```
conda install -c cjh1 -c anaconda -c conda-forge
```

Which gives preference to packages being installed from the `anaconda` channel over `conda-forge`.

PR #324, opened 2020-06-05
Add -c anaconda to conda instructions

psavery:
`-c anaconda` is added before `-c conda-forge` to give preference to
packages on anaconda if they are there, and then conda-forge if they
are not.

Hopefully, the packages on anaconda are more stable because they are
maintained by anaconda, rather than the community.

Additionally, numpy on anaconda uses mkl, and numpy on conda-forge
uses libopenblas. We have had problems with libopenblas on the mac,
so maybe mkl will be more stable.

Fixes: #298
Fixes: #299

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ pip install -e hexrdgui
 
 ```bash
 # Install deps using conda package
-conda install -c cjh1 -c conda-forge hexrdgui
+conda install -c cjh1 -c anaconda -c conda-forge hexrdgui
 # Now using pip to link repo's into environment for development
 pip install --no-deps -U -e hexrd
 CONDA_BUILD=1 pip install --no-deps -U -e hexrdgui
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -25,7 +25,6 @@ requirements:
     - fabio
     - pyyaml
     - hexrd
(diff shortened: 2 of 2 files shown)


psavery:
This is now working for me on Linux.

Will try Mac soon.

psavery:
I gave this a try on my older mac that I have here. I think it worked very well for the most part. No crashing when switching tabs. I was able to do many different things:

1. Load GE yaml file and dataset, view all three view modes, use slider widgets, run powder calibration, run line-picked calibration, play with overlays
2. Load tardis yaml file and dataset, view raw and polar views, use slider widgets
3. Load dexela yaml file and dataset with max aggregation and median for dark, view all three view modes, use slider widgets.

So it seems to be working well for the most part. I only ran into a few issues:

1. The process was killed (not a seg fault) during aggregation of a large dataset. But I think this is because I am running it on an older mac. It worked the second time I tried.
2. After loading images, with the quick-load, I ran into an error when I tried to load new images with the load data panel. I'll post an issue about it if I can reproduce it.
3. The pop-outs (materials table and line-picked calibration zoom window) do not remain in front of the application when you click on the application window. I'll post an issue about this.

But it looks like the installation worked for me.

joelvbernier:
works!

Issue #321, opened 2020-06-05
Polar view error with hexrd master


psavery:
With the [current hexrd master](https://github.com/HEXRD/hexrd/tree/031d348dd27661d22d726dfdd4cc0838ca73a38b), there is an error in hexrdgui if you try to display the polar view for our tiff image 4-detector ge example. I am not sure what may have changed to cause this.

The error is pasted below.

```
Traceback (most recent call last):
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 14, in polar_viewer
    return InstrumentViewer()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 32, in __init__
    self.draw_polar()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 51, in draw_polar
    self.pv.warp_all_images()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polarview.py", line 247, in warp_all_images
    self.create_warp_image(det)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polarview.py", line 209, in create_warp_image
    self.warp_dict[det] = panel.interpolate_bilinear(
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/instrument.py", line 2061, in interpolate_bilinear
    + (ij_frac[:, 1] - j_floor)*img[i_floor, j_ceil]
IndexError: index 2048 is out of bounds for axis 1 with size 2048
```

joelvbernier:
Ok, figured this out.  I added an attribute on the detector panels to hold the buffer/mask that defaults to `None`; the routines that clip points to the panel extent uses the panel_buffer attribute.  This probably needs a bit more thought.  Anyhow, if you add a "buffer: <nonzero float in mm>" line in the instrument config, it works (see TARDIS example below).

This attribute in the yml file obviously leaves something to be desired if the buffer is an mask array (will currently print it out as a flattened list!).

Might need to consider migrating the instrument config to something more amenable to storing arrays, like HDF5?  

```
beam:
[...]

PR #326, opened 2020-06-08
Add a default panel buffer to the detectors

psavery:
If a panel buffer is not specified, then default to 0.5. This is
required with some changes in hexrd master.

Fixes: #321

### DIFF
--- a/hexrd/ui/resources/calibration/default_instrument_config.yml
+++ b/hexrd/ui/resources/calibration/default_instrument_config.yml
@@ -11,6 +11,7 @@ detectors:
       columns: 2048
       rows: 2048
       size: [1.0, 1.0]
+    buffer: 0.5
     saturation_level: 14000.0
     transform:
       translation: [0.0, 0.0, 0.0]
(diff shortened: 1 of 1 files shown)


joelvbernier:
@psavery there is something more sinister going on here that I need to get to the bottom of.  I am testing the PolarView class to see how it is possible to get an index error when we are checking points that originated on the detector...  Shouldn't need a panel buffer.

joelvbernier:
Ok, I figured out what the issue was.  The edge case for points on the edges of the detector were not being handled properly in the `instrument.PlanarDetector.interpolate_bilinear()` method.  [This](https://github.com/HEXRD/hexrd/commit/cc1efd0eff98826368c5a92d23f0eacc1f10dc9a) commit to hexrd fixes that.  You should be able to remove the forced panel buffer attr set, as there is no good universal number for that anyhow.  Sound reasonable @psavery ?

psavery:
Sounds good, Joel! This is fixed in #334.

Issue #320, opened 2020-06-05
Use new overlay classes for creating overlays


psavery:
I think I have been making some good progress on this so far. I have added ranges and range indices (to check for merged ranges) as output to the powder overlay class, and Cartesian and Polar are working so far with some re-centering (and the Cartesian is receiving a fake single-detector instrument).

However, I'm not sure what's going on with the overlays for the raw view. They are appearing in a different corner, facing the wrong way, and are smaller than they should be.

![image](https://user-images.githubusercontent.com/9558430/83891162-c1c31f80-a71a-11ea-9958-f60f2de3968a.png)

I'm actively working on this to figure it out.

By the way, for the re-centering, the output from the polar version gave a 0° - 360° range for η. I just needed to convert this to a -180° to 180° range, which is what we had been using before.

For the Cartesian version, the output appears to be centered on (0, 0), whereas our Cartesian view has had (0, 0) in the upper left corner. Re-centering the output fixes this.

One last thing we might want to consider changing in the powder overlay class is that it currently outputs the overlays in a single list with nan-delimiters, like this:
```
[1, 2, NaN, 3, 4]
```

We might consider at some point changing it to this (where each overlay is in its own list):
```
[[1, 2], [3, 4]]
```

We are currently using the first case and creating a single matplotlib Line2D object for all of the overlays. It might make more sense to have a Line2D object for each overlay instead. But there may also be some performance advantages to the first method.

joelvbernier:
Looks like we need to conver from (x, y) cartesian to pixel coordinates.  The detectors each have a `Planardetector.cartToPixel()` method that takes in vstacked (x, y) pairs and outputs vstack (i, j) -- row, col -- pairs.  So depends if we want to display the raw images with cartesian axes or pixel-based.  If the latter, we need to convert the overlay arrays.

PR #309, opened 2020-04-16
Draw material overlays in raw view

psavery:
This utilizes `PlanarDetector.make_powder_rings()` to draw the material
overlays on the raw view.

It looks good for the most part, except that some detectors produce
`ring_xys` values from `PlanarDetector.make_powder_rings()` that wrap
around to the other side, so that a line is drawn across the detector
connecting the ends. I am not sure if this is something that should
be fixed in hexrdgui, or in hexrd. See "ge2" below for an example.

Additionally, the values do not appear to go all the way to the image
borders for some reason.

Screenshot from 2020-04-16 14-50-50

Screenshot from 2020-04-16 14-52-19

Screenshot from 2020-04-16 14-52-51

Fixes: #297

PR #329, opened 2020-06-08
Migrate overlay generators to new overlay class

psavery:
This migrates the old overlay generators for each of the three views
to use the new PowderLineOverlay class.
    
This also includes some slight re-formatting of the overlay data that
comes out, mostly to re-center it.

There is additionally another commit that modifies the PowderLineOverlay class:

Return ranges from PowderLineOverlay as well

This modifies the PowderLineOverlay so that ranges and range indices
can be returned from the function as well. It does so by now returning
a dict that contains:

1. The rings
2. The ranges
3. The range indices

A large part of the `overlay()` function was moved into another function
so that it can be called multiple times for both the rings and the ranges.

A modification was also made so that the raw mode converts to pixel
coordinates, which is used in the raw images.

Finally, the overlay data is now also returned as a list of separate
overlays, rather than as a single nan-delimited list containing all 
overlays. This better matches the behavior before, and it seems to
make more sense to have a separate matplotlib Line2D for each overlay.

Fixes: #320


psavery:
To do:

- [ ] Rings are drawn twice when made visible. Fix this.
- [ ] If polar view is active when new images are loaded, azimuthal integration plot doesn't update properly. Fix this.

psavery:
This is ready for review.

psavery:
The overlays in the raw view look much nicer with these changes, by the way, compared to #309

image

hexrd/ui/image_canvas.py
```diff
@@ -233,6 +248,28 @@ def draw_detector_borders(self):
...
+        if self.mode == 'cartesian':
+            # It's centered on (0, 0). Recenter it in place.
+            self.recenter_points((0, 0), x, y)
+        elif self.mode == 'polar':
+            # Swap axes and convert to degrees.
+            x, y = np.degrees(y), np.degrees(x)
+            # It's centered on 180. Recenter on 0.
+            y -= 180.0
+
+        return x, y
```
joelvbernier, line 262: Is this necessary?

hexrd/ui/image_canvas.py
```diff
@@ -233,6 +248,28 @@ def draw_detector_borders(self):
...
+
+    def recenter_points(self, old_center, x, y):
+        # Recenters the points in place
+        new_extent = self.axes_images[0].get_extent()
+        new_center = ((new_extent[1] + new_extent[0]) / 2.0,
+                      (new_extent[3] + new_extent[2]) / 2.0)
+
+        x += new_center[0] - old_center[0]
+        y += new_center[1] - old_center[1]
```
joelvbernier, line 271: Same here; the coordinate system doesn't change.  Not sure why we need to recenter?  Or is this on mode change?  I noticed some weird behavior with the overlays in the current master with them moving around.  What precipitated the need to "recenter"?

psavery: Yeah, this is what I get without the re-centering:

cartesian

See that the rings are centered on (0, 0), rather than in the center of the image.

If our Cartesian viewer should be centered on (0, 0) as well, I can look into why it isn't doing that. Otherwise, I might be able to find a better fix for moving the overlays. I agree that this might cause some strange issues where the overlays move when we change parameters.

hexrd/ui/image_canvas.py
psavery: I do need to convert it to degrees and swap the axes, but maybe I can do those in the powder overlay class instead of here.

Without the re-centering on 0, it looks like this:

polar

But actually, since we are drawing vertical lines over the whole image, I can probably just use [axvline](https://matplotlib.org/3.1.0/api/_as_gen/matplotlib.axes.Axes.axvline.html) in matplotlib. That way, I don't need to supply y coordinates, and it won't be an issue.

psavery: I have moved the reformatting of the polar view to the overlay class in b75d04645ccd60bda2b7523c12d4169b74fdf8fd. This should hopefully be resolved.

hexrd/ui/image_canvas.py
psavery: Okay, I have some more info and I think I figured out a better way to do it.

Looks like it might be more ideal for me to do it this way:

1. Always convert Cartesian overlays to pixel coordinates in the overlay class (only raw overlays were being converted to pixel coordinates before).
2. Rescale the Cartesian overlays to match the rescaling of the Cartesian image.

The reason 2 is necessary, is because we are rescaling the extents of the Cartesian image to correct physical units in millimeters (#186). Thus, we need to rescale the Cartesian overlays as well.

psavery: This should be fixed in 0c5caae1567f377449c0c4d7ea407b567e49bed0. Let me know if there are any issues with it.

cjh1:
We will merge and changes can be done in a follow up.

Issue #305, opened 2020-04-01
Auto-select cartesian view params based on loaded instrument config


joelvbernier:
Feature request -- auto select the `Distance` and `Pixel Size` based on the average values of the `translation` (the last component, labeled `2`), and the `2*pixels:size` across the detectors.

psavery:
@joelvbernier Should we remove the options to set these in the image mode widget, or just reset their values via the calculation every time the cartesian tab is selected?
![Screenshot from 2020-04-10 10-17-13](https://user-images.githubusercontent.com/9558430/78997387-7aeaec00-7b14-11ea-9afe-b17ebe8cea70.png)

joelvbernier:
I think we should keep the interactors -- only auto-populate on load/reload of instrument config.  If the user wants a nice high-res figure (read: slow rendering), they can choose to do that.  I think a good rule of thumb is to use the average detector distance at 3x the native (or largest) pixel size.  We can adjust as necessary.

psavery:
This is fixed in #338

PR #338, opened 2020-06-15
Auto-generate Cartesian params from instrument config

psavery:
When a new instrument config is loaded (either by
"File"->"Open"->"Configuration", or by resetting the instrument
configuration), automatically generate Cartesian parameters based
upon the config.

The parameters are currently calculated as follows:

1. Distance: absolute value of the average of detector z translation values
2. Pixel size: 5x the average of the detector pixel sizes

I ended up using 5x the average for the pixel size because that gave me a
reasonably interactive setup for my personal computer. But we can adjust it
if something else is better. Less than 5x, such as 3x, made for much slower
interaction for me.

I did a little bit of re-formatting int his PR as well. But the main changes of
interest are [here](https://github.com/HEXRD/hexrdgui/compare/master...psavery:auto-generate-cartesian-params?expand=1#diff-30bfcc80b3692dfa8947f36bbec62f14R109-R130).

Fixes: #305

### DIFF
--- a/hexrd/ui/image_mode_widget.py
+++ b/hexrd/ui/image_mode_widget.py
@@ -50,6 +50,8 @@ class ImageModeWidget(QObject):
             HexrdConfig().set_polar_snip1d_width)
         self.ui.polar_snip1d_numiter.valueChanged.connect(
             HexrdConfig().set_polar_snip1d_numiter)
+        HexrdConfig().instrument_config_loaded.connect(
+            self.auto_generate_cartesian_params)
 
         self.ui.polar_show_snip1d.clicked.connect(self.polar_show_snip1d.emit)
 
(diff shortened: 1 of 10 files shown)


psavery:
@joelvbernier Let us know what you think of this.

hexrd/ui/image_mode_widget.py
```diff
@@ -79,41 +81,50 @@ def all_widgets(self):
...
+        self.ui.polar_res_tth_max.setValue(
+            HexrdConfig().polar_res_tth_max)
+        self.ui.polar_apply_snip1d.setChecked(
+            HexrdConfig().polar_apply_snip1d)
+        self.ui.polar_snip1d_width.setValue(
+            HexrdConfig().polar_snip1d_width)
+        self.ui.polar_snip1d_numiter.setValue(
+            HexrdConfig().polar_snip1d_numiter)
+
+    def auto_generate_cartesian_params(self):
```
joelvbernier, line 109: This is fine for the default setting (_i.e._, using the z component alone) with the plane normal anti-parallel to the beam direction; not if the plane is oblique (would need to use norm ar something).  But for generating the *initial* parameters for the view upon load, the logic makes sense.

Issue #341, opened 2020-06-17
Cartesian view mode ignores beam vector


joelvbernier:
Just caught this with a new example.  The cartesian mode overlays do not recognize the beam vector.  Didn't catch it before with the examples we were using because the beam vector default was correct for those examples.  I think the best way to fix this is to make sure that the fake instrument for the display plane inherits the beam vector from the config dict.  I havent' checked yet, but I assume that it currently does not...

joelvbernier:
Yep @psavery -- looks like the `temp_instr` needs to have its bvec attribute set to the instrument's `beam_vector` attribute.  It looks like it might not be?  The intrument copy presumably has the correct beam vector, but at the detctor level, [`self.dplane`](https://github.com/HEXRD/hexrdgui/blob/f6c124a455be94f27b2773607d946eaf2b6cd17b/hexrd/ui/calibration/cartesian_plot.py#L83) does not.  When you instantiate an instrument class, it braodcasts its beam vector to all the component detectors.  However, if we clobber and assign directly to the `detectors` dict, it carries whatever beam vector it had attached to it.  The `__init__()` method of the instrument.PlanarDetector` class has an optional kwarg "bvec" that sets the `_bvec` attribute, which is protected/settable under the `bvec` property.

joelvbernier:
Question for @psavery or @cjh1: which dict hanging off of `HexrdConfig()` has the current instrument config values?  Is it `HexrdConfig().internal_instrument_config`?

psavery:
@joelvbernier Sounds good. I'll have a fix for this soon.

psavery:
I gave the above proto fix a try to see if I could reproduce the elliptical overlays for my GE example. Unfortunately, they still look circular. You can see the diff for my proto fix below.

Let me know if you upload that example somewhere so I can try it out.

![image](https://user-images.githubusercontent.com/9558430/84946033-0acd8900-b0b6-11ea-9b97-70308fb0964e.png)

```diff
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -65,6 +65,7 @@ class InstrumentViewer:
         # must update self.dpanel from HexrdConfig
         self.pixel_size = HexrdConfig().cartesian_pixel_size
         self.make_dpanel()
+        self.dpanel.bvec = self.instr.beam_vector
 
         # The overlays for the Cartesian view are made via a fake
         # instrument with a single detector.
```

PR #342, opened 2020-06-17
Update overlays when beam vector is changed

psavery:
When the beam vector is changed, update the beam vector on the
instrument that the overlay has, and then redraw the overlays.

Peek 2020-06-17 14-03

Fixes: #341

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -50,7 +50,8 @@ class InstrumentViewer:
     def make_dpanel(self):
         self.dpanel_sizes = self.dplane.panel_size(self.instr)
         self.dpanel = self.dplane.display_panel(self.dpanel_sizes,
-                                                self.pixel_size)
+                                                self.pixel_size,
+                                                self.instr.beam_vector)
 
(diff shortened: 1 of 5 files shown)


psavery:
@joelvbernier I think this is working, except that it doesn't look to me that the rings change shape, which is what it looks like [here](https://github.com/HEXRD/hexrdgui/issues/341#issuecomment-645142309).

Can you confirm that this is working properly?

joelvbernier:
Yeah, somehting isn't right...  they should be going from round to elliptical (or other conic section).  I'll look right now.

hexrd/ui/calibration/cartesian_plot.py
```diff
@@ -50,7 +50,8 @@ def __init__(self):
     def make_dpanel(self):
         self.dpanel_sizes = self.dplane.panel_size(self.instr)
         self.dpanel = self.dplane.display_panel(self.dpanel_sizes,
-                                                self.pixel_size)
+                                                self.pixel_size,
+                                                self.instr.beam_vector)
```
joelvbernier, line 54: maybe assign "bvec=" just to be explicit?  I know is doesn't matter...

PR #343, opened 2020-06-17
Transform menu

bnmajor:
Allow detector images to be transformed after load. Although the transform can be set for each detector for quick load, mistakes may accidentally be made. This adds an option in the `Edit` menu to `Transform Detectors`. The default option is to apply the transform to all detectors, but if `Update Individual Detectors` is selected each detector can be adjusted separately.

transform_dialog

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -110,7 +110,7 @@ class HexrdConfig(QObject, metaclass=Singleton):
         self._tab_images = False
         self.previous_active_material = None
         self.collapsed_state = []
-        self.load_panel_state = None
+        self.load_panel_state = {}
         self.polar_masks = []
         self.ring_styles = {}
         self.backup_tth_maxes = {}
@@ -191,7 +191,7 @@ class HexrdConfig(QObject, metaclass=Singleton):
 
         self.previous_active_material = settings.value('active_material', None)
         self.collapsed_state = settings.value('collapsed_state', [])
-        self.load_panel_state = settings.value('load_panel_state', None)
+        self.load_panel_state = settings.value('load_panel_state', {})
         self.ring_styles = settings.value('ring_styles', {})
 
         # Set this manually since we don't have any materials yet
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -162,7 +162,7 @@ class ImageLoadManager(QObject, metaclass=Singleton):
         # Process the imageseries
         self.apply_operations(HexrdConfig().imageseries_dict)
         if self.data:
-            if self.state['agg']:
+            if 'agg' in self.state and self.state['agg']:
                 self.display_aggregation(HexrdConfig().imageseries_dict)
             else:
                 self.add_omega_metadata(HexrdConfig().imageseries_dict)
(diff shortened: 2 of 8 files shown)


cjh1:
@bnmajor Is this ready for review?

bnmajor:
@cjh1 Yes

cjh1:
@bnmajor When does the `Transform  Detectors` option become enabled? I have tried changing the image transform, but it doesn't seem to become enabled.

bnmajor:
The transforms dialog was not being enabled when images were loaded through the load panel, only through the quick load option (`File` -> `Open` -> `Images`).

This should be fixed now to enable the menu option no matter which way the data is loaded.

cjh1:
1. Loaded the images using the image load panel
2. Change the aggregation.
3. Click 'Transform Detectors'

I get the following exception:

```
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/main_window.py", line 623, in on_action_transform_detectors_triggered
    td = TransformDialog(self.ui).exec_()
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/transform_dialog.py", line 15, in __init__
    self.update_gui()
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/transform_dialog.py", line 22, in update_gui
    for i, det in enumerate(HexrdConfig().get_detector_names()):
AttributeError: 'HexrdConfig' object has no attribute 'get_detector_names'
```

bnmajor:
To try and reproduce this error I:

- Loaded the images using the load panel
- Changed the aggregation on the load panel and then read the images in again
- Used the transform dialog to flip the images

I did not get the `AttributeError: 'HexrdConfig' object has no attribute 'get_detector_names'`, so maybe I did not take the same steps you did? I do however get the following error:

```
  File "/home/brianna/hexrdgui/hexrd/ui/transform_dialog.py", line 67, in apply_transforms
    ilm.apply_operations(HexrdConfig().imageseries_dict)
  File "/home/brianna/hexrdgui/hexrd/ui/image_load_manager.py", line 255, in apply_operations
    ims_dict[key], ops, frame_list=frames)
  File "/home/brianna/hexrd/hexrd/imageseries/process.py", line 30, in __init__
    self._meta = copy.deepcopy(imser.metadata)
AttributeError: 'list' object has no attribute 'metadata'
```
This is being returned because the transform is being attempted on the aggregated image, not an ImageSeries object. The transform needs to be applied first, then the ImageSeries can be aggregated. This requires re-running the aggregation function, and that is precisely what the load panel would automatically do if the transform were set there instead of through the dialog. Should the dialog go back to only being available for images loaded through the file menu or should the dialog re-run the aggregation function if one is set?

cryos:
@joelvbernier may have other ideas, but I think it would be reasonable to only enable this for images loaded through the menu. It does speak to the bigger question of whether the load data panel would be improved if it kept the original image series around so that the transform could be modified. This is more a usability question as it makes no sense to a user to restrict applying transforms to one particular way of loading data.

cjh1:
@cryos @bnmajor and I spoken in person and decided that is made sense to allow the transformation regardless of how it is loaded. In the case of data loaded via the panel the aggregation will be re-run for now. There is a performance hit but at least the behavior is consistent, rather than having it enabled based on how the data was loaded.

bnmajor:
@cjh1 This should now fix the `AttributeError: 'HexrdConfig' object has no attribute 'get_detector_names'` error you were seeing earlier - the function call was updated to be a property in 326435a and has now been updated in this branch as well.

Also, as discussed, the transformation is allowed no matter which way the images are loaded. If loaded through the load panel any selections made there are kept, the transformation is applied, and if an aggregation has been selected it is then re-applied. The `TransformDialog` exists outside of the load panel pipeline which means it should work with any loaded images, but it should also be noted that this means that any selections made through the dialog will not be reflected in the load panel.

PR #344, opened 2020-06-17
Remove some unused parts and re-organize a little

psavery:
This removes some unused parts from the polar view, and it re-organizes the cartesian overlays to be a little more efficient.


joelvbernier:
thanks!

PR #347, opened 2020-06-19
Add github action packaging workflow

cjh1:
Here we go! :crossed_fingers: it builds first time :smile:


.github/workflows/main.yml
```diff
@@ -0,0 +1,177 @@
...
+      # command works.
+      shell: bash -l {0}
+
+    - name: Set version environment variables used by CPack
+      working-directory: hexrdgui
+      run: |
+          echo `python packaging/github_action_version.py ${{ steps.hexrdgui_describe.outputs.version }} full`
+          echo `python packaging/github_action_version.py ${{ steps.hexrdgui_describe.outputs.version }} major`
+          echo `python packaging/github_action_version.py ${{ steps.hexrdgui_describe.outputs.version }} minor`
+          echo `python packaging/github_action_version.py ${{ steps.hexrdgui_describe.outputs.version }} patch`
```
psavery, line 121: I think these don't need to be "echo'ed". I did something similar [here](https://github.com/hydroframe/SandTank/blob/b1a6d50ba33187ab1fcb54c760ba71fa1cfb3ab4/.github/workflows/docker.yml#L50), where the script itself sets a variable, and I don't think an echo was needed when calling the script.

cjh1: But it's a python script?

psavery: I think github actions just checks all of the stdout that comes out of the `run` command, and if anything starts with `::set-env`, it considers it to be telling it to set a variable.

psavery:
It works! :+1:

.github/workflows/main.yml
psavery: Or `::set-output`

.github/workflows/main.yml
```diff
@@ -0,0 +1,177 @@
...
+        shell: bash
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        path: hexrdgui
+
+    - name: Checkout HEXRD master for local build ( PRs only )
+      if: github.ref != 'refs/heads/master'
```
psavery, line 43: By the way, I think these could also be `if: github.event_name == 'pull_request'`. But it's probably not that important.

.github/workflows/main.yml
```diff
@@ -0,0 +1,177 @@
...
+      shell: bash -l {0}
+
+    - name: Build the HEXRD conda package ( PRs only )
+      if: github.ref != 'refs/heads/master'
+      working-directory: hexrd
+      run: |
+          conda activate hexrd
+          mkdir output
+          conda build --output-folder output/ conda.recipe/
+          pwd
```
psavery, line 88: This might not be needed.

.github/workflows/main.yml
```diff
@@ -0,0 +1,177 @@
...
+      with:
+        name: InstallOutput.log
+        path: ${{ github.workspace }}/hexrdgui/packaging/_CPack_Packages/*/InstallOutput.log
+
+    - name: Upload WIX log ( Windows only )
+      if: ${{ failure() && matrix.config.os == 'windows-latest'}}
+      uses: actions/upload-artifact@v2
+      with:
+        name: wix.log
+        path: ${{ github.workspace }}/hexrdgui/packaging/_CPack_Packages/WIX/wix.log
```
psavery, line 147: These are nice!

.github/workflows/main.yml
cjh1: Ok, thanks, will fix that up

.github/workflows/main.yml
cjh1: Yup, I think either work in this case

.github/workflows/main.yml
cjh1: Removed

PR #357, opened 2020-06-24
Add some basic indexing

psavery:
Run the whole indexing and write results

The user can now run the whole indexing step, and the results get 
written out to files in their current working directory. They do it by:

1. Loading the correct instrument configuration file in hexrdgui.
2. Using the load panel to load images without aggregation.
3. Loading the correct materials.hexrd file for the materials.
4. Setting the active material.
5. Using the materials table to select the active hkls to use.
6. Click "Run"->"Indexing", select options in the tree view, and click "Ok".
7. Wait until it finishes. Results will be written to current working directory.

In the find_orientations config:

1. The instrument gets automatically set from the hexrdgui instrument.
2. The materials dict gets automatically set from the hexrdgui config.
3. The active material gets automatically set from the hexrdgui config.
4. The active hkls get automatically set from the materials' planeData.
5. The image series gets automatically set from the hexrdgui config.

This eliminates some of the options that the user sets in the tree view.

Some improvements that will be made in the future:

1. Bring up a window that will display the log information as it comes.
2. Add some visualization.
3. Allow for user feedback at various steps.

Depends on: 
hexrd/hexrd#34
hexrd/hexrd#35


psavery:
Just to note: this does not include the omega eta map viewer. That will come in a PR soon...

Issue #327, opened 2020-06-08
Add indexing interfaces


psavery:
Part of this workflow is provided in [this file](https://github.com/HEXRD/hexrd/blob/master/hexrd/cli/find_orientations.py).

I've uploaded the example yaml we are working with in the shared drive at `examples/indexing/multidexela/mruby_config_composite.yml`.

The CLI can be ran via:
```
hexrd find-orientations <hedm yaml config>
```

PR #363, opened 2020-06-30
Add eta omega maps viewer and widgets for indexing options

psavery:
This is all content for under "Run"->"Indexing".

This first brings up a dialog, where you can either load or generate eta omega maps:
image

Next, a dialog appears like the following:
image

Whichever method tab is selected is the method that will be performed.

The minimum value that is set in the color map is what is used for the threshold.

Currently, all hkls in the eta omega map are ran in the indexing.

Once the user selects "Ok", the indexing is ran. No visualization of the results is performed yet.

Depends on:
hexrd/hexrd#38
hexrd/hexrd#40

Fixes: #327


cjh1:
Looking good!

psavery:
This is working and in good condition. Ready for review.

PR #365, opened 2020-07-01
LLNL Masking

bnmajor:
This is a first pass at the features for the LLNL masking needs, currently in a separate dialog. These features can be added into the main GUI down the line, or remain separate, whichever is deemed best. Given that this may be adapted into the main GUI at some point, not as much time has been spent on dialog design and this may need to be improved on if this tool remains its own dialog.

Before opening the dialog the correct configuration file for the detector(s) that are being worked on must be loaded through the main GUI. The dialog can then be found under `File -> Open -> LLNL Images` (again, a temporary location that can be better named/re-located should this remain separate).


bnmajor:
image_loader
The top group of the left panel contains the options for loading in an image. The detector selection is there because the detector may not be able to be inferred from the raw image name, and this information is needed in order to pull data from the config file. The file can then be loaded through the 'Load Image' button, and once the image has been loaded the masking options will be made available.

bnmajor:
mask_image
The bottom group in the left panel contains the options for masking the image. To minimize confusion and allow layering of masks, once an option (template or threshold) has been selected, the other will become unavailable until the mask has been accepted or discarded.

If a template is selected it will be drawn over the image and can be drug around an repositioned. Once it has been accepted it cannot be moved any more.

If the template has been positioned or the threshold has been set the mask can either be discarded (if it was a mistake/incorrectly selected) or accepted. At this point there is no notable change when the mask is accepted, but the template and threshold options will be reset. The `Preview Masks` option can be toggled on or off to view the image with the mask applied.

bnmajor:
Selecting cancel will close the dialog and discard any changes. Selecting save will allow the mask to be saved as `.npz`

template_dialog

bnmajor:
Core functionality that is still needed:
- [ ] Mask out selections by drawing
- [ ] Ability to scale template
- [ ] Ability to rotate template

cjh1:
@joelvbernier It would be good to get some early feedback on this when you get a chance.

bnmajor:
Closing in favor of #410

joelvbernier:
This looks like some really nice progress @bnmajor!  Will have to play with it this weekend.

PR #367, opened 2020-07-01
Use signal to update progress dialog text

johnkit:
On some machines, a hard crash can occur when the progress dialog label is updated. To avoid that, the direct call is replaced with a signal-slot connection.


joelvbernier:
ready to merge this one?

PR #377, opened 2020-07-06
Don't call clear from __del__

cjh1:
The clear function accesses the canvas with will already be cleaned up my this time. Basically clear is already called by either  accept or reject.

Fixes: #376


cjh1:
Fixes #376

psavery:
I cannot remember why I did this. I thought there was some reason for it. But I can't figure it out.

We can approve this and re-visit if we run into issues in the future.

PR #378, opened 2020-07-07
Recompute polar masks in main window update_all()

### LINKED ISSUE #361: Apply mask to polar causing IndexError
I noticed that if changing resolution in the polar view, the mask throws and IndexError (makes sense)

```
IndexError: boolean index did not match indexed array along dimension 0; dimension is 360 but the corresponding boolean dimension is 180
```

I think we will need a manager for mask elements, and for a change of polar view parameters should either

1. save the mask points in angular coordinates and reapply, or 
2. clobber existing masks

we would also want the ability to add to a mask "layer" successively.

johnkit:
The new code stores the mask line data in HexrdConfig and constructs the image masks when called by MainWindow.update_all(). The create-mask logic is updated to obtain the current image dimensions to use when generating the masks.

Closes #361


hexrd/ui/hexrd_config.py
```diff
@@ -156,6 +157,7 @@ def __init__(self):
 
         self.update_visible_material_energies()
 
+
```
cjh1, line 160: Extract space?

hexrd/ui/create_polar_mask.py
```diff
@@ -5,15 +5,30 @@
 from hexrd.ui.hexrd_config import HexrdConfig
 
 
-def create_polar_mask(line_data, rsimg, pv):
+def create_polar_mask(line_data):
+    # Calculate current image dimensions
+    eta_min = -180.0
+    eta_max = 180.0
```
cjh1, line 11: We  can probably pull these out as constants

cjh1: We have a constant module [here](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/constants.py)

hexrd/ui/hexrd_config.py
johnkit: Force-pushed fix for this

hexrd/ui/create_polar_mask.py
johnkit: Added UI_ETA_MIN_DEGREES and UI_ETA_MAX_DEGREES to constants.py (force push)

cjh1:
One question about this that @psavery raise is that we are currently recreating the mask every time we re-render, which may not be necessary. If the create of the mask isn't expensive, then I don't think that is a problem. I will do a quick test to see how long  the mask creation takes.

cjh1:
Take about 0.9 milliseconds to generate a mask, so for I don't think we need to worry!

Issue #252, opened 2020-02-11
auto-compute initial polarview pixel sized based on instrument config


joelvbernier:
Given the large variation in detector resolutions we deal with, I'd like to use the current instrument config to auto generate starting values for the polarview pixel sizes (tth, eta).  The code snippet below calculates the min, median, and max angular pixel sizes over the whole instrument.  For lack of anything better to try first, let's use the median as the 'default' starting resolution.  We can probably do something similar for the cartesian mode as well.  This would prevents it from rendering at an unreasonably high resolution by default.

```
    # next find the median angular resolution on the instrument
    # to use for interpolation
    ang_px_all = []
    for k, v in instr.detectors.items():
        X, Y = np.meshgrid(v.col_pixel_vec, v.row_pixel_vec)
        ang_px_all.append(
            v.angularPixelSize(np.vstack([X.flatten(), Y.flatten()]).T)
        )
    ang_px_all = np.vstack(ang_px_all)
    min_px = np.degrees(np.min(ang_px_all, axis=0))
    med_px = np.degrees(np.median(ang_px_all, axis=0))
    max_px = np.degrees(np.max(ang_px_all, axis=0))
    pv_pixel_size = med_px
```

psavery:
This sounds good. However, I should note that it looks like this code takes ~13 seconds to run on my desktop with our GE example (see image below). 

And for reference: our default (tth, eta) resolutions are (0.05, 0.2) degrees. This function produces (tth, eta) resolutions of (0.004, 0.028) degrees for the example below.

![image](https://user-images.githubusercontent.com/9558430/74277095-eb2beb80-4ce4-11ea-9223-d877282e3eb7.png)

joelvbernier:
Hmmm...  good point.  Maybe we should 10x median?  Or some scaling of the max pixel size...  But yes, I agree that is too slow.  Not sure if there are many ways to speed the calc up internally -- could multiprocess it over the list of multiple detectors if it isn't already.

psavery:
Yeah, it's currently doing it all in one thread. Multiprocessing might be an option.

Also, I wonder if a random sampling might work?

joelvbernier:
Patrick, how about we set the defaults to 100x the median?  That way it will render fast for the initial view.  If it looks too chunky, then maybe we relax to 10.  That should be generally applicable.

donald-e-boyce:
I'm starting to think average might be a better choice than median for
estimating the background. I noticed that the median image for the frame
[...]

PR #388, opened 2020-07-08
Auto-compute polar resolutions from config

psavery:
When an instrument config file is loaded, or the instrument config
is reset, the polar resolutions are automatically computed and set
based upon the config.

This is currently being done by scaling the max. However, we should
also consider median and min. We most likely need to do some
kind of scaling, so that we can come up with reasonable values for
most instrument configs.

Fixes: #252

To do:
- [x] Figure out how to speed this up. It makes loading new configs and resetting the config much slower.
- [x] Try out a few instrument configs, and ensure [the max](https://github.com/HEXRD/hexrdgui/compare/master...psavery:auto-compute-polar-parameters?expand=1#diff-30bfcc80b3692dfa8947f36bbec62f14R149) is a good option to scale. We can also try the median and the min.
- [x] Try out a few instrument configs, and ensure that [these scaling numbers](https://github.com/HEXRD/hexrdgui/compare/master...psavery:auto-compute-polar-parameters?expand=1#diff-30bfcc80b3692dfa8947f36bbec62f14R155-R156) are satisfactory.

We'd prefer to err on the side of too high of numbers, rather than too low, because too low can potentially make the rendering take a very long time.

### DIFF
--- a/hexrd/ui/image_mode_widget.py
+++ b/hexrd/ui/image_mode_widget.py
@@ -1,8 +1,10 @@
 import copy
+import numpy as np
 
 from PySide2.QtCore import QObject, QSignalBlocker, Signal
 
 from hexrd.ui.constants import UI_RAW, UI_CARTESIAN, UI_POLAR
+from hexrd.ui.create_hedm_instrument import create_hedm_instrument
 from hexrd.ui.create_raw_mask import apply_raw_mask, remove_raw_mask
 from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.ui_loader import UiLoader
@@ -70,6 +72,8 @@ class ImageModeWidget(QObject):
             HexrdConfig().set_polar_snip1d_numiter)
         HexrdConfig().instrument_config_loaded.connect(
             self.auto_generate_cartesian_params)
+        HexrdConfig().instrument_config_loaded.connect(
+            self.auto_generate_polar_params)
 
         self.ui.polar_show_snip1d.clicked.connect(self.polar_show_snip1d.emit)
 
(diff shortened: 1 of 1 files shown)


cjh1:
I will that a look a speeding this up.

joelvbernier:
The call to the `pixel_angles` property is the beast.  Could potentially memoize that so that it only updates on change of instrument parameters?  I am all for making this faster, as it affects other things as well.

*OR* we just calculate the ranges _once_ on loading an instrument config.  This should be good enough, as when changing parameters live in the GUI we aren't typically moving things that much, and certainly don't need to update the ranges and pixel size to that precision.  I think that would be the best option, IMHO.  The idea is to get parameters that are roughly appropriate for a given instrument config (_e.g._, the TARDIS goes from ~15-105 degrees, while HEDM instruments cover ~0-20).  The goal is to avoid having an empty polar plot when loading a new config.

joelvbernier:
Thoughts on my second note above?  Only calculate the polar (_and_ cartesian) view parameters on config load?  That would make it much less onerous...

psavery:
@joelvbernier Yep! I think that's how they are set up currently actually. It won't calculate them when you first start up the program (but just load the previously used values). It will only calculate if you either:
1. Load a new instrument config
2. Reset the instrument config

psavery:
@joelvbernier Okay, we can add that code here as well. This branch is currently [only computing the pixel sizes for polar](https://github.com/HEXRD/hexrdgui/pull/388/files#diff-30bfcc80b3692dfa8947f36bbec62f14R135-R156).

But we can add that code to this function to compute the tth min and max too.

joelvbernier:
edited my comment above...

joelvbernier:
Bump on this thread -- I'd say auto-calculating the extents are of primary concern; pixel sizes less so (see previous) until we can speed it up. Thoughts?

psavery:
I think this is working reasonably, and it is ready for review.

I encountered a bug with the polar image extents, but it is unrelated to this PR, and I'll report it in another issue.

joelvbernier:
Thanks guys.

PR #392, opened 2020-07-09
Add workflow configuration and selection

cjh1:
This PR as the concept of a workflow configuration. Currently we have two HEDM and LLNL ( name can be changed ). The user is asked on startup if one is not already set and can be changed by the Edit menu. The config is currently not used, but will be used to configure workflow specific part of the GUI. The is also a Signal that is fire when the workflow is changed.


hexrd/ui/workflow_selection_dialog.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def setup_connections(self):
+        self.ui.accepted.connect(self.accepted)
+        self.ui.rejected.connect(self.rejected)
+
+    def update_gui_from_config(self):
+        self.ui.workflowComboBox.setCurrentText(HexrdConfig().workflow)
+
+    def accepted(self):
+        HexrdConfig().workflow = self.ui.workflowComboBox.currentText()
```
psavery, line 29: FYI, we came to the realization that on some operating systems (I don't know which ones), the text in the combo box can be altered by Qt (it adds things like ampersands). This would cause problems if anyone uses one of these operating systems.

I've personally resorted to always using item data since then for reliable string comparisons. When you add an item to the combo box via `addItem` (above), you can do `addItem(w, w)` to add it as item data, and then here, you can call `self.ui.workflowComboBox.currentData()` instead.

hexrd/ui/workflow_selection_dialog.py
```diff
@@ -0,0 +1,36 @@
+from PySide2.QtCore import QObject
+
+from hexrd.ui.hexrd_config import HexrdConfig
+from hexrd.ui.ui_loader import UiLoader
+from hexrd.ui import constants
+
+
+class WorkflowSelectionDialog(QObject):
```
psavery, line 8: I think this only needs to inherit from `QObject`, by the way, if it emits signals, or if we need to install an event filter. Otherwise, it can be just a regular class.

hexrd/ui/workflow_selection_dialog.py
psavery: (you might need to modify `update_gui_from_config()` as well, to find the item data that matches)

joelvbernier:
Defer to you guys to get this merged.

hexrd/ui/workflow_selection_dialog.py
psavery: I take these comments back. The issue we ran into earlier had to do with the label for a QCheckBox. It appears that labels can have things (like ampersands) automatically inserted into them, but I haven't found any evidence that QComboBox items can be tampered with.

We can leave this the way it is.

Issue #391, opened 2020-07-08
Ensure Laue overlays are correct


psavery:
When you choose to display Laue overlays, *something* gets plotted. But I'm not sure if it's correct or not.

![laue_overlays](https://user-images.githubusercontent.com/9558430/86970235-105e4200-c13d-11ea-836a-11487c02adb1.png)

We should compare this to some real examples for the Raw, Cartesian, and Polar views, and ensure that they are truly correct.

Some things that could be wrong:
1. We could be in pixel coords instead of Cartesian, or vice versa
2. We could have swapped x and y
3. In the polar view, we may have radians/degree units wrong.
4. In the polar view, we might not have the same ranges. For instance, sometimes eta is -180 to 180, and sometime its 0 to 360. We should make sure we are using a consistent range.

Most changes to the Laue overlays will likely be made [here](https://github.com/HEXRD/hexrdgui/blob/ef0a8d1ad996c6f30f8e684bc815bec7ccaa9962/hexrd/ui/overlays/laue_diffraction.py#L73-L98) and [here](https://github.com/HEXRD/hexrdgui/blob/ef0a8d1ad996c6f30f8e684bc815bec7ccaa9962/hexrd/ui/image_canvas.py#L208-L221).

Once we confirm the Laue overlays are being plotted correctly for raw, cartesian, and polar, we can close this issue.

psavery:
This is fixed in #394 

An example dataset that we can use for testing the Laue spots was uploaded to our shared google drive, in "examples/simulated_laue".

Issue #381, opened 2020-07-07
Add masking via Laue overlays


psavery:
Laue overlays were added in #373, and widths for the Laue overlays in the polar view are added in #394.

We want to add an option to mask based upon the Laue overlay widths (which are rectangles in the polar view).

In the polar view, the overlay widths are stored as corners of the rectangle, and they are stored in this location:
```
HexrdConfig().overlays[<overlay_idx>]['data'][<detector_id>]['ranges']
```

Where `overlay_idx` is the index of the overlay, and `detector_id` is the id of the detector (note that in the Cartesian view, there will be only one fake detector. You can just use whichever one that one is).

You can check the type of the detector via:
```
HexrdConfig().overlays[<overlay_idx>]['type']
```

Where the type is `laue` for laue overlays.

Related: #395

Issue #31, opened 2019-07-03
Single Crystal Calibration


psavery:
We need to receive a single crystal example first, but we will need to make the calibrations work for single crystal data as well.

For single crystal data, the overlays will be dots instead of rings.

cjh1:
@joelvbernier Did you manage to get us an example?

psavery:
An example dataset that we can use for testing the Laue spots was uploaded to our shared google drive, in "examples/simulated_laue".

joelvbernier:
I will add an example this weekend.

cjh1:
@joelvbernier Any progress on an example?

psavery:
I think this has been fixed for a while, starting with #394.

PR #394, opened 2020-07-09
Add Laue overlay widths to the polar view

psavery:
This adds an option in the overlay editor to enable widths for the Laue
overlays, and it also adds options for tth_width and eta_width.

If this is enabled, these options get passed on to the Laue overlay
generator. For the polar view, these are used to construct range data,
which appear as rectangles around the Laue spots.

Currently, these options are ignored for the raw and cartesian views.
They will need to be added at a later time.

image

laue_widths

This also fixes issues with our plotting of the Laue spots.

Fixes part of: #31
Fixes part of: #381 

Fixes: #391

### DIFF
--- a/hexrd/ui/calibration/raw_iviewer.py
+++ b/hexrd/ui/calibration/raw_iviewer.py
@@ -39,9 +39,8 @@ class InstrumentViewer:
                 'plane_data': mat.planeData,
                 'instr': self.instr
             }
-            if type == 'laue':
-                # Modify kwargs here
-                pass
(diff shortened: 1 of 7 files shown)


joelvbernier:
seems good to me -- replied in other thread re: cartesian cases.

psavery:
I also pushed some commits to this PR to fix the Laue overlays for raw, cartesian, and polar views. You can see the current state in the attached.

laue_examples

This fixes #391

PR #397, opened 2020-07-09
Use unaggregated images for indexing if possible

psavery:
If there are unaggregated images set on the image load manager, use
those for the indexing.

This fixes an issues where, if the user had aggregated images, they
could not perform indexing. Now, the user can perform indexing
whether they aggregated the data or not.

I tested this with images loaded via the load panel, once with a maximum
aggregation, and once with no aggregation, and I was able to generate the
eta omega maps and perform the indexing in both cases.


joelvbernier:
We might get more explicitly with how we handle aggregated data...  will make an issue.

PR #401, opened 2020-07-10
Disable run-indexing menu until an image series is loaded

### LINKED ISSUE #374: Add validation for indexing steps
We should disable indexing by default and only enable it when:

1. Data has been loaded
2. There is more than one image in the unaggregated imageseries that was loaded

The above two are the only ones I know of, but we can add more if we find more.

johnkit:
Another tiny mod but all I have time for today. This branch disables the Run => Indexing menu at startup, and enables it after an image series has been loaded. This will fulfill #374 so we'll go ahead and close it, though more validation logic might become evident in the future.

Closes #374

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -142,6 +142,7 @@ class MainWindow(QObject):
         self.ui.action_run_indexing.triggered.connect(
             self.on_action_run_indexing_triggered)
         self.new_images_loaded.connect(self.update_color_map_bounds)
+        self.new_images_loaded.connect(self.update_indexing_menu)
         self.new_images_loaded.connect(self.color_map_editor.reset_range)
         self.new_images_loaded.connect(self.image_mode_widget.reset_masking)
         self.ui.image_tab_widget.update_needed.connect(self.update_all)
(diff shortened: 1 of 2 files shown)


hexrd/ui/main_window.py
```diff
@@ -626,3 +627,12 @@ def on_action_transform_detectors_triggered(self):
         self.image_mode_widget.reset_masking()
         td = TransformDialog(self.ui).exec_()
         self.image_mode_widget.reset_masking(mask_state)
+
+    def update_indexing_menu(self):
+        enabled = False
+        image_series_dict = HexrdConfig().imageseries_dict
```
psavery, line 633: Can this line be changed to these two lines instead?

```python
image_series_dict = ImageLoadManager().unaggregated_images
image_series_dict = HexrdConfig().imageseries_dict if image_series_dict is None else image_series_dict
```

Basically, we want to use unaggregated images if possible. But if it is `None`, then we use the `HexrdConfig().imageseries_dict` instead.

johnkit: Will do. just fyi: (a) I wasn't sure that the ImageLoadManager contents were persistent, and (b) I think that Hexconfig.imageseries_dict is loaded the way you've prescribed. (but I might have extrapolated too much.)

psavery: Yeah, it's a singleton as well. Its contents are persistent.

The complication is that, if the images were aggregated, we don't want to use `HexrdConfig().imageseries_dict`, but we want to use `HexrdConfig().unaggregated_images` instead. This follows the logic [here](https://github.com/HEXRD/hexrdgui/blob/5168755ee256cca68c9917793a67a134a29d7a12/hexrd/ui/indexing/create_config.py#L39-L44).

psavery:
I also gave this a try, and it worked very well. Thank you, John!

Issue #395, opened 2020-07-09
Add Laue overlay widths for raw and cartesian views


psavery:
Laue overlay widths are added to the polar view in #394 

The polar view is probably the easiest for this, because the widths result in rectangles. The Cartesian and raw views might be a little trickier, since two of the edges will likely be curved, if I'm imagining it correctly.

The code for creating the polar ranges is [here](https://github.com/HEXRD/hexrdgui/blob/03bfd7f26feacc8768ffc859864972f49e137c72/hexrd/ui/overlays/laue_diffraction.py#L110-L127). Similar code will need to be added for the raw and Cartesian views. Once the ranges data is added to the overlay data, they then get plotted [here](https://github.com/HEXRD/hexrdgui/blob/03bfd7f26feacc8768ffc859864972f49e137c72/hexrd/ui/image_canvas.py#L219-L222).

Laue overlay widths to do:
- [x] Cartesian
- [x] Raw

joelvbernier:
For the cartesian versions of the angular, here is an code snippet:

```
refl_patches = dict.fromkeys(instr.detectors)
for ip_key, ip in instr.detectors.items():
    patches = xrdutil.make_reflection_patches(
        ip.config_dict(chi=instr.chi, tvec=instr.tvec, beam_vector=instr.beam_vector),
        valid_angs[ip_key], ip.angularPixelSize(valid_xy[ip_key]),
        rmat_c=xfcapi.makeRotMatOfExpMap(expmap_c),
        tth_tol=tth_tol, eta_tol=eta_tol,
        npdiv=npdiv, quiet=True)
    refl_patches[ip_key] = patches
    ii = np.vstack([[patches[i][-1][0][0, 0],
                     patches[i][-1][0][0, -1],
                     patches[i][-1][0][-1, -1],
                     patches[i][-1][0][-1, 0],
                     patches[i][-1][0][0, 0]] for i in range(len(patches))])
    jj = np.vstack([[patches[i][-1][1][0, 0],
                     patches[i][-1][1][0, -1],
                     patches[i][-1][1][-1, -1],
                     patches[i][-1][1][-1, 0],
                     patches[i][-1][1][0, 0]] for i in range(len(patches))])
    fig2, ax2 = plt.subplots(1, 1, figsize=None)
    cmap = cm.bone_r
    cmap.set_under('b')
    sclimg = log_scale_img(log_scale_img(img_dict[ip_key]))
    vmin = np.percentile(sclimg, 5)
[...]

PR #404, opened 2020-07-13
Draw Laue widths for raw and cartesian views also

psavery:
Previously, Laue widths would only be drawn for polar views.

This draws them in raw and cartesian views as well.

raw

cartesian

Fixes: #395

### DIFF
--- a/hexrd/ui/overlays/laue_diffraction.py
+++ b/hexrd/ui/overlays/laue_diffraction.py
@@ -2,6 +2,8 @@ import numpy as np
 
 from hexrd import constants
 
+from hexrd.ui.constants import UI_RAW, UI_CARTESIAN, UI_POLAR
+
 
 class LaueSpotOverlay:
     def __init__(self, plane_data, instr,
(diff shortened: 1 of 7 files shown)


hexrd/ui/overlays/laue_diffraction.py
```diff
@@ -130,4 +133,31 @@ def polar_ranges(self, spots):
...
+            data = []
+            for i in range(len(corners) - 1):
+                tmp = np.linspace(corners[i], corners[i + 1])
+                data.extend(panel.angles_to_cart(tmp))
+
+            data = np.array(data)
+            if display_mode == 'raw':
+                data = panel.cartToPixel(data)
+                data[:, [0, 1]] = data[:, [1, 0]]
+            elif display_mode == 'cartesian':
```
cjh1, line 158: Lets extract out the display modes into constants

psavery: Done

Issue #402, opened 2020-07-10
Polar and Raw views resize when they draw out-of-bound overlays


psavery:
We should keep the extents of the polar view fixed, because they are set by the user in the image mode widget.

Keeping them fixed should prevent problems like this, when out-of-bound overlays are drawn:

![polar_out_of_bounds_overlays](https://user-images.githubusercontent.com/9558430/87202218-10417c00-c2ce-11ea-8e61-b19ce098788f.png)

PR #406, opened 2020-07-13
Turn off image canvas rescaling

psavery:
This prevents the raw and polar views from rescaling automatically when overlays are drawn outside of their bounds.

This has been happening lately especially with Laue overlays.

Fixes: #402


hexrd/ui/image_canvas.py
```diff
@@ -522,7 +524,7 @@ def update_azimuthal_integral_plot(self):
         axis.axis('auto')
 
     def on_detector_transform_modified(self, det):
-        if not self.iviewer:
+        if self.mode not in ['cartesian', 'polar']:
```
cjh1, line 527: Same here

PR #407, opened 2020-07-14
Add ability to mask polar view via Laue overlays

psavery:
If the polar view is displayed, and there are Laue overlays displayed
with widths (boxes are drawn around them), the user may mask the polar
view with these Laue widths via "Edit"->"Apply Laue Masks to Polar".

Note: there currently isn't a way to remove masks...

apply_laue_masks

Fixes: #381

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -127,6 +127,8 @@ class MainWindow(QObject):
             self.on_action_edit_euler_angle_convention)
         self.ui.action_edit_apply_polar_mask.triggered.connect(
             self.on_action_edit_apply_polar_mask_triggered)
+        self.ui.action_edit_apply_laue_mask_to_polar.triggered.connect(
+            self.on_action_edit_apply_laue_mask_to_polar_triggered)
         self.ui.action_edit_reset_instrument_config.triggered.connect(
             self.on_action_edit_reset_instrument_config)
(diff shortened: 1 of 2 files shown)


psavery:
@joelvbernier Do we want to be able to mask the Cartesian and Raw views as well? Or just the polar?

joelvbernier:
@psavery just the polar view for now, as the function of the masking is really to clean up the azimuthal integration...

psavery:
Okay. We are probably going to add a mask manager at some point, so the user can actually remove/edit masks that they have applied. But that's probably going to be in a separate PR.

I think this one is ready for review, @cjh1

hexrd/ui/main_window.py
```diff
@@ -486,6 +488,34 @@ def run_apply_polar_mask(self, line_data):
...
         self.update_all()
 
+    def on_action_edit_apply_laue_mask_to_polar_triggered(self):
+        if not HexrdConfig().show_overlays:
+            msg = 'Overlays are not displayed'
+            QMessageBox.critical(self.ui, 'HEXRD', msg)
+            return
+
+        all_overlays = HexrdConfig().overlays
+        laue_overlays = [x for x in all_overlays if x['type'] == 'laue']
```
cjh1, line 498: `laue` ideally should be a constant, but that can come in a follow up PR if necessary.

Issue #405, opened 2020-07-13
Overlays re-created too many times for raw tabbed view


psavery:
In the raw tabbed view, there is an image canvas for each raw image. Each image canvas [re-creates the overlays](https://github.com/HEXRD/hexrdgui/blob/ba9a96d7c8e168934c5de13f61334ca02e26cfa8/hexrd/ui/image_canvas.py#L234) when an `overlay_config_changed` signal is emitted from `HexrdConfig`.

The overlays only need to be recreated once. And, for some Laue overlays, it takes a while to generate them, so it is noticeably very slow if, for instance, the 8-detector example is used in the tab view, and the overlays are re-created once for each detector.

PR #408, opened 2020-07-14
Update overlays only when needed

psavery:
Rather than re-generating and re-drawing the overlays from scratch every time, only re-generate and re-draw them when they are needed. This provides a few benefits:

1. Significant speedup for generating and drawing a single overlay when multiple overlays are visible.
2. Big speedups for hide/show operations (the overlay data is not re-generated, but only re-drawn).
3. Big speedup in tabbed view (previously, each tab was re-generating the data from scratch).
4. More resilience against accidentally re-creating overlays multiple times in the code.

This PR additionally combines some common code into a function to reduce repetition.

It keeps track of the artists used for drawing via the id of the data that was used to to draw them.

Fixes: #405

### DIFF
--- a/hexrd/ui/image_canvas.py
+++ b/hexrd/ui/image_canvas.py
@@ -175,7 +175,6 @@ class ImageCanvas(FigureCanvas):
             self.overlay_draw_func(type)(axis, data, style)
 
     def draw_powder_overlay(self, axis, data, style):
-        print('Drawing powder...')
         rings = data['rings']
         rbnds = data['rbnds']
         rbnd_indices = data['rbnd_indices']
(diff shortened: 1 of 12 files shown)


hexrd/ui/image_canvas.py
```diff
@@ -168,10 +182,12 @@ def draw_powder_overlay(self, axis, data, style):
         data_style = style['data']
         ranges_style = style['ranges']
 
+        artists = []
+        self.overlay_artists[id(data)] = artists
         for pr in rings:
             x, y = self.extract_ring_coords(pr)
             artist, = axis.plot(x, y, **data_style)
```
cjh1, line 189: This an instance of `Line2D` a subclass of artist?

psavery: Yep! It is a subclass of `Artist`.

Issue #399, opened 2020-07-10
Use detector perimeter information for cropping


joelvbernier:
In the shared Google Drive, I have uploaded detector perimeters for the TARDIS instrument (for now we will only deal with IMAGE-PLATE-[1,2]; IMAGE-PLATE-3 is cylindrically bent and I have to get that detector type finished).  The image data in the assoc. hdf5 archives is scanned into a 750 x 1250 array (100µm pixels), which is much bigger than the actual film.  Here is a view of ip4:
<img width="1680" alt="Screen Shot 2020-07-10 at 8 58 04 AM" src="https://user-images.githubusercontent.com/1154130/87174120-98526200-c28b-11ea-839a-015c28745862.png">
And here is a plot of the boundary, which is in physical coordinates (mm):
<img width="752" alt="Screen Shot 2020-07-10 at 9 01 17 AM" src="https://user-images.githubusercontent.com/1154130/87174431-01d27080-c28c-11ea-9037-7bed5e10db0e.png">
You can already see that a flip (mirror) about the horizontal axis is required in the image load.

What we need is the ability to translate and rotate (preferrably with mouse imput, but 3 sliders would work fine too) this contour on the raw image so the user can set up the the crop operation.  The actual cropping area with be the rectangle that circumscribes the image plate boundary.  The mask (valid values) would be the intersection of said rectangle and the area inside the image plate boundary.  Sound good?  Have fun :-)

bnmajor:
This is great, thanks @joelvbernier!

joelvbernier:
I'll get perimeters for the PXRDIP and BBXRD up as well -- those are the cases where all of the image plates (read: detectors) are scanned into a single big image, and the user will have to specify the crop for each detector's image sequentially.  The TARDIS case is easier in that there is one file per detector.

bnmajor:
@joelvbernier Awesome, thank you. At this point I will proceed with the plan outlined in #384 to create a new load panel that will apply to this workflow, unless you see any problems with the current ideas.

joelvbernier:
@bnmajor yeah, Let's call it an "Import Data" panel.  At this point, we could just have profiles for the different instruments (might be easiest way?).

bnmajor:
@joelvbernier What will the config file look like for this use case? I'm curious what may be missing or need to be updated (like the rows and columns for each detector) after the cropping and detector masking has been applied.

joelvbernier:
[...]

PR #410, opened 2020-07-14
LLNL IO

bnmajor:
This branch addresses the needs discussed in #399


bnmajor:
Summary of current features and capabilities on this branch:

There is now an `Import Data Panel` available on the left side of the GUI, under the `Load Panel` tab. This will be updated to only be available for the LLNL workflow.

association

The association group requires that you first select the instrument that you would like to work with, and a default config file for that instrument is loaded. Once the instrument has been selected the detectors list will be populated with the available detector shapes. Selecting a detector will enable file load and disable instrument selection. Setting the detector selection back to `None` will enable instrument selection and disable file loading.

data_load

Once a file has been loaded the `Detector Outline` group will be enabled so that the template can be added to the image. This currently only supports loading a single image.

detector_outline

Clicking `Add Detector Outline` will draw the detector outline on the image and disable the add option but enable the options for positioning and accepting or rejecting the outline placement.
The `Translate` and `Rotate` options can be toggled between in order to position the outline, using the mouse to drag it into place.
Once the outline is where it should be, pressing `OK` will mask the area between the outline and the detector border and then crop the image down to the size of the template. The image is updated in the canvas, but the config file is not currently updated to reflect the new rows and columns so it will appear skewed.
If `Cancel` is selected the template will be removed and detector selection, file loading, and the option to add the outline will be enabled again.
Selecting `Save` will allow you to save the edited image as an `ImageSeries`.

bnmajor:
At this point only one image can be loaded at a time and in order to view all of the new images the correct config file for the instrument (with each of the detectors) must be loaded and the edited images must be saved once the position has been accepted so that they can be re-loaded with the config. This process needs to be handled automatically.

Needed:
- [ ] Allow loading of multiple images
- [ ] Update config file with new dimensions after crop
- [ ] Provide way to indicate all detectors have been completed
- [ ] On completion re-load data with updated configuration file and cropped and masked images
- [ ] Need detector outlines for `BBXRD` and `PXRDIP` detectors
- [ ] Need to determine appropriate default configurations for each detector (currently only have a best-guess for `TARDIS`) 
- [ ] (Optional) Add sliders with rotate and translate options to allow fine-tuning positioning

hexrd/ui/hexrd_config.py
```diff
@@ -716,6 +716,10 @@ def default_detector(self):
         return copy.deepcopy(
             self.default_config['instrument']['detectors']['ge1'])
 
+    def detector_pixel_size(self, detector_name):
+        detector = self.detector(detector_name)
+        return detector.get('pixels', {}).get('size', {}).get('value', [1, 1])
```
joelvbernier, line 721: @bnmajor is [1, 1] here the default pixel [row, col] size?  If so, we can make it [0.1, 0.1] to be more appropriate for the LLNL data.

bnmajor: Sounds good, it has been updated.

hexrd/ui/image_file_manager.py
```diff
@@ -75,11 +75,19 @@ def load_aps_imageseries(self, detectors, directory_names):
                 return
 
     def open_file(self, f):
-        ext = os.path.splitext(f)[1]
-        if self.is_hdf5(ext):
-            ims = imageseries.open(f, 'hdf5',
-                path=self.path[0],
-                dataname=self.path[1])
+        ext = os.path.splitext(f)[1] if isinstance(f, str) else None
```
cjh1, line 78: What other type can `f` be?

hexrd/ui/import_data_panel.py
```diff
@@ -0,0 +1,247 @@
...
+                self.ui.detector_label.setEnabled(True)
+                self.ui.detectors.setEnabled(True)
+
+    def get_instrument_detectors(self, instrument):
+        self.mod = resource_loader.import_dynamic_module(
+            'hexrd.ui.resources.templates.' + instrument)
+        contents = resource_loader.module_contents(self.mod)
+        dets = ['None']
+        for content in contents:
+            if isinstance(content, str) and not content.startswith('__'):
```
cjh1, line 69: What other type are we dealing with here?

hexrd/ui/import_data_panel.py
```diff
@@ -0,0 +1,247 @@
...
+        else:
+            self.it.disconnect_rotate()
+        img = self.it.get_mask()
+        bounds = np.array([int(val) for val in self.it.crop()]).reshape(2, 2)
+        self.finalize(img, bounds)
+        self.completed_detectors.append(self.ui.detectors.currentText())
+        self.ui.trans.setDisabled(True)
+        self.ui.rotate.setDisabled(True)
+        self.ui.button_box.setDisabled(True)
+        self.ui.add_template.setDisabled(True)
```
cjh1, line 162: We might be able to make this more concise with loop?

```python
for w in [self.ui.trans, self.ui.rotate, self.ui.button_box, self.ui.add_template]:
    w.setDisabled(True)
```

hexrd/ui/image_file_manager.py
bnmajor: `f` may not be a file at all but instead a `numpy` array that needs to be converted to an image series. [Here](https://github.com/HEXRD/hexrdgui/blob/ebbc5e368788e2ba4b9a1480a4a1669044890624/hexrd/ui/import_data_panel.py#L173) the masked image is passed as an array an an image series is created from the processed raw image (and then cropped). Maybe this would be better factored out into its own function?

hexrd/ui/import_data_panel.py
bnmajor: I believe that I originally added the check because `importlib.resources.contents` can also return directory objects, but since there would be no reason for sub-directories to be present in the templates directories this check can be removed.

hexrd/ui/import_data_panel.py
bnmajor: There are a couple of places where multiple widgets are being enabled/disabled at once (like [here](https://github.com/HEXRD/hexrdgui/blob/ebbc5e368788e2ba4b9a1480a4a1669044890624/hexrd/ui/import_data_panel.py#L189-L196), [here](https://github.com/HEXRD/hexrdgui/blob/ebbc5e368788e2ba4b9a1480a4a1669044890624/hexrd/ui/import_data_panel.py#L105-L111), and [here](https://github.com/HEXRD/hexrdgui/blob/ebbc5e368788e2ba4b9a1480a4a1669044890624/hexrd/ui/import_data_panel.py#L126-L132)). This loop could be factored out into a function and reused, unless that may make things more confusing? For example:
```
def set_enabled(self, widgets, enabled):
  for w in widgets:
    w.setEnabled(enabled)

self.setEnabled([self.ui.trans, self.ui.add_template,
                 self.ui.button_box, self.ui.rotate], False)
```

hexrd/ui/image_file_manager.py
cjh1: OK, lets just add a comment indicating the the two possible types.

hexrd/ui/import_data_panel.py
cjh1: Perfect, even better!

cjh1:
@bnmajor Can you point me to a hdf4 dataset to test this with?

bnmajor:
Yes, in the shared drive under `sample_images/LLNL/PXRDIP` there is an hdf4 file called `pxrdip_s77774_pxrdip1_-1-[phosphor].hdf`

cjh1:
@bnmajor Thank you

joelvbernier:
Ah yes, I forgot LLE still writed their data to HDF4...

cjh1:
@joelvbernier Are currently plan is to support in the conda environment ( as we can easily install pyhdf ) and then fail gracefully else where.

joelvbernier:
that is great!

hexrd/ui/load_hdf5_dialog.py
```diff
@@ -17,16 +17,38 @@ def __init__(self, f, saved_path=None, parent=None):
...
+  def get_HDF4_paths(self, f):
+    try:
+      from pyhdf.SD import SD, SDC
+      img = SD(f, SDC.READ)
+      self.file = img
+      self.paths = [p for p in img.datasets()]
+    except (Exception, IOError) as error:
+      msg = ('ERROR - This installation does not support '
+              + 'this file type.')
+      QMessageBox.warning(None, 'HEXRD', msg)
```
cjh1, line 29: Can we pass the parent here, it should then center on app then.

cjh1:
@bnmajor I just tested the hdf4 logic, works well, nice!

cjh1:
:tada:

Issue #379, opened 2020-07-07
Fixes for Distortion of None


psavery:
For LLNL, a distortion of `None` will most often be used. When the distortion is `None`, there shouldn't be parameters for the distortion function in the tree view. There will be a few things that need to be done.

- [x] Remove the distortion function parameters from [the default instrument config](https://github.com/HEXRD/hexrdgui/blob/89d70ded426eee1e9993ac7e6de90d90d7a0e86c/hexrd/ui/resources/calibration/default_instrument_config.yml#L8-L9), so they don't get set automatically when a config file is loaded
- [x] When the distortion is selected to be `None` via the form view (see below), remove the distortion parameters from the `HexrdConfig` instrument config, so they do not appear in the tree view
- [x] When the distortion is changed from `None` to something else, add distortion parameters to the `HexrdConfig` instrument config

It appears that the form view already hides the parameters when `None` is chosen for the distortion function, which is correct. Now, we'd just need to get it to remove the parameters from the internal instrument config, and then add them back in if if the user switches to a different distortion other than `None`.
![distortion_none](https://user-images.githubusercontent.com/9558430/86833075-08cc6980-c067-11ea-856c-9338c65f5a2d.gif)

[This function](https://github.com/HEXRD/hexrdgui/blob/9f1923038ae5dc493f4665d1593c0b9507316067/hexrd/ui/hexrd_config.py#L1211-L1218) might be useful for this.

cjh1:
Thanks

PR #415, opened 2020-07-15
Update default_instrument_config.yml

joelvbernier:
Small edit to make it represent a single GE/Perkin-Elmer detector with Re absorption edge energy.

Fixes: #379

### DIFF
--- a/hexrd/ui/cal_tree_view.py
+++ b/hexrd/ui/cal_tree_view.py
@@ -1,4 +1,4 @@
-from PySide2.QtCore import QModelIndex, Qt
+from PySide2.QtCore import QObject, QModelIndex, Qt
 from PySide2.QtWidgets import (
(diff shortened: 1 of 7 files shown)


psavery:
I added some changes that fix the distortion dict issues. Now, when it is `None`, the distortion dict is removed and will not show up in the tree view. This fixes #379 as well.

The form view now also correctly recognizes that a a missing distortion dict means it is `None`.

joelvbernier:
I have a feature branch to finally fix the kludge that is the distortion handling.  We will have to edit the configuration forms accordingly. In brief, I am replacing it with a simple class, and they’ll be a registry with a couple of examples. So I envision in the future, the distortion can either be absent, or populated according to the selection from the registry.  This will be necessary, because different choices of functions will have different numbers of parameters with different names.

Issue #387, opened 2020-07-08
Input calibration crystal stretch values


psavery:
Currently, in the overlay manager added in #373, the user inputs stretch values as inverse stretch values into the calibration crystal parameters:
![image](https://user-images.githubusercontent.com/9558430/86951027-f9a8f280-c11e-11ea-9b18-52b6baf10db6.png)

As noted in #92 ([this comment](https://github.com/HEXRD/hexrdgui/issues/92#issuecomment-584807140)), it would be better to have the user input stretch values instead, and then we will compute the inverse stretch values from those.

psavery:
Note that for the `Calibration Crystal` section, the 12 spin boxes are set/get as a list of 12 numbers (this is what gets passed directly to hexrd). The getter/setter for this are [here](https://github.com/HEXRD/hexrdgui/blob/5d7484027daa416cd41214b65d4984d1afb2e483/hexrd/ui/overlay_editor.py#L142-L161).

We can put more code in there alongside the euler angle convention conversion code, that essentially converts from `inverse stretch` to `stretch` in the setter, and converts from `stretch` to `inverse stretch` in the getter. This way, the `stretch` values are only displayed in the GUI, but the actual values that are stored and provided to hexrd are the `inverse stretch` values.

joelvbernier:
Lets go one further:  Let's lay the stretch out in matrix form, where the 6 values make up the diagonal and upper triangle of the symmetric matrix (3 x 3).  The components 0-5 map to: 
```
[0]    [5]    [4]
[5]    [1]    [3]
[4]    [3]    [2]
```
Then we can internally take the inverse (and the off diagonal components get a scaling of √2 as well).

psavery:
So, the stretch matrix off-diagonal components get a scaling of √2, and then it gets inverted. Is that right? Or is the scaling after the inversion?

joelvbernier:
inversion, then scaling.  You can use the following utility from hexrd to take the unscaled V components to the scaled inv(V) that the code needs in the grain parameters:
```
from hexrd import matrixutil

vinv_vecs = matrixutil.symmToVecMV(np.linalg.inv(v_matx), scale=True)
```
where `v_matx` is symmetric, (3, 3) and `vinv_vecs` are the 6 values that go into the grain parameter list for the Laue simulation.

joelvbernier:
[...]

PR #419, opened 2020-07-16
Add calibration crystal editor with stretch matrix

psavery:
This separates out the calibration crystal editor as a stand-alone
widget, because it will be re-used for the mono rotation series.

Additionally, the stretch matrix is used instead of an inverse
stretch matrix. This is automatically converted to inverse stretch
values behind the scenes, which then get passed to hexrd.

If the user inputs values for a stretch matrix that is a noninvertible
(singular) matrix, the matrix boxes appear as red, and the values
are not set. But once the matrix is invertible, the values will
appear as white again.

Additionally, values across the diagonal are kept identical in the
stretch matrix automatically.

calibration_crystal_editor

Fixes: #387

### DIFF
--- a/hexrd/ui/calibration_crystal_editor.py
+++ b/hexrd/ui/calibration_crystal_editor.py
@@ -1,0 +1,212 @@
+import copy
+import numpy as np
+from numpy.linalg import LinAlgError
+
+from PySide2.QtCore import QObject, QSignalBlocker, Signal
+
+from hexrd.ui.hexrd_config import HexrdConfig
+from hexrd.ui.ui_loader import UiLoader
+from hexrd.ui.utils import convert_angle_convention
+
+
+class CalibrationCrystalEditor(QObject):
+
+    # Emitted when the params get modified
+    params_modified = Signal()
+
+    def __init__(self, params=None, parent=None):
+        super().__init__(parent)
+
+        loader = UiLoader()
+        self.ui = loader.load_file('calibration_crystal_editor.ui', parent)
+
+        self.params = params
+
+        self.update_gui()
+        self.update_orientation_suffixes()
+
+        self.setup_connections()
+
+    def setup_connections(self):
+        HexrdConfig().euler_angle_convention_changed.connect(
+            self.euler_angle_convention_changed)
+
+        for w in self.all_widgets:
+            w.valueChanged.connect(self.value_changed)
+
+    @property
+    def params(self):
+        return self._params
+
+    @params.setter
+    def params(self, v):
+        self._params = copy.deepcopy(v)
+        self.update_gui()
+
+    def value_changed(self):
+        sender = self.sender()
+
+        if sender in self.orientation_widgets:
+            self.params[:3] = self.orientation
+        elif sender in self.position_widgets:
+            self.params[3:6] = self.position
+        else:
+            # If the stretch matrix was modified, we may need to update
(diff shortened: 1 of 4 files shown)


psavery:
Note: I just also made a change so that the off-diagonal components are scaled by square root of two.

We should come up with an example where we can verify that I am doing the math correctly....

joelvbernier:
You see my comment in #387?  The utility method to do the scaling upon conversion exists (see code snippet).

psavery:
Okay. I am now using the utility functions from hexrd. I think it should hopefully be correct now, although we probably want to get confirmation at some point with an example.

hexrd/ui/calibration_crystal_editor.py
```diff
@@ -155,28 +157,15 @@ def position(self, v):
...
-
-        return scaling
-
     @property
     def inverse_stretch(self):
-        m = (self.stretch_matrix * self.matrix_scaling).reshape(3, 3)
-        m = np.linalg.inv(m).flatten()
-        indices = [0, 4, 8, 5, 2, 1]
-        return [m[i] for i in indices]
+        m = np.array(self.stretch_matrix).reshape(3, 3)
```
joelvbernier, line 162: self.stretch matrix is the raw user input?

hexrd/ui/calibration_crystal_editor.py
```diff
@@ -155,28 +157,15 @@ def position(self, v):
...
+        return matrixutil.symmToVecMV(np.linalg.inv(m), scale=True)
 
     @inverse_stretch.setter
     def inverse_stretch(self, v):
-        indices = [0, 5, 4, 5, 1, 3, 4, 3, 2]
-        m = np.array([v[i] for i in indices]).reshape(3, 3)
-        m = np.linalg.inv(m).flatten()
-        self.stretch_matrix = m / self.matrix_scaling
+        m = matrixutil.vecMVToSymm(v, scale=True)
```
joelvbernier, line 167: Have to be careful about what we are using in this setter.  If the 6-vector is the the same as the values that go into the grain parameters, then the scaling is already in there and the "scale" kwarg should be False.  What is the use case you were envisioning?

hexrd/ui/calibration_crystal_editor.py
psavery: Yes, `self.stretch_matrix` returns a flattened list of the matrix values from the GUI, with indices `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`, where the 3D shape is:
```
0 1 2
3 4 5
6 7 8
```

psavery: It then gets reshaped by numpy into a 3x3 matrix.

hexrd/ui/calibration_crystal_editor.py
psavery: This setter for `inverse_stretch` is supposed to be the opposite of the getter for `inverse_stretch`. This setter is where we take the 6-vector and convert it into the raw matrix that gets displayed in the GUI.

So, the `inverse_stretch` getter converts the raw matrix into scaled 6-vector.
The `inverse_stretch` setter converts the scaled 6-vector into the raw matrix.

psavery:
@cjh1 Done
image

cjh1:
@joelvbernier I think @psavery addressed you comments, are we good to go?

PR #422, opened 2020-07-17
Use an enum class for overlay types

psavery:
This might be better than using strings...


cjh1:
@johnkit Can you test this on your Mac? Just want to make sure we don't run into the same issue.

joelvbernier:
Is this related to #426 ?

PR #431, opened 2020-07-22
Do not disable dark subtraction for single images

### LINKED ISSUE #424: Load data options become inactive after file read
am trying to load a single image for which I have a standalone dark image (same GE format).  I select "file" for the dark mode and select the filename, but when I read the image, it toggles to None and the buttons become inactive.  see below.
Screen Shot 2020-07-19 at 10 41 55 AM
Loading single frames is not what this dialog is really meant for, but still, this is not the correct behavior.

bnmajor:
Fixes #424

### DIFF
--- a/hexrd/ui/load_panel.py
+++ b/hexrd/ui/load_panel.py
@@ -234,14 +234,11 @@ class LoadPanel(QObject):
         total_frames = np.sum(self.total_frames)
         if total_frames - self.empty_frames < 2:
             enable = False
-        self.ui.darkMode.setEnabled(enable)
         self.ui.aggregation.setEnabled(enable)
 
         if not enable:
             # Update dark mode settings
             num_dets = len(HexrdConfig().detector_names)
-            self.state['dark'] = [5 for x in range(num_dets)]
-            self.ui.darkMode.setCurrentIndex(5)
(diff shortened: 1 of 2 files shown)


bnmajor:
When a single image is loaded through the load panel the only options for dark subtraction are now `File` or `None`.

hexrd/ui/load_panel.py
```diff
@@ -234,14 +234,16 @@ def enable_aggregations(self, row, column):
...
         self.ui.aggregation.setEnabled(enable)
+        for i in range(4):
+            self.ui.darkMode.model().item(i).setEnabled(enable)
 
         if not enable:
             # Update dark mode settings
-            num_dets = len(HexrdConfig().detector_names)
-            self.state['dark'] = [5 for x in range(num_dets)]
-            self.ui.darkMode.setCurrentIndex(5)
+            if self.ui.darkMode.currentIndex() < 4:
```
cjh1, line 243: Lets use the constants, that we have defined [here](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/constants.py#L9-L14)

bnmajor: Fixed

PR #438, opened 2020-07-28
File upload paths

### LINKED ISSUE #414: Need to unify approach for associating image filename with detector names
I am still having the occasional issue with image loading.  Currently, I have a single-detector instrument with the detector name `ge5`; I get an error when I try to load `CeO2_0pt3s_000024.ge5` with the Load Data panel, but I can load it with the File --> Open --> Images dialog.

Is there a reason why we can't simply use the `in` (or `find`, or `count`) built-in to detect the occurrence of the detector name in the input image filenames?  There seems to be no need to mess with different regex for our use case, all we need to know is if the detector name string is anywhere in the image filename; stem, prefix, suffix, whatever.  Hell, we could even have a selector for ignoring case sensitivity.

At any rate, the imports should both behave the _same_ (there is still the need to handle the special case where the images get stored to different directory names that match the detector names, which I believe the "APS Imageseries" button does?) and need to be more robust.

High priority fix.

bnmajor:
Fixes #414

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -2,6 +2,7 @@ import copy
 import functools
 import os
 import time
+import glob
 from concurrent.futures import ThreadPoolExecutor
 
 from hexrd import imageseries
@@ -45,7 +46,6 @@ class ImageLoadManager(QObject, metaclass=Singleton):
         manual = False
         if not files:
             files = self.match_files(fnames)
-            print('match files: ', files)
             matched = self.check_success(files)
             if not matched:
                 manual = True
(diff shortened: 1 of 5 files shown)


bnmajor:
This approach allows the user to load files with quick load (through the `File` menu) or through the load panel (part of the `HEDM` workflow). Either way files can be selected either:
1. Explicitly - the user will select all of the files that they need for all of the detectors. In the case of the quick load path this is still limited to one file per detector. In the case of the load panel this can still be one or more files per detector.
2. Implicitly - the user can select the file(s) that they would like for one detector and the equivalent files will be found for the remaining detectors. This will search the current directory first to see if all files exist there. If they are not found in the current directory, sibling directories will be searched next. This was the method previously used when the `Use Subdirectories` option was selected in the load panel. That option has been removed now.
3. Manually - the user can select all of the files that they would like to load for all of the current detectors. If the files cannot be matched to detectors by name (the explicit check) or by search pattern (the implicit check), the user will be provided the load images confirmation dialog and will be able to set image and detector association manually.
manual_association

hexrd/ui/image_load_manager.py
```diff
@@ -154,10 +162,11 @@ def process_ims(self, postprocess, update_progress):
...
                 for i, det in enumerate(det_names):
-                    if self.data is None:
-                        dirs = self.parent_dir
-                    elif 'directories' in self.data:
-                        dirs = self.data['directories'][i]
+                    # if self.data is None:
+                    dirs = self.parent_dir
+                    # elif 'directories' in self.data:
+                    #     print('directories: ', self.data['directories'])
+                    #     dirs = self.data['directories'][i]
```
cjh1: let remove the commented out code.

hexrd/ui/image_load_manager.py
```diff
@@ -40,70 +41,77 @@ def __init__(self):
...
+        # and at least one file per detector
+        if (not files[0]
+                or len(files) != len(detectors)
+                or any(len(files[0]) != len(elem) for elem in files)):
+            return False
+        # If the files do not contain detector names they will need to
+        # be manually matched
+        if not all(any(d in f for d in detectors) for f in files[0]):
+            return False
+        return files
```
cjh1, line 67: It seems that this function either returns the `files` or `False`? May be it should just return `True` or `False`?

hexrd/ui/image_load_manager.py
bnmajor: Removed

hexrd/ui/image_load_manager.py
bnmajor: Fixed

hexrd/ui/image_load_manager.py
```diff
@@ -40,70 +41,80 @@ def __init__(self):
...
+        # and at least one file per detector
+        if (not files[0]
+                or len(files) != len(detectors)
+                or any(len(files[0]) != len(elem) for elem in files)):
+            return False
+        # If the files do not contain detector names they will need to
+        # be manually matched
+        if not all(any(d in f for d in detectors) for f in files[0]):
+            return False
+        return True
```
cjh1, line 67: I think we can now simplify this to:

`return all(any(d in f for d in detectors) for f in files[0])`

bnmajor: Sounds good - fixed.

PR #439, opened 2020-07-29
Dark sub

### LINKED ISSUE #424: Load data options become inactive after file read

bnmajor:
Dark subtraction uses functions from the `stats` package in `HEXRD` and requires an `ImageSeries` object to operate on. Typically this is the loaded `ImageSeries`, but in the case of dark subtraction using a file the selected file is what needs to be used. This branch fixes that appropriate `ImageSeries` is used in each case.

During testing I noticed that files could be loaded even when the `File` option was selected for dark subtraction but no file had been selected. This branch also fixes that by disabling the `Read Files` option until a dark file is loaded.

Fixes #424

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -217,7 +217,7 @@ class ImageLoadManager(QObject, metaclass=Singleton):
         else:
             f = imageseries.stats.median_iter
 
-        return (f, frames)
+        return (f, frames, ims)
 
     def get_dark_aggr_ops(self, ims_dict):
         """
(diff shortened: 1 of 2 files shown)


cjh1:
Thanks for fixing that you can go ahead and merge.

PR #440, opened 2020-07-29
fix to use mask from projection method

joelvbernier:
was causing an index error for detectors with invalid projections removed.


joelvbernier:
I rebased this feature branch before adding my fix, which looks like it triggered a bunch o fmessages from pep8speaks; you want to fix those too?

cjh1:
@joelvbernier The reason you are getting all the pep8 issues is it seems that you added bunch of duplicate files to ` build/lib/`? I am assuming this was by mistake?

joelvbernier:
crap, yes -- that was not the intent.  Anyhow, must fix this because the TARDIS example won't run inpolar view without it.

cjh1:
@joelvbernier Would you like me to try a clean it up?

cjh1:
@joelvbernier I took the liberty of creating #444, I cherry picked what I think the two changes you made here on to a clean master. Let me know if that looks good to you.

PR #441, opened 2020-07-30
Multi file load

bnmajor:
This fixes a bug created by #438 where the parent directory was no longer saved when images were loaded through the load panel. This also removes an used code path that should have been removed in that same PR


bnmajor:
@joelvbernier @cjh1 With the removal of the `Use Subdirectories` option the `Change Images Folder` no longer serves much of a purpose. Unless there is a use for the option that I'm overlooking I suggest removing the option altogether.
file_reader

cjh1:
I agree do you concur @joelvbernier ?

PR #444, opened 2020-07-31
Polarview fixes cherry pick

cjh1:
This PR contains the changes included in #440, but I have cherry-picked the two changes onto a clean master, thus removing the files added to the build directory.


joelvbernier:
@cjh1 -- how would I have changed those in the commit?  It certainly wasn't intentional.  Asking to avoid it in future PR's.

cjh1:
Not sure how it happened, did you do a `git add -A` that might have add the `build/` directory.

PR #445, opened 2020-08-04
Add support for saving and load configs with numpy arrays

cjh1:
This PR builds on HEXRD/hexrd#64 to support numpy arrays inside the YAML configuration. The numpy arrays are serialized in files store relative to the YAML file. The problem with this multi-file approach is that it hard for the user to know exactly which files are associated with a particular configuration, making things like share a configuration a little tricky. To overcome this we borrow from [Exdir](https://exdir.readthedocs.io/en/latest/) and define a instrument configuration as a directory that can contain all the files associated with a configuration. At the root of the directory there is the original YAML configuration name `config.yml`, sub paths store any associated numpy arrays.

As part of this new menu entries have been added to Save/Open configurations using the new directory format, support for save or opening without the directory is still supported. Its likely if we adopt the directory format, we can remove the old entry.


hexrd/ui/main_window.py
```diff
@@ -217,25 +222,84 @@ def add_materials_panel(self):
...
-            HexrdConfig().load_instrument_config(selected_file)
+            path = Path(selected_file)
+            HexrdConfig().working_dir = str(path.parent)
+
+            HexrdConfig().load_instrument_config(str(path))
             self.update_config_gui()
 
-    def on_action_save_config_triggered(self):
+    def on_action_open_config_dir_triggered(self):
+        dialog = QFileDialog(self.ui, 'Save Configuration',
```
psavery, line 239: This should probably be something like `Load Configuration` instead of `Save Configuration` here, right?

cjh1: Good catch!

cjh1: Fixed

hexrd/ui/main_window.py
```diff
@@ -217,25 +222,97 @@ def add_materials_panel(self):
...
+            HexrdConfig().working_dir = str(Path(selected_file).parent)
             return HexrdConfig().save_instrument_config(selected_file)
 
+    def on_action_save_config_dir_triggered(self):
+        dialog = QFileDialog(self.ui, 'Save Configuration',
+                             HexrdConfig().working_dir,
+                             'HEXRD directories (*)')
+        dialog.setFileMode(QFileDialog.Directory)
+        dialog.setAcceptMode(QFileDialog.AcceptSave)
+        dialog.setOptions(QFileDialog.ReadOnly)
```
psavery, line 292: Unfortunately, it looks like it is still creating the non-suffixed directory for me. :slightly_frowning_face: 

If we can't stop it from doing that, I guess we can just not worry about the suffix.

cjh1: it is? I will check again

cjh1:
@psavery I added a hack to work around `QFileDialog`'s behavior, let know what you think!

psavery:
@cjh1 This works for me.

Issue #385, opened 2020-07-08
Add mask manager component


cjh1:
Should allow:

- Adding
- Removing
- Turning on and off

psavery:
This might be related to #380

cjh1:
We will need to take into account the masks being added by #407

cjh1:
@bnmajor Is going to look at this once we have IO work done.

joelvbernier:
got it.

cjh1:
I think once we have most of the templating functionality we will better understand the requirements for the manager component.

bnmajor:
Added in #446

PR #446, opened 2020-08-04
Create a basic mask management class

bnmajor:
This branch adds a dialog for mask management. Masks are added to a table as they are created and can be toggled on or off, removed, or renamed.

Addresses #385

### DIFF
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -22,6 +22,8 @@ class MaskManagerDialog(QObject):
         self.create_masks_list()
         self.threshold_name = ''
 
+        self.setup_connections()
+
     def show(self):
         self.setup_table()
         self.ui.show()
@@ -49,6 +51,9 @@ class MaskManagerDialog(QObject):
             self.masks['mask_' + str(len(data) - 1)] = data[-1]
             self.visible['mask_' + str(len(data) - 1)] = data[-1]
         self.setup_table()
+    def setup_connections(self):
+        self.ui.masks_table.cellDoubleClicked.connect(self.get_old_name)
+        self.ui.masks_table.cellChanged.connect(self.update_mask_name)
 
     def setup_table(self, status=True):
         self.ui.masks_table.setRowCount(0)
@@ -89,3 +94,18 @@ class MaskManagerDialog(QObject):
 
     def remove_mask(self):
         return
+
+    def get_old_name(self, row, column):
+        if column != 0:
+            return
+
+        self.old_name = self.ui.masks_table.item(row, 0).text()
+
+    def update_mask_name(self, row, column):
+        if not hasattr(self, 'old_name') or self.old_name is None:
+            return
+
+        new_name = self.ui.masks_table.item(row, 0).text()
+        if self.old_name != new_name:
+            self.masks[new_name] = self.masks.pop(self.old_name)
+        self.old_name = None
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -89,10 +89,11 @@ class MaskManagerDialog(QObject):
         checkbox.toggled.connect(HexrdConfig().set_threshold_mask_status)
 
     def toggle_visibility(self, checked):
-        name = self.threshold_name
-        if self.ui.masks_table.currentRow() > 0:
-            row = self.ui.masks_table.currentRow()
-            name = self.ui.masks_table.item(row, 0).text()
+        if self.ui.masks_table.currentRow() < 0:
+            return
+
+        row = self.ui.masks_table.currentRow()
+        name = self.ui.masks_table.item(row, 0).text()
         if checked and name and name not in self.visible.keys():
             self.visible[name] = self.masks[name]
         elif not checked and name in self.visible.keys():
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -119,7 +119,7 @@ class MaskManagerDialog(QObject):
         if item.text() == self.threshold_name:
             self.reset_threshold()
         HexrdConfig().polar_masks_line_data = (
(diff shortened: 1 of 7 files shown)


bnmajor:
The actual table for managing the masks could go in a couple of locations - 

Next to the Image Mode and Color Map widgets:
mask_manager_main_window

As another tab in the Image Mode widget:
image_mode_masks

Or as a stand-alone dialog opened from the `Edit` menu:
mask_manage_dialog


Or if there is another location that seems like it would flow best or be the most practical let me know and we can use that instead.

bnmajor:
mask_manager

joelvbernier:
Very nice work @bnmajor I like it!

cjh1:
@bnmajor How hard would it be to get the threshold mask in there too?

bnmajor:
@cjh1 Should be pretty simple. If we add that to the manager should we change the current setup in the raw view tab, or leave it as is for now?
threshold

cjh1:
@bnmajor If its not to hard to keep them in sync, then leave the controls in the tab as well.

cjh1:
It seems to get a little confused if you give two masks the same name and then remove one:

```
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/mask_manager_dialog.py", line 82, in remove_mask
    del self.masks[item.text()]
KeyError: 'mask2'
single click: button=1, x=385, y=468, xdata=7.410278, ydata=-20.608611
single click: button=1, x=734, y=463, xdata=14.899466, ydata=-15.749117
single click: button=1, x=435, y=379, xdata=8.483227, ydata=65.890375
single click: button=1, x=634, y=372, xdata=12.753567, ydata=72.693666
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/mask_manager_dialog.py", line 100, in update_mask_name
    self.masks[new_name] = self.masks.pop(self.old_name)
KeyError: 'mask2'
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/mask_manager_dialog.py", line 100, in update_mask_name
    self.masks[new_name] = self.masks.pop(self.old_name)
KeyError: 'mask2'
single click: button=1, x=275, y=559, xdata=5.049789, ydata=-109.051394
single click: button=1, x=869, y=530, xdata=17.796430, ydata=-80.866331
single click: button=1, x=297, y=326, xdata=5.521886, ydata=117.401007
single click: button=1, x=816, y=360, xdata=16.659103, ydata=84.356451
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/mask_manager_dialog.py", line 100, in update_mask_name
    self.masks[new_name] = self.masks.pop(self.old_name)
KeyError: 'mask2'

```

bnmajor:
Sounds good, I'll leave the original controls in place then.

The same name issue makes sense, it should really enforce that all names are unique, I'll fix that.

cjh1:
@bnmajor A couple of things I found during testing:

- If you uncheck "Set Threshold Mask" the threshold mask stays in the manager. I think it should probably be removed? Thoughts?
- If you  create three masks say 'mask_0', 'mask_1', 'mask_2'. Then rename 'mask_2' => 'mask_3'. Then try to create another mask it doesn't appear in the manager.

bnmajor:
> * If you uncheck "Set Threshold Mask" the threshold mask stays in the manager. I think it should probably be removed? Thoughts?
I left this this way intentionally because toggling `Set Threshold Mask` in the image mode tab acts more like a visibility toggle than removing the mask - toggling will re-use the previous values, but removing the mask (from the manager) resets the values. If the mask should be removed from the manager when toggled in the image mode tab, should it also reset the values to match the `Remove Mask` behavior in the manager?

> * If you  create three masks say 'mask_0', 'mask_1', 'mask_2'. Then rename 'mask_2' => 'mask_3'. Then try to create another mask it doesn't appear in the manager.
I'll look into this and get it fixed.

cjh1:
@bnmajor Lets leave the toggling functionality as it is, we can follow up if necessary. I will try out your fix.

hexrd/ui/mask_manager_dialog.py
```diff
@@ -38,20 +38,28 @@ def create_masks_list(self):
...
             data = HexrdConfig().polar_masks_line_data
             if not data:
                 return
             if any(np.array_equal(data[-1], m) for m in self.masks.values()):
                 return
-
-            self.masks['mask_' + str(len(data) - 1)] = data[-1]
-            self.visible['mask_' + str(len(data) - 1)] = data[-1]
+            name = self.create_unique_name(
+                'mask_' + str(len(data) - 1), len(data) - 1)
```
cjh1, line 60: Would it be easier just to pass `'mask', 0` here and then create_unique_name would take care of finding the right name?

hexrd/ui/mask_manager_dialog.py
```diff
@@ -38,20 +38,28 @@ def create_masks_list(self):
             self.masks['threshold'] = HexrdConfig().threshold_mask
         self.visible = copy.copy(self.masks)
 
+    def create_unique_name(self, name, value=0):
+        while name in self.masks.keys():
+            value += 1
+            name = name.split('_')[0] + '_' + str(value)
```
cjh1, line 44: This could also be write as:

```python
prefix  = name.split('_')[0]
name = f'{prefix}_{value}'
```

hexrd/ui/mask_manager_dialog.py
```diff
@@ -0,0 +1,154 @@
...
+            if not any(np.array_equal(m, v) for m in self.masks.values()):
+                self.masks['mask_' + str(i)] = copy.copy(v)
+        if HexrdConfig().threshold_mask_status:
+            self.threshold_name = 'threshold'
+            self.masks['threshold'] = HexrdConfig().threshold_mask
+        self.visible = copy.copy(self.masks)
+
+    def create_unique_name(self, name, value=0):
+        while name in self.masks.keys():
+            value += 1
```
cjh1, line 43: Should we increment at the end of the loop so, we start at `mask_0` ?

bnmajor:
Simplified the `create_unique_name` function with the suggestions, thanks!

PR #447, opened 2020-08-10
Add default options for overlay types

psavery:
Previously, if there were no options provided, the overlay generator
would set defaults. These defaults would be different from the GUI
defaults, which led to a mismatch.

To avoid this, create default options for the different overlay types,
and always set those default options when the overlay is created or
the type is changed. This ensures that the overlay generator and the
GUI both use these defaults, to avoid a mismatch.

### DIFF
--- a/hexrd/ui/constants.py
+++ b/hexrd/ui/constants.py
@@ -1,4 +1,5 @@
 from enum import Enum
+import numpy as np
 
 from hexrd import constants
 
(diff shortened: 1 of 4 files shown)


joelvbernier:
Noticed an error regarding these:  The default stretch tensor should be identity, _not_ zeros, as that causes a singularity and raises an exception.  Also, only 6 DOF in the matrix should be accessible to the user:  the 3 diagonal terms (default to 1) and the three upper triangle terms (default to 0).  The matrix is symmetric by construction.

psavery:
This PR actually fixes the default stretch tensor error.

Also, the matrix automatically updates values opposite the diagonal to keep the symmetry (so it stays symmetrical). But, we can remove the bottom three triangle terms if that's preferable.

Issue #386, opened 2020-07-08
Add mono rotation series overlays


psavery:
An overlay manager and Laue overlays are added in #373 

This overlay manager allows the user to select mono rotation series, but it does not do anything (or will produce a `NotImplemented` error) if they do.

We will, at some point, need to add mono rotation series overlays.

This will likely require changes in the following sections of the code:
1. The [overlay data generator](https://github.com/HEXRD/hexrdgui/blob/04d533b077cb6ce5a72e9913dd3441f9ee8079e9/hexrd/ui/overlays/mono_rotation_series.py#L99)
2. Some sections in [the style picker](https://github.com/HEXRD/hexrdgui/blob/04d533b077cb6ce5a72e9913dd3441f9ee8079e9/hexrd/ui/overlay_style_picker.py#L185-L194)
3. Plotting the data [in the image canvas](https://github.com/HEXRD/hexrdgui/blob/04d533b077cb6ce5a72e9913dd3441f9ee8079e9/hexrd/ui/image_canvas.py#L223-L224)

Some extra changes might be needed as well. Some of the mono rotation series overlays may require 3D plotting.

joelvbernier:
The omega channel of the output points has a 1:1 correspondence with a frame index in the imageseries; if the display is aggregated then the omega channel is ignored, and we plot them as we do the Laue spots.

psavery:
To whoever works on this: before doing this, by the way, we might want to separate all of the laue stuff in the [`OverlayEditor`](https://github.com/HEXRD/hexrdgui/blob/593e874401c3f9910e844443c1f200420330db5a/hexrd/ui/overlay_editor.py#L12) into a separate `LaueEditor` widget, so we don't need to prefix/suffix all of the functions and variable names with `laue`.

This isn't so bad for laue, but prefixing/suffixing all the variable names with `mono_rotation_series` might become burdensome.

PR #450, opened 2020-08-12
Add mono rotation series overlays

psavery:
This includes a basic editor for the overlay, and some basic drawing.

image

However, what is being drawn may not be correct. We will need to look into this more.

To do:
- [ ] Ensure data is being generated correctly (there is an aggregation option, but it is currently not in use)
- [ ] Ensure x/y axes are correct for the different views

Fixes: #386


psavery:
@joelvbernier Any feedback on the editor above?

joelvbernier:
Let's come back to this after the demo

psavery:
Sounds good

joelvbernier:
Let's try to revive this issue this week.  I'll take a look at where things are at.

psavery:
@joelvbernier Sounds good, I think some good test data would be helpful.

psavery:
@joelvbernier A simulated example like we did for the Laue spots would be helpful for this.

joelvbernier:
@psavery -- one of the ruby examples (say singleGE) would be a good comparative data set since there is only 1 grain.  I can make a simulated imageseries as well (current simulator only makes point groups, not images, but I have a first cut at making images almost ready to merge).

Issue #33, opened 2019-07-03
Add Check to See if Cartesian Calibration is Feasible


psavery:
Sometimes, only polar calibrations will be able to be performed. Add in a validation check to make sure cartesian calibrations can be done before doing it.

It is considered feasible to do a cartesian calibration if the dot product of detector normals is less than 120 degrees or so.

psavery:
@joelvbernier I have the tardis example here, where the angle between the two detectors is 180 degrees, and it looks like hexrdgui is still able to generate a Cartesian view of it.

![tardis](https://user-images.githubusercontent.com/9558430/90278143-e92e1b00-de34-11ea-90ab-776ad88247b0.png)

Should we still follow this issue, and produce an error if the angle between the two detectors is greater than 120 degrees? If so, we won't be able to view examples like this one.

PR #453, opened 2020-08-14
Add Cartesian feasibility checks and better error messages

psavery:
This adds a few feasibility checks for the Cartesian plot, and it also adds better error messages for both polar and Cartesian modes. The following were added:

1. A memory check for the Cartesian view. If more memory will be used than the available memory, an exception is produced.
2. An angle check for the Cartesian view. If any two detectors have an angle between them greater than 120 degrees, an exception is produced.
3. Better error messages for Cartesian and polar views. Now, if an exception is raised when one is being performed, the message of the exception gets displayed in a QMessageBox for the user to see.

An example is provided below

example

Fixes: #33

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -1,6 +1,7 @@
 import copy
 
 import numpy as np
+import psutil
 
 from hexrd.gridutil import cellIndices
 
@@ -42,6 +43,10 @@ class InstrumentViewer:
 
         self.dplane = DisplayPlane(tvec=dplane_tvec, tilt=dplane_tilt)
         self.make_dpanel()
+
+        # Check that the image size won't be too big...
+        self.check_size_feasible()
+
         self.plot_dplane()
 
     def check_keys_match(self):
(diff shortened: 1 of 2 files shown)


joelvbernier:
This looks great, although I need to rethink the condition for feasibility -- the projection plane normal and beam vector should come into play as well.  Will get back to this tomorrow.

psavery:
@joelvbernier Okay. Yeah, it's currently checking to see if the angle between any two detectors is greater than 120 degrees. But we can change it to whatever is best.

cjh1:
@joelvbernier Any update on what the feasibility check should be?

joelvbernier:
I think perhaps a better approach would be to default to a different view mode.  We could do stereographic instead, which can be computed from the polar view.  Upon further thought, this would make for a better UX than just saying "nope"/  Give me a few hours to test some code.

joelvbernier:
We could just merge this and change later, or update this PR.  Your call.

cjh1:
Yes, lets rebase and merge.

psavery:
I'm on it

psavery:
@cjh1 Ready for review

cjh1 (approved):
LGTM

joelvbernier:
@psavery and @cjh1 -- I think I have a solution for what to do here.  I will make a stereographic projection to replace the planar projection in the failure case this PR picks up.  Working the function now.

PR #454, opened 2020-08-17
Set the hexrd icon on both appication and main window

johnkit:
Sets hexrd icon on both application and main window, in order to show up in the macOS dock panel.


johnkit:
@cjh1 would you please try this on linux to make sure I didn't break anything?

cjh1:
@johnkit The icons still shows up fine on Linux.

johnkit:
I also checked on windows and icons OK there too.

PR #455, opened 2020-08-18
fixed extents used in imshow

joelvbernier:
Pyplot extents are [left, right, bottom, top] and the y-axis was flipped.
Screen Shot 2020-08-17 at 5 37 39 PM
and after fix:
Screen Shot 2020-08-17 at 5 42 51 PM
Although after the fix something odd is happening with the detector borders (@psavery?)


psavery:
@joelvbernier Should we change polar to be this way as well? It's currently going top -> bottom.
Screenshot from 2020-08-18 13-02-56

PR #456, opened 2020-08-18
Associate quick load files with selected detector

bnmajor:
Addresses #414

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -285,6 +285,11 @@ class MainWindow(QObject):
 
             if dialog.exec_():
                 detector_names, image_files = dialog.results()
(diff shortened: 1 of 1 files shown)


hexrd/ui/main_window.py
```diff
@@ -285,6 +285,11 @@ def open_image_files(self):
 
             if dialog.exec_():
                 detector_names, image_files = dialog.results()
+                image_files = [img for f in files for img in f]
+                files = [[] for det in HexrdConfig().detector_names]
```
cjh1, line 289: Just for interest you can achieve the same with:

```python
[[]] * len(HexrdConfig().detector_names)
```
No  need to change, just added for interest :-)

PR #470, opened 2020-08-20
Single letter det names

bnmajor:
Fixes problems with automatic association that @joelvbernier was seeing, discussed in #414. Also adds warning dialogs when there is not at least one file for each detector.


joelvbernier:
Yeah -- I'm thinking it would be best practice to avoid single letter detector names.  If we have such data, at least we can manually assign them to the corresponding detectors.  We should change the reference PXRDIP names to "LFT", "RGT", "TOP", "DWN", "BCK"

Issue #432, opened 2020-07-23
Add mask to elliptical, rectangular regions


joelvbernier:
In addition to the hand-drawn component already merged in, can we add a tool to specify generic elliptical and rectangular areas?  Hopefully you have code to steal from other projects...  I am stealing my inspiration from ImageJ:
![ImageJ_rect_ellp_selectors](https://user-images.githubusercontent.com/1154130/88346319-e8560d80-ccfc-11ea-9ba8-e47068eaacbe.gif)
Somewhat lower priority than other issues, but I wanted to put it up there.

cjh1:
@joelvbernier as a side note the animated gifs are really useful thanks.

joelvbernier:
Bumping this up in priority as we have a detector case where we need to be able to generate a drawn mask in the raw view (see interpanel stripes below) using a series of rectangular regions (can't be done with intensity thresh in this case).  Also need to be able to export as we discussed today.
![Screen Shot 2020-08-18 at 3 20 23 PM](https://user-images.githubusercontent.com/1154130/90571465-a3d85900-e166-11ea-85be-1d7429634f98.png)

cjh1:
@joelvbernier This is in @bnmajor 's queue

psavery:
I believe this was fixed by #481. Feel free to open it back up if I'm wrong.

PR #481, opened 2020-08-24
Mask by shape

bnmajor:
Add the ability to mask images with ellipse or rectangle shapes.

Addresses #432


bnmajor:
mask-by-shape

bnmajor:
This branch currently provides basic functionality for masking images by shape:

- Can select between rectangle or ellipse
- Selections are applied to each view (raw, cartesian, and polar)
- Masked selections are remembered when toggling between tabbed/untabbed views
- All selections made before pressing `Okay` are treated as one mask

Features that are still needed:

- [ ] Masks available in mask manager
- [ ] Option to export mask(s) - Will add to mask manager dialog
- [ ] Speed up drawing of shape (need to redraw subplot instead of entire canvas)

bnmajor:
@joelvbernier This currently works for masking while in the raw view only. Do these masks also need to be able to be drawn in polar view?

joelvbernier:
@bnmajor yes, would be nice to do either.  We can transform coordinates back from the polar view to raw pixel indices.

bnmajor:
The mask manager now has the option to export masks. Right-clicking on a mask will provide a context menu to export an individual mask, or the export button at the bottom can be used to export all visible masks into a single `npz` file.
export_masks


The option to mask in polar view has also been added, and masks applied in either raw view are applied to polar and vice versa.
applytopolar


Masking cannot currently be done in and is not applied to the cartesian view. This would be simple to add if needed.

Issue #486, opened 2020-08-24
TARDIS boundardy issues


joelvbernier:
@bnmajor I noticed that the TARDIS boundaries are coming in upside-down.  I want to have the boundary files define the detector coordinate system, so they have to be consistent.  I realize that I supplied the templates in cartesian coordinates; the logic behind this (rather than pixel coordinates) is that the scanned pixel size can change (though it defaults to 0.1mm).  Since displaying images defaults to putting the origin of pixel coordinates in the upper left corner, this is likely the source of the apparent flip.  I checked that the template files load in the intended orientation:
![TARDIS_templates](https://user-images.githubusercontent.com/1154130/91104198-23ac6a80-e622-11ea-8e66-29490e71bfd0.png)
There are two ways to proceed:
1. load the images into the cartesian frame using the pixel size and image extents; or
2. transform the overlays into pixel coordinates
Regarding 2, the easiest was to transform the cartesian overlays to pixels is via the reference instrument.  For each detector panel, there is a `cartToPixel()` that will output the list/array of `[(x_0, y_0), (x_1, y_1), ..., (x_n, y_n)]` coordinates and output the corresponding `[(i_0, j_0), (i_1, j_1), ..., (i_n, j_n)]` row/col pixel coordinates.  The conversion is based on the pixel size property set on the detector.  As long as the panel has the proper pixel size, we should be good to go.

bnmajor:
@joelvbernier I've updated the code use the `cartToPixel` function [here](https://github.com/HEXRD/hexrdgui/blob/70bbacffb7f6a0cdf63ff3ded199f5978282ca12/hexrd/ui/interactive_template.py#L26) in #487, which should be loading the templates properly now.

PR #487, opened 2020-08-25
Tardis boundaries

bnmajor:
Addresses #486 

tardis_ip2_boundary
tardis-ip4-boundary

### DIFF
--- a/hexrd/ui/interactive_template.py
+++ b/hexrd/ui/interactive_template.py
@@ -12,6 +12,7 @@ class InteractiveTemplate:
         self.parent = parent.image_tab_widget.image_canvases[0]
         self.ax = self.parent.axes_images[0]
         self.raw_axes = self.parent.raw_axes[0]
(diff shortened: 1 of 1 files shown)


hexrd/ui/interactive_template.py
```diff
@@ -12,6 +12,7 @@ def __init__(self, img, parent=None):
         self.parent = parent.image_tab_widget.image_canvases[0]
         self.ax = self.parent.axes_images[0]
         self.raw_axes = self.parent.raw_axes[0]
+        self.panels = self.parent.iviewer.instr._detectors
```
cjh1, line 15: This seems to be reaching into a private member, is there a better way we can access the detectors? If we are going to use a private member, we should make it public ...

bnmajor: Yes! There is a public property that can be accessed, this has been fixed.

joelvbernier: was about to say...  Thanks!

PR #490, opened 2020-09-01
compatability modifications for new distortion package

joelvbernier:
These changes will actually be irrelevant with the planned fixes to the calibration, but for the sake of consistency...

### DIFF
--- a/hexrd/ui/calibration/line_picked_calibration.py
+++ b/hexrd/ui/calibration/line_picked_calibration.py
@@ -48,11 +48,6 @@ def calibrate_instrument_from_picks(
         param_flags=None, xtol=1e-4, ftol=1e-4):
     """
     arguments xyo_det, hkls_idx are DICTs over panels
-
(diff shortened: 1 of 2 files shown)


psavery:
@joelvbernier I gave this a try, and it worked for me for both cases except one thing:

For the powder calibration, in hexrd, I got an error [here](https://github.com/HEXRD/hexrd/blob/eb8fb3cbf1e973b5e6344e1574df3200b2aad879/hexrd/instrument.py#L1977) that said `AttributeError: 'list' object has no attribute 'tolist'`, which implies it is already a list.

However, I see that it should be converted to a numpy array [here](https://github.com/HEXRD/hexrd/blob/eb8fb3cbf1e973b5e6344e1574df3200b2aad879/hexrd/distortion/ge_41rt.py#L200). So I'm not sure why its a list?

Removing `tolist()` fixed it for me, and the powder calibration ran successfully.

Issue #482, opened 2020-08-24
Add adapt to new material package in hexrd


cjh1:
The following [PR](https://github.com/HEXRD/hexrd/pull/55) replaces the material module. Material information will be stored in a HDF5 format and will contain atomic information. The first stage will be to replace our existing [materials.hexrd](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/resources/materials/materials.hexrd) with the new HDF5 format, @joelvbernier  and @saransh13 are going to provide this file.

joelvbernier:
@saransh13, perhaps we should hold off on deleting the sglite and existing (legacy) materials modules if there still there in order to maintain compatibility with cpickles materials spec for the near term.  We can mark them for deletion in the future.

saransh13:
Yeah, that sounds like a good plan.

psavery:
@joelvbernier @saransh13 Should I also try to migrate from using `hexrd.spacegroup` to `hexrd.symbols`? We are currently using `hexrd.spacegroup` to get the following:

1. Hall symbol
2. Hermann–Mauguin symbol
3. Laue group
4. Lattice type
5. Required params (ie., do we need just **a**, or **a** and **c**, etc.).

saransh13:
I think we can just stick to `hexrd.spacegroup` for now. Can eventually be replaced by `hexrd.symbols` in the future. 
@joelvbernier ?

saransh13:
@psavery the latest push should have the `material.dum_material` function.

psavery:
Thanks @saransh13. I'll give it a try.

PR #495, opened 2020-09-03
LLNL I/O Bugs

bnmajor:
This branch addresses issues brought up in #492

- [x] Overlays need to be cleared before import panel can be used
- [ ] Rotate/translate requires app re-focus before they can be used (Mac issue)
- [ ] Rotate requires right-click in order to work (Mac issue)
- [x] Error on detector transform
- [x] Fix flip horizontally/vertically names to be clearer
- [x] Image axis limits not updated on crop
- [x] Detector keys need to be changed
- [ ] Tilts and translations need to be assigned to each plate

PR #496, opened 2020-09-03
Updates for the new hexrd Material class

psavery:
This provides a few updates in order to migrate to using the
new hexrd Material class. The main changes are:

1. Use the new HDF5 files for saving and loading the materials.
2. Migrate max HKL to min d spacing in the material editor widget.

Many parts were tested and seem to work. A couple of parts that might
require some work or further thought:

1. Space groups with a colon followed by a number produce an error in
   hexrd. I'm not sure if we need to change something in hexrdgui, or
   in hexrd.
2. Space group settings are ignored, even though they appear to be
   options. For instance, 167:H and 167:R are treated the same. We
   can either remove the space group setting options, or make them work.

A couple of other usability issues are:

1. If the user accidentally sets a min d spacing that is very small
   (less than 0.01, for instance), the GUI will freeze for a long time
   while the HKLs are being generated. We may want to put a lower limit
   in the GUI to help prevent accidents. But the user may also want to
   use a small number for their particular use case.
2. If the user accidentally sets a min d spacing that is too high,
   no hkl's will be generated, and various exceptions will be raised
   in hexrd.

Depends on: HEXRD/hexrd#55

Fixes: #482


cjh1:
@psavery I tested this out, seems to work well. The only thing I noticed is I think we should ensure the `.h5` ending when we save out the materials, as in if the user doesn't type it we should add it.

psavery:
The build error we are getting on Mac is the following (occurs during CPack):
```
  File "package.py", line 226, in <module>
    build_package()
  File "/usr/local/miniconda/envs/hexrdgui-package/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/miniconda/envs/hexrdgui-package/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/miniconda/envs/hexrdgui-package/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/miniconda/envs/hexrdgui-package/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "package.py", line 201, in build_package
    tar_path = build_conda_pack(base_path, tmp, hexrd_package_channel, hexrdgui_output_folder)
  File "package.py", line 149, in build_conda_pack
    format=archive_format
  File "/usr/local/miniconda/envs/hexrdgui-package/lib/python3.7/site-packages/conda_pack/core.py", line 447, in pack
    ignore_editable_packages=ignore_editable_packages)
  File "/usr/local/miniconda/envs/hexrdgui-package/lib/python3.7/site-packages/conda_pack/core.py", line 139, in from_prefix
    files = load_environment(prefix, **kwargs)
  File "/usr/local/miniconda/envs/hexrdgui-package/lib/python3.7/site-packages/conda_pack/core.py", line 773, in load_environment
    raise CondaPackException(_missing_files_error.format(packages))
conda_pack.core.CondaPackException: 
Files managed by conda were found to have been deleted/overwritten in the
following packages:

- python='3.8.5'

This is usually due to `pip` uninstalling or clobbering conda managed files,
resulting in an inconsistent environment. Please check your environment for
conda/pip conflicts using `conda list`, and fix the environment by ensuring
only one version of each package is installed (conda preferred).
```

A similar error is occurring on Windows as well.

cjh1:
Looks like this PR is showing it as well #495, I will take a look

Issue #389, opened 2020-07-08
Slider widgets for Laue overlay options


psavery:
In the overlay manager added in #373, there are widgets for editing Laue overlay options.

![image](https://user-images.githubusercontent.com/9558430/86964324-dd638080-c133-11ea-84da-666f411601b5.png)

We should look into adding slider widgets to them to make them more interactive. The slider widgets may operate in a similar way that the detector slider widgets do, in the slider view:
![image](https://user-images.githubusercontent.com/9558430/86964556-40551780-c134-11ea-806e-33a315fbf25d.png)

The UI for the Laue overlays is part of the [OverlayEditor](https://github.com/HEXRD/hexrdgui/blob/794792a0e7c11c5b3caf3f85e1449a521532b365/hexrd/ui/overlay_editor.py#L12) and the [CalibrationCrystalEditor](https://github.com/HEXRD/hexrdgui/blob/794792a0e7c11c5b3caf3f85e1449a521532b365/hexrd/ui/calibration_crystal_editor.py#L12).

We should make sure that the overlay manager doesn't become too big with the addition of these widgets. The user most likely wants to see the image canvas while they are interacting with the overlay manager.

psavery:
Also note that as we work on this, we may also need to look into speeding up some of the Laue spot generation code, to make this as interactive as possible.

joelvbernier:
We are going to push some updates to the material, which might help.  But yes, I am sure there is room for optimization in the simulator.

psavery:
Should we add sliders for the stretch matrix values as well? Or only for the orientation, position, and min and max energies?

johnkit:
@joelvbernier   Here is a strawman/mockup for revising the crystal calibration widget to include "form" and "slider" tabs. (The "form" tab is essentially the current widget.) To conserve vertical space, the slider tab has a dropdown box to select between position and orientation.

![mockup-calcrystal-form2](https://user-images.githubusercontent.com/5559383/90061393-1606f480-dcb4-11ea-84fb-7d1dfe951dbc.png) 
![mockup-calcrystal-slider2](https://user-images.githubusercontent.com/5559383/90061401-1a331200-dcb4-11ea-801f-bf40c5de77a8.png)

cjh1:
@johnkit This look great. My one question for @joelvbernier is do you want to be able adjust the orientation and position at the same time. In which case we might want the duplicate the sliders, so they are accessible without using the drop down.

joelvbernier:
[...]

PR #498, opened 2020-09-04
Laue sliders

johnkit:
Connected to Issue #389. Adds tab with sliders to the Laue overlay editor. Current UI has 3-slider layout with switch between orientation and position. The current design updates the slider range for each transition between modes and between tabs.


hexrd/ui/calibration_crystal_slider_widget.py
```diff
@@ -0,0 +1,173 @@
...
+        self._mode = WidgetMode.ORIENTATION
+
+        loader = UiLoader()
+        self.ui = loader.load_file(
+            'calibration_crystal_slider_widget.ui', parent)
+        for w in self.spinbox_widgets:
+            w.setStyle(SpinBoxStyle())
+
+        self._orientation = [0.0] * 3
+        self._orientatin_range = 30.0
```
cjh1, line 45: Let pull the default range into a constant

cjh1 (approved):
LGTM, just one suggest.

hexrd/ui/calibration_crystal_slider_widget.py
johnkit: Good idea. Just pushed commit with that change.

PR #501, opened 2020-09-04
Remove button defaults from dialog widgets

### LINKED ISSUE #472: Deactivate enter key capture for dialogs
Many of the separate dialogs (_e.g._, the overlay manager, the picker widow, ...) have a button in focus by default that captures the enter key.  Can we disable this behavior in all cases?  I know it is not necessary, but I invariably hit enter after inputing text into widgets (habit...).

johnkit:
Filters `<Enter>` key events from dialogs so that there is no default behavior for that key.

* For buttons in the .ui file, this PR clears autoDefault flag
* For QButtonBox buttons, this PR adds an EnterKeyFilter class to filter  <Enter> key events.

resolves #472

### DIFF
--- a/hexrd/ui/__init__.py
+++ b/hexrd/ui/__init__.py
@@ -1,0 +1,24 @@
+from PySide2.QtCore import QEvent, QObject, Qt
+from PySide2.QtWidgets import QDialog
+
+
+class EnterKeyFilter(QObject):
+    """An event filter to ignore <Enter> keys in QDialog instances.
+
+    QButtonBox will **always** assign a default button to the <Enter> key,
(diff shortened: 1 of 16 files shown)


cjh1 (approved):
LGTM, thank you

cjh1:
@johnkit Is going to look at using `QWidget.setFocusPolicy(QtCore.Qt.NoFocus)` as an alternative to the filter.

johnkit:
Side note: tried using `QWidget.setFocusPolicy(QtCore.Qt.NoFocus)` however QButtonBox is overriding the behavior. So we'll stick with the event filter. Will rebase and fix conflict shortly.

Issue #492, opened 2020-09-02
New errors in LLNL I/O


joelvbernier:
I am getting some new errors when I try to execute the LLNL I/O.  This is on the Mac (I'll try to reporduce on Ubuntu).

For TARDIS selected:

1. with overlays selected (_e.g._, from a saved state) I get the following exception
```
Traceback (most recent call last):
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/main_window.py", line 775, in update_all
    self.ui.image_tab_widget.load_images()
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/image_tab_widget.py", line 119, in load_images
    self.load_images_untabbed()
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/image_tab_widget.py", line 99, in load_images_untabbed
    self.image_canvases[0].load_images(
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/image_canvas.py", line 123, in load_images
    self.update_overlays()
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/image_canvas.py", line 282, in update_overlays
    self.iviewer.update_overlay_data()
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/calibration/raw_iviewer.py", line 17, in update_overlay_data
    update_overlay_data(self.instr, self.type)
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/overlays/__init__.py", line 90, in update_overlay_data
    overlay['data'] = generator.overlay(display_mode)
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/overlays/powder_diffraction.py", line 46, in overlay
    indices, ranges = self.plane_data.getMergedRanges()
  File "/Users/joel/Documents/GitHub/hexrd/hexrd/crystallography.py", line 957, in getMergedRanges
    (tThRanges[:-1, 1] < tThRanges[1:, 0], True)
IndexError: too many indices for array
```
and the image doesn't load.  Note that we **do not** want to do overlays in the I/O graphics canvas, rather only after the images are loaded and processed.

2. when trying to use the rotate/translate tools, I am getting:
```
Traceback (most recent call last):
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/import_data_panel.py", line 139, in setup_translate
    self.it.disconnect_rotate()
[...]

PR #502, opened 2020-09-08
LLNL I/O Transforms

bnmajor:
Addresses the tilt and transform issues discussed in #492 

Need to confirm that the tilt parameter is being set correctly/from the correct data and the defaults being used for the instrument are the correct defaults.

In order to select defaults for by detector, the config file is not updated until the detector is selected (a minor change, but the behavior is different than before).


bnmajor:
@joelvbernier will be adding reference instruments to use as defaults, closing this in favor of a re-worked PR using the correct defaults.

Issue #493, opened 2020-09-02
Strange behavior in Instrument Calibration widget


joelvbernier:
I've noticed some odd inconsistencies in the instrument calibration widget on the Mac -- haven't tested the same on linux yet but I'll try to and update this issue accordingly.

1. In the Slider View:
  - The labels for the tilt (X, Y ,Z) are misleading if you are using a tilt parameterization other than extrinsic XYZ; we should either update them to match the selected parameterization (if possible), or give them more generic names.
  - The spin controller widgets seem to be very restricted in terms of the precision of the number you can enter.  For example, I cannot enter '-90' in a tilt box.  I know we have hashed over this before, but ideally, we can decouple the displayed number of digits and what it can take as input (surely we must be able to do this, right?).  Either way, I need to be able to input more than one digit.  **_UPDATE_**: this seems to be a Mac thing as it is ok in Ubuntu -- oh boy...
2. In the Form View:
  - The field do not update with a change in the detector selector -- unless I toggle between one of the other tabs and back.
  - Regardless of the checkbox state in the tree view, any edits to the text input widgets affect no change in the underlying vaules; _i.e._, they disappear if I toggle to either of the other view tabs.
3. In the Tree View:
  - Not sure if we tried this before, but is it possible to change the units to degrees for the tilts if the chosen parameterization is not `None`?  The other two view modes are in degrees, but this is in radians for the tilts.

psavery:
For the X Y Z labels in the slider view, we can probably just add a function that sets them based on the current Euler angle convention. 

Indeed the spin controllers are restricted to two decimal places. Maybe the best option is to use the [ScientificSpinBox](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/scientificspinbox.py) instead?

The min and max of the spinboxes are dynamically set to match the min/max of the sliders. This min/max is determined via the range spin box in the tops of the groups. This is by design, and it restricts the range of values you can enter in the spinboxes so the slider can be moved to match it (which is probably why you couldn't enter "-90", if it was outside the slider range). We can re-design this, though, if that is desirable. Maybe if the user sets a value in the spinbox that is outside the slider range, we can reset the range so the new value is centered in the slider, for instance.

For the form view, the fields appear to update for me with the detector selector, and the edits to the text input widgets appear to affect the underlying values. I think these might be part of #505. The python binary we have uploaded should help.

For the tree view, we haven't tried that before. But we can do it I believe!

[...]

PR #507, opened 2020-09-08
Calibration widget

bnmajor:
Addresses some of the issues discussed in #493:
- Updates the labels for tilt to match selected parameterization
- Change the units to degrees for tilts in tree view if parameterization is not None

Not yet fixed:
- Possibly use `ScientificSpinBox` to allow for more than two decimal places in spin controller widgets
- Possibly recalculate slider range if value is entered into the spin controller widget is outside of the current range
- Make sure form view fields update when detector selector is changed*
- Make sure edits made in the form view are not lost when calibration view tabs are toggled*
*May be Mac-specific issues

### DIFF
--- a/hexrd/ui/cal_tree_view.py
+++ b/hexrd/ui/cal_tree_view.py
@@ -76,6 +76,11 @@ class CalTreeItemModel(BaseTreeItemModel):
         item.set_data(index.column(), value)
 
         if item.child_count() == 0:
+            parent = item.parent_item.data(KEY_COL)
+            if (parent == 'tilt' and
+                    HexrdConfig().rotation_matrix_euler() is not None):
+                # Convert tilt values to radians before saving
+                value = np.radians(value).item()
             self.cfg.set_instrument_config_val(path, value)
             dist_func_path = ['distortion', 'function_name', 'value']
             if len(path) > 4 and path[2:5] == dist_func_path:
@@ -130,7 +135,8 @@ class CalTreeItemModel(BaseTreeItemModel):
         for key in keys:
             if key == 'value':
                 data = cur_config[key]
-                if cur_tree_item.data(0) == 'tilt':
+                if (cur_tree_item.data(KEY_COL) == 'tilt' and
+                        HexrdConfig().rotation_matrix_euler() is not None):
                     data = [np.degrees(rad).item() for rad in cur_config[key]]
                 self.set_value(key, data, cur_tree_item)
                 continue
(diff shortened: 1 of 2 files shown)


hexrd/ui/cal_tree_view.py
```diff
@@ -129,7 +129,10 @@ def recursive_add_tree_items(self, cur_config, cur_tree_item):
 
         for key in keys:
             if key == 'value':
-                self.set_value(key, cur_config[key], cur_tree_item)
+                data = cur_config[key]
+                if cur_tree_item.data(0) == 'tilt':
+                    data = [np.degrees(rad).item() for rad in cur_config[key]]
+                self.set_value(key, data, cur_tree_item)
```
psavery, line 135: The underlying instrument config will likely be expecting the angles to be in radians.

If the user modifies the tilt values when they are degrees, will this convert them back to radians?

psavery: Also note that if the Euler angle convention is "None", then we shouldn't convert to degrees.

hexrd/ui/calibration_slider_widget.py
```diff
@@ -290,3 +293,12 @@ def emit_update_if_polar(self):
     def reset_config(self):
         HexrdConfig().restore_instrument_config_backup()
         self.update_gui_from_config()
+
+    def update_labels(self):
```
psavery, line 297: This should probably be called in `update_gui_from_config()` as well, so that if the user starts the program and the Euler angle convention is already like "ZXZ", then when they switch to the slider tab, it will update the labels to match.

hexrd/ui/calibration_slider_widget.py
```diff
@@ -290,3 +293,12 @@ def emit_update_if_polar(self):
...
+        if HexrdConfig().euler_angle_convention is None:
+            return
```
psavery, line 299: We might still want to set them to `XYZ` here, because if the user switches to "ZXZ", and then switches "None", it will still display "ZXZ".

hexrd/ui/cal_tree_view.py
bnmajor: Thank you for catching that, the values were not being converted back to radians. Fixed conversion on value modification and only convert when the angle convention is `None`.

hexrd/ui/calibration_slider_widget.py
bnmajor: Fixed

hexrd/ui/calibration_slider_widget.py
bnmajor: Fixed

psavery (approved):
LGTM

Issue #483, opened 2020-08-24
Add new "structure" tab to materials panel


cjh1:
This will be used provide the atomic information. We are going to need some sort of periodic table widget, to be able to pick the elements. [this is one](http://www.silx.org/doc/silx/0.9.0/modules/gui/widgets/periodictable.html) possibility.

psavery:
@joelvbernier I think we're going to try out [silx](http://www.silx.org/doc/silx/0.9.0/modules/gui/widgets/periodictable.html) (also linked above), because it doesn't add many dependencies, and it seems good so far.

psavery:
@joelvbernier Do we want to be able to select multiple atoms in the periodic table for a site, or just one atom for a site? I see the example we have appears to allow multiple atoms in a single site, and then you specify a partial occupancy for each.

psavery:
@saransh13 @joelvbernier Okay, I've got a fair amount of the GUI in place so far.

Here is a site editor:
![Screenshot from 2020-09-09 14-50-17](https://user-images.githubusercontent.com/9558430/92640818-d5e36380-f2ab-11ea-9e55-989cadc15c73.png)

And if you press "Select Atom Types", a periodic table appears where you can select all of the atom types (yellow means it is selected, but we can change that):
![image](https://user-images.githubusercontent.com/9558430/92640878-f0b5d800-f2ab-11ea-8655-b5e1b9dd4632.png)


I'm working now on a way to convert the atom info in the Material [here](https://github.com/HEXRD/hexrd/blob/740d86326da2b687936df7015235d0d127a07ffe/hexrd/material.py#L99-L100), to the symmetry sites in the GUI, and then back to the expanded atom info in the Material class. I think I can do this using the functions and matrices from [hexrd.symmetry](https://github.com/HEXRD/hexrd/blob/master/hexrd/symmetry.py). Let me know if you have any thoughts.

psavery:
I think something like this works, for example, for expanding the sites:
```python
import numpy as np
from hexrd import symmetry

def expand_sites(spgnum, coords):
    return np.matmul(symmetry.GenerateSGSym(spgnum)[0], (*coords, 1))[:, :3]


expand_sites(8, (0.25, 0.25, 0.25))
# [[ 0.25  0.25  0.25]
#  [ 0.75  0.75  0.25]
#  [ 0.25 -0.25  0.25]
#  [ 0.75  0.25  0.25]]

expand_sites(62, (0.33, 0.25, 0.5))
[...]

PR #512, opened 2020-09-11
Add a structure tab to the materials panel

psavery:
The structure tab modifies sites and atomic properties for the
material.

Only a single U value for each atom type is currently allowed. However,
in the near future, we may need to allow for a symmetric tensor for
the U values for each atom type.

Currently, to update the `hexrd` objects, it updates the attributes directly
on the `Material` class, and then re-creates the `unitcell` from scratch, since
that seems to be easier than updating all the properties inside the `unitcell`.
It then calls a function to update the structure factor on the `PlaneData` object.

example

Depends on: hexrd/hexrd#97
Fixes: #483


cjh1 (approved):
This looking great! Thank you.

saransh13:
I agree. This look fantastic!

PR #515, opened 2020-09-14
Allow 'inf' and '-inf' in ScientificDoubleSpinBox

psavery:
These correspond to python's `math.inf` and `-math.inf`, respectively.
They can be alternatively obtained via `float('inf')` and `float('-inf')`.

If there is a max and min set on the spinbox, the value will be clipped
to the max and min. If not, however, `inf` and `-inf` are allowed and valid.


cjh1 (approved):
LGTM

PR #521, opened 2020-09-16
Change colormap QSpinBox to ScientificDoubleSpinBox

### LINKED ISSUE #511: Colormap editor currently not suitable for float images with small intensities
The QSpinBox widgets do not allow for adjusting the colormap appropriately for float images.  The fade corrected images in the NIF archive are of float type and I can't set the max appropriately low.  The simplest option would be to replace the widget with a DoubleSpinBox; we could alternatively rescale images, but that might be dicey.  Thoughts?

johnkit:
Updates colormap min/max fields to ScientificDoubleSpinBox for floating-point imagery with low intensity values.

Fixes #511

### DIFF
--- a/hexrd/ui/resources/ui/color_map_editor.ui
+++ b/hexrd/ui/resources/ui/color_map_editor.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>329</width>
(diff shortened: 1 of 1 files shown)


cjh1 (approved):
LGTM, thanks @johnkit

Issue #484, opened 2020-08-24
Integrating WPPF


cjh1:
Provide support for spectrum fitting.

joelvbernier:
Tasks 
- [ ] Provide form design for parameter input
- [ ] dialog to capture fit execution messages (if any)
- [ ] update polar plot to display spectrum fit
- [ ] dialog for fit results (modify active materials?)

psavery:
@saransh13 Do you have any examples of using WPPF in hexrd that I can start with?

saransh13:
@psavery I do have a two example datasets for Ceria and PbSO4. Right now `WPPF` is i/o from file. Do you want to wait until I convert it to input from values or get started with the file input?

saransh13:
@psavery I've uploaded ceria files on the google drive in the WPPF folder with an example script to initialize the class. I'm working on making the inputs directly from data instead of files.

cjh1:
Thanks @saransh13

saransh13:
I think we should repopulate the values in the WPPF dialogue box with the current values in `WPPF` class. Default values should be used when WPPF is run for the first time.

cjh1:
Implemented in #523 and #534

PR #523, opened 2020-09-16
Add WPPF generation and plot to the polar view

psavery:
This adds a new menu item, "Run"->"WPPF", which makes an options
dialog appear, and then generates WPPF data using the active
material, current beam energy, selected options, and the current
azimuthal integral plot.

Currently, only the "LeBail" WPPF method works. "Rietveld" is an option
but does not currently work.

Additional options appear in the dialog if the user selects a background
method that uses additional parameters. The user can either choose to
use an experiment file for the fitting data, or the current azimuthal
integral plot data will be used.

When the azimuthal integral axis in the polar view is drawn, the image
canvas automatically checks for the presence of WPPF data. If any was
generated, it will automatically plot it. There is currently no way
to clear the WPPF data (you can only replace it with a new one). So
once it has been generated, it will stick around. We can probably make
certain operations clear it at some point.

Currently, there can only be one set of WPPF data at a time. Also, the
options in the dialog are saved and re-loaded during the current run,
but are not saved in between runs. It is an easy fix if we wish to save
the options in between runs.

wppf_example

Depends on: hexrd/hexrd#103

Fixes part of #484


saransh13:
Looks great!

cjh1 (approved):
This is looking good, I have tested the basics. We can do follow up PRs as necessary. Thank you!

PR #517, opened 2020-09-14
apply offset corrections to overlays

joelvbernier:
see issue #516

Issue #516, opened 2020-09-14
Overlay fixes for offsets


joelvbernier:
Working on a PR to apply corrections to the angular overlays in the polar view mode consistent with an input offset vector.  The Laue overlays were actually broken in this regard already (_i.e._, wrong behavior when changing the crystal translation).  The issue in the polar view comes from the fact that the polar rebinning is done from the lab reference frame.  If you have a material point in your sample that is **_not_** at the origin -- such as a powder lines coming from a layered sample or a laue pattern from a crystal not at the origin -- the (2θ, η) must be adjusted to reflect this origin offset.  I've written a new function in hexrd.xrdutil to address this.

There is a [draft PR on hexrd](https://github.com/HEXRD/hexrd/pull/102) that includes the new function, as well as the beginning of the new calibration package.  The fixes to the hexrdgui overlays (just powder and Laue for the time being) include:
- [x] fix laue overlay to adjust the angular output using the new `hexrd.xrdutil._convert_angles()` method
- [x] fix powder overlay
- [x] add a translation offset to the overlay manager for powder overlays
The associated PR is linked.

psavery:
@joelvbernier I have a commit that adds the translation offset to the powder overlay editor, and it seems to work. Should I push it to #517?

joelvbernier:
Yes please!  I am pushing a few more commits to #517 right now as well.

cjh1:
@psavery do we still need to add the translation offset?

psavery:
@cjh1 This was added in #527 and #541. Closing this PR.

PR #527, opened 2020-09-17
apply offset corrections to overlays

psavery:
see issue #516

This PR contains the same relevant commits as #517, but this PR rebases on master to avoid introducing all of the extra commits.

The commits were cherry-picked via `git cherry-pick ...`. In the future, we can rebase on master to avoid the extra commits, like the following:

```bash
git checkout master
git pull origin master
git checkout overlay-fixes
git rebase origin/master
```

### DIFF
--- a/hexrd/ui/overlays/powder_diffraction.py
+++ b/hexrd/ui/overlays/powder_diffraction.py
@@ -15,6 +15,8 @@ class PowderLineOverlay:
                  eta_steps=360, eta_period=np.r_[-180., 180.]):
         self._plane_data = plane_data
         self._instrument = instr
+        tvec = np.asarray(tvec, float).flatten()
+        assert len(tvec) == 3, "tvec input must have exactly 3 elements"
         self._tvec = tvec
         self._eta_steps = eta_steps
         self._eta_period = eta_period
--- a/hexrd/ui/overlays/powder_diffraction.py
+++ b/hexrd/ui/overlays/powder_diffraction.py
@@ -1,5 +1,7 @@
 import numpy as np
 
+from hexrd.xrdutil import _convert_angles
+
 from hexrd.ui.constants import ViewType
 
 
@@ -7,9 +9,10 @@ nans_row = np.nan*np.ones((1, 2))
 
 
 class PowderLineOverlay:
-    def __init__(self, plane_data, instr, eta_steps=360):
+    def __init__(self, plane_data, instr, tvec=np.zeros(3), eta_steps=360):
         self._plane_data = plane_data
         self._instrument = instr
+        self._tvec = tvec
         self._eta_steps = eta_steps
 
     @property
@@ -19,6 +22,16 @@ class PowderLineOverlay:
     @property
     def instrument(self):
         return self._instrument
+
+    @property
+    def tvec(self):
+        return self._tvec
+
+    @tvec.setter
+    def tvec(self, x):
+        x = np.asarray(x, float).flatten()
+        assert len(x) == 3, "tvec input must have exactly 3 elements"
+        self._tvec = x
 
     @property
     def eta_steps(self):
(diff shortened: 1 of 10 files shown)


psavery:
@joelvbernier This PR also contains my commits that added the powder overlay offset options in the GUI.

I think raw and Cartesian look good. But polar looks something like this:

bad_polar

I think you mentioned this in one of our earlier meetings. What should we do about it? Should we try to make the overlays in the azimuthal plot look curved like those in the polar plot? Or something else.

I think the problem is that I coded it assuming it would only ever use vertical lines. See [here](https://github.com/HEXRD/hexrdgui/blob/909f7df6c1cf3b2e82ff967ce836dc1a932adfc2/hexrd/ui/image_canvas.py#L226)

hexrd/ui/resources/ui/import_data_panel.ui
```diff
@@ -7,26 +7,13 @@
     <x>0</x>
     <y>0</y>
     <width>491</width>
-    <height>711</height>
+    <height>472</height>
```
joelvbernier, line 10: this is messy, sorry. The size is irrelevant here, yes?

joelvbernier (approved):

joelvbernier:
Two issues with the plot.  First, I just realized that we need to remap the eta angles to the same period as the polar plot to avoid the "dollar sign" line plots.  I'll add a commit for this.

As far as the overlays in the lineout, let's draw line at the associated **_averaged_** postion for each 2θ line from the plot above.  Can't think of an option that would make more sense in this context.  @saransh13?

hexrd/ui/resources/ui/import_data_panel.ui
psavery: Yeah, I don't think it matters.

hexrd/ui/overlays/powder_diffraction.py
```diff
@@ -13,7 +13,7 @@ class PowderLineOverlay:
     def __init__(self, plane_data, instr, tvec=np.zeros(3), eta_steps=360):
         self._plane_data = plane_data
         self._instrument = instr
-        self.tvec = tvec
+        self._tvec = tvec
```
psavery, line 16: I had modified this to `self.tvec = tvec` so that it is guaranteed to be converted to a numpy array. `self._tvec = vec` is accepting a list and producing errors in the powder overlay editor.

Can we either:
1. Change it back to `self.tvec = tvec`
2. Do `self._tvec = np.array(tvec)`
?

psavery: Allowing a list to be passed in is easier, by the way, because it is easier to save via `QSettings` when we save the overlay arguments.

joelvbernier: solved, keeping option 2, but casting the kwarg to ndarray (with assertion error check)

joelvbernier (approved):

joelvbernier (approved):

joelvbernier:
A couple of observations testing this

1. the input to the translation vector causes a major lag; can we increase the delay upon changing the spin controllers before application of the updates to the overlay (I assume that would help ameliorate).
2. we need to connect the eta ranges (period) from the polar view to the overlays.  I added a kwarg for this in the merge above, but now we need to actually apply it.  I say this because I am getting the following:
Screenshot from 2020-09-17 13-40-59
3. I will have to do the same for the Laue overlays for the case of the polar view.  Once we add the ability to change the eta ranges, we'll need to make sure the overlays are consistent.  Only affects the case of the polar view type.

psavery:
I think 1 is being caused primarily by the bad behavior of the overlays in the azimuthal integration plot. Once we fix that, it should be much better I think. Do you want me to go ahead and plot an average of the x points as the vertical overlay there?

I will look into 2

It also looks like some parts are broken for the Laue widths. I started looking into it a little, but let me know if you'd rather fix it.

psavery:
Yeah I did a quick test, and the slowness in 1 is definitely being caused by the bad overlays in the azimuthal integration plot.

PR #529, opened 2020-09-17
Do not reset the color map ranges on I/O transform

### LINKED ISSUE #528: When applying a transform to the image in the I/O widget, the colormap ranges get reset. This should be disabled.

bnmajor:
Fixes #528

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -85,7 +85,7 @@ class MainWindow(QObject):
         self.load_widget.ui.setVisible(False)
         self.workflow_widgets[WORKFLOW_HEDM].append(self.load_widget.ui)
 
-        self.import_data_widget = ImportDataPanel(self.ui)
+        self.import_data_widget = ImportDataPanel(self.color_map_editor, self.ui)
         self.ui.load_page.setLayout(QVBoxLayout())
         self.ui.load_page.layout().addWidget(self.import_data_widget.ui)
         self.import_data_widget.ui.setVisible(False)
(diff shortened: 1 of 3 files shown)


bnmajor:
This branch now respects any color map range that is set for the entire LLNL I/O workflow - the range should not be automatically reset by any selections or changes until the workflow has been completed by selecting the `Complete and Reload Instrument` option.

hexrd/ui/main_window.py
```diff
@@ -85,7 +85,8 @@ def __init__(self, parent=None, image_files=None):
         self.load_widget.ui.setVisible(False)
         self.workflow_widgets[WORKFLOW_HEDM].append(self.load_widget.ui)
 
-        self.import_data_widget = ImportDataPanel(self.ui)
+        self.import_data_widget = ImportDataPanel(self.color_map_editor,
```
cjh1, line 88: Its not great to have the import panel depend on the color map editor. I wonder if there is another way, may be with signal connected together here? May be the import data widget could emit an 'updating started' and 'complete signal' which could connected by some function here to enable/disable the color map editor. Lets think about it a bit.

joelvbernier: Ok, we can hold on this for a bit to figure a better implementation.

joelvbernier:
Should we merge this up, or rework?

cjh1:
@joelvbernier We can merge and rework, I will raise a issue for it. @bnmajor can you rebase on master, thanks!

bnmajor:
@cjh1 Sure thing!

cjh1 (approved):
I have raise #547 for a future rework.

Issue #525, opened 2020-09-16
Update simulated spectrum plot between each WPPF iteration


saransh13:
Can we update the simulated plot after each iteration cycle i.e. everytime `LeBail.RefineCycle` is called?

PR #531, opened 2020-09-17
Update WPPF plot with each refinement

psavery:
This will increase the amount of time needed for WPPF to finish since
re-rendering takes a little time. However, it gives live updates as
the refinement process occurs.

wppf_refinement_updates

Fixes: #525


cjh1 (approved):

cjh1:
@psavery This looks really cool by the way :+1:

PR #532, opened 2020-09-17
Re-phrase transform flip labels

bnmajor:
Re-name all instances of  `Flip Horizontally/Vertically` to `Flip about Horizontal/Vertical Axis`

@joelvbernier if there is a preferred way to re-word the transform labels than `Flip about Horizontal/Vertical Axis` let me know and we can use that instead.

### DIFF
--- a/hexrd/ui/load_images_dialog.py
+++ b/hexrd/ui/load_images_dialog.py
@@ -81,7 +81,7 @@ class LoadImagesDialog:
                 table.setItem(i, 0, d)
 
             trans_cb = QComboBox()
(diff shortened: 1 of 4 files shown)


cjh1 (approved):
@joelvbernier Are you OK with the wording?

joelvbernier:
Ah crap, I already changes these all in #527

bnmajor:
@joelvbernier No problem, we can close this out if it has already been fixed.

joelvbernier:
Thanks, and sorry!

Issue #476, opened 2020-08-20
Polarview enhancements


joelvbernier:
We'd like to add the ability to specify the min/max eta ranges.  Currently defaults to (-180, 180), but there are cases where we'd like to have the ability to shift the view and even limit the coverage range to < 360° (_e.g._, for instruments with limited coverage).  The hooks are there in polarview.py to set the ranges.  Just need a check that `eta_max - eta_min` is in (0, 2π).

psavery:
This sounds good.

There are a few places in the code where -180 to 180 are hard-coded, like [here](https://github.com/HEXRD/hexrdgui/blob/292327b95247a683c289bba1682a2b46d7409731/hexrd/ui/calibration/polar_plot.py#L50), [here](https://github.com/HEXRD/hexrdgui/blob/292327b95247a683c289bba1682a2b46d7409731/hexrd/ui/constants.py#L32-L33), and [here](https://github.com/HEXRD/hexrdgui/blob/292327b95247a683c289bba1682a2b46d7409731/hexrd/ui/overlays/powder_diffraction.py#L40).

We'll just need to use something like `HexrdConfig().polar_res_eta_min` and max instead, similar to `HexrdConfig().polar_res_tth_min` and max. And we can add the eta options to the image mode widget.

joelvbernier:
bumping the priority on this.

psavery:
@johnkit Maybe do a grep of '180' on the source code to make sure we catch all the cases. We can replace them with something like `HexrdConfig().polar_res_eta_min`, and max.

johnkit:
Notes:
* Adding eta range to the image mode widget increases height slightly from 138 to 152 pixels (even though spinbox vertical height is 20px).
* For now, I'll set the spinbox min/max to [-540, 540] so that you can offset a full 2pi on either end.

![image](https://user-images.githubusercontent.com/5559383/93512013-a9ee6080-f8f1-11ea-9596-bbdc1463b8a7.png)

psavery:
We might want to use the unicode η similar to what is being done for 2θ.

And yeah, decreasing the size of the main canvas isn't that nice. But I think we can think of ways to reduce the image mode widget size later, if it isn't urgent now.

psavery:
@joelvbernier Question. Are we going to require eta_max - eta_min to be 360? Is eta_max - eta_min going to be the eta_period? Or is it always going to be eta_min + 360?

joelvbernier:
[...]

PR #536, opened 2020-09-17
Update polarview.py

joelvbernier:
adding the ability to map things to the proper eta range based on settings.

### DIFF
--- a/hexrd/ui/calibration/polarview.py
+++ b/hexrd/ui/calibration/polarview.py
@@ -28,6 +28,7 @@ def log_scale_img(img):
 class PolarView:
     """Create (two-theta, eta) plot of detectors
     """
+
     def __init__(self, instrument):
 
         self.instr = instrument
@@ -121,8 +122,8 @@ class PolarView:
 
     @property
     def eta_period(self):
-        return self.eta_min + 360.
-    
+        return self.eta_min + np.r_[0., 360.]
+
     def detector_borders(self, det):
         panel = self.detectors[det]
 
--- a/hexrd/ui/calibration/polarview.py
+++ b/hexrd/ui/calibration/polarview.py
@@ -122,7 +122,7 @@ class PolarView:
 
     @property
     def eta_period(self):
-        return self.eta_min + np.r_[0., 360.]
+        return HexrdConfig().polar_res_eta_period
 
     def detector_borders(self, det):
         panel = self.detectors[det]
--- a/hexrd/ui/calibration/polarview.py
+++ b/hexrd/ui/calibration/polarview.py
@@ -119,6 +119,10 @@ class PolarView:
     def angular_grid(self):
         return self._angular_grid
 
+    @property
+    def eta_period(self):
+        return self.eta_min + 360.
+    
     def detector_borders(self, det):
         panel = self.detectors[det]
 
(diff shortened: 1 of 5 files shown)


joelvbernier:
ok, so I added a property to the polarview class that yields the eta period (in degrees) as 360 + eta_min;  @psavery we need to somehow make this available to the kwarg in the overlay code.

psavery:
I think @johnkit has almost gotten the eta ranges in the image mode widget working for #476. This might be easier to use since it will be in the `HexrdConfig()` singleton.

joelvbernier:
@psavery Looks like we need to set the period at least [here](https://github.com/HEXRD/hexrdgui/blob/294a416ae3c09451a6452ae6714a78289f929042/hexrd/ui/calibration/polar_plot.py#L44)

joelvbernier:
@psavery sorry for the barrage of activity -- lmk if the changes I'm referring to are clear.

johnkit:
Hmm, yes we might be overlapping. I just created #538 to add eta min/max range. I'll mark it as draft if I can figure out *how*.

psavery:
@joelvbernier I think I just had a question I added here: https://github.com/HEXRD/hexrdgui/issues/476#issuecomment-694508546

joelvbernier:
@psavery answered.  @johnkit you are good -- I am setting the attributes to keep track of the actual branch cut used to represent the etas but adding an `eta_period` attribute that should default to 360. + eta_min.  That what when we calculate angles for the overlays, we can make sure they are consistent with the eta extents on the polarview plot :-)

joelvbernier:
Should I merge this, or is there some wiring you want to do on this branch @psavery ?

hexrd/ui/calibration/polarview.py
```diff
@@ -119,6 +119,10 @@ def update_angular_grid(self):
     def angular_grid(self):
         return self._angular_grid
 
+    @property
+    def eta_period(self):
+        return self.eta_min + 360.
```
psavery, line 124: Quick question: should this be a list instead of a single number, like `np.r_[self.eta_min, self.eta_min + 360.]`?

psavery:
@joelvbernier I think we can go ahead and merge it and I'll hook up things in a separate PR, if you'd like.

cjh1:
@joelvbernier @psavery yes, lets get this in.

hexrd/ui/calibration/polarview.py
joelvbernier: Oops, yes, should be array_like exactly how you have suggested.

joelvbernier:
I'll make the fix @psavery pointed out an then merge.

hexrd/ui/calibration/polarview.py
joelvbernier: it should be handled exactly how it is in the overlay modules (_e.g._, https://github.com/HEXRD/hexrdgui/blob/01c532b3fe55e69f6ac128268b2479da094b7841/hexrd/ui/overlays/laue_diffraction.py#L101)

joelvbernier:
man, I think I carelessly applied the rebase in the wrong order (forgot to pull changes into feature branch).  Sorry for being a dumbass.  Can you check me @psavery ?

psavery:
@joelvbernier I'll fix it

psavery:
@joelvbernier I think it's fixed now. You might have to delete your local branch and pull again.

psavery (approved):
LGTM

cjh1 (approved):

psavery:
@joelvbernier I added a commit here that passes the eta period to the overlays. They should be getting them now. Merging.

PR #537, opened 2020-09-17
Fix laue widths

psavery:
This fixes some errors in the Laue widths and seems to work for all three views.


cjh1 (approved):

joelvbernier (commented):
Oops, my bad :-)

PR #538, opened 2020-09-17
Add eta min/max values to image mode widget and config

### LINKED ISSUE #476: Polarview enhancements

johnkit:
Adds spinbox fields to the image mode widgets. Some notes:

* The polar plots don't update if no image data have been loaded. This is filtered by `MainWindow.update_all()`
* Remember to enable View => Live Updates

Resolves #476


cjh1:
@johnkit Looking good. I think we what the interaction to be the same as 2 theta, as in if you type a number its only applied when you leave the spinbox. How it is now its hard to type something like 170, as it renders the view as soon as you type 1.

psavery:
@johnkit I think you can do this by turning keyboard tracking off. You can do it either in Qt designer (preferable if possible) or programmatically.

cjh1:
Indeed thanks @psavery

johnkit:
Thanks @psavery @cjh1 I was not familiar with the keyboardTracking option. Just force-pushed the change.

cjh1:
@johnkit Thanks, the interactioon is better one final question, should the increment interval be the same as 2 theta? Rather than a single degree? Thoughts @psavery ? 

Also seeing the following if you try to update the spin box too quickly. 

```
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 14, in polar_viewer
    return InstrumentViewer()
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 29, in __init__
    self.draw_polar()
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 45, in draw_polar
    self.pv.warp_all_images()
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/calibration/polarview.py", line 247, in warp_all_images
    self.create_warp_image(det)
  File "/home/cjh/work/source/hexrdgui/hexrd/ui/calibration/polarview.py", line 209, in create_warp_image
    self.warp_dict[det] = panel.interpolate_bilinear(
ValueError: cannot reshape array of size 674230 into shape (1769,382)

```

johnkit:
My guess is it's because the image is updated with an asyncworker. Maybe we need to freeze/throttle the input some way to prevent launching a second worker when one is already running?

cjh1:
Yes, I think that is the issue. We could just debounce the spinbox updates ...

cjh1:
or check if we are already running an update ...

cjh1:
This can be a follow up PR

cjh1:
@joelvbernier Any opinion on the spinbox increment?

joelvbernier:
Increment in what sense?  Because the update can take a little time, I'd also be in favor of an apply button to regent the polar plot.  There aren't great solutions to live updating from the pv spec spin controls in a general sense, and there is no need for them to be updating live; an "apply" or "update" signal would simplify/solve the issue methinks.  Can skip the update if the spin controllers haven't been updated as well.  Thoughts?

cjh1:
@joelvbernier In terms of incrementing by a degree or fraction of a degree like the 2 theta spinbox.

joelvbernier:
For eta ranges a degree is fine -- the typical use case would be to enter, not adjust by spin control since they are coarse bounds.  In fact, all of the interactors on the image mode selector for polar view could be plainTextEdit widgets rather than QDoubleSpinBox widgets, as I've used the spin arrows exactly never.  The exception would be the n iter box (integer) for SNIP.  Would that simplify things?  Or is there another reaso to keep them as spin controllers?

psavery:
We can hide spin arrows with this enum, by the way: https://doc.qt.io/qt-5/qabstractspinbox.html#ButtonSymbols-enum

cjh1:
No reason to keep them, other than not having to change anything :smile:

cjh1:
Lets merge what we have and the do any improvements as a follow up.

joelvbernier:
@cjh1 since we have to monkey with that form anyhow...  I pitched a new design in #476 to address the form factor issues for adding eta min/max input widgets.

cjh1:
@joelvbernier Yes, that looks like a nice way to space some space. We can do that as a follow up if that is OK?

cjh1 (approved):

joelvbernier:
ok -- I'm sure this still leaves some things broken w.r.t. the overlays.  Shall I just merge, test, then open a new PR?

johnkit:
We'll merge this now and discuss additional features at next Tuesday's meeting. Additional features might include

* Hiding up/down arrows in spinboxes
* Revise widget layout to reduce vertical footprint
* Guard against re-launching image generation worker until current worker finished

cjh1:
> ok -- I'm sure this still leaves some things broken w.r.t. the overlays. Shall I just merge, test, then open a new PR?

@joelvbernier This PR was to added the ability to set eta min and max are there problems with the way it was implemented? Yes followup PRs would be good.

joelvbernier:
@cjh1 We need to hook these values into the definition of an eta period (branch cut) so the actual eta values coming out of the overlay generators in the polar view case fall inside the eta extents on the imshow.  I added the "slot" in the overlay classes (well, powder and laue so far) to take the period specification, which by definition will be `eta_min + 360.`, in #536.  Not merged yet, so yes, things will be "broken" to some extent (_e.g._, if I use (0, 360) for eta ranges and the overlays export points in (-180, 180)...)

@johnkit I think my form redesign in #476 is a good way to go (perhaps with replaceing the spin boxes with plain text input.  Although then we have to convert all the values to floats...).  And given the time it takes to redraw, I **_do not_** want the form trying to update the plot while I am inputting ranges -- especially since this is typically done for both the min and max.

cjh1:
@joelvbernier OK, can we update #536 to use the values added here?

PR #539, opened 2020-09-18
Resize bbox

bnmajor:
This branch adds additional support for the `PXRDIP` I/O:

- Allow the `PXRDIP` and `BBXRD` instrument boundaries to be resized if necessary. The original bounds are used as the maximum values.
- Adds support for transforms to be applied to the image after it has been cropped
- `PXRDIP` boundaries are rotated 90 degrees before being applied to the image and the cropped image is then rotated back -90 degrees


bnmajor:
The polar view looks like it is lining up correctly:
pxrdip-polar

bnmajor:
@joelvbernier This adds the rotation for the `PXRDIP` boundaries that we discussed

cjh1 (approved):
LGTM

PR #540, opened 2020-09-18
Average x for azimuthal integral plot powder overlays

psavery:
This fixes an issue where there are overlays for every x value,
which is problematic if the powder offset is not at the origin,
because the line overlays are curved.

There may be alternatives to averaging that may be preferred. But
we'll do that if we find a better alternative. For now, this fixes
the extremely slow GUI when the line overlays are curved.

example

@joelvbernier This should fix the slowness issue mentioned in https://github.com/HEXRD/hexrdgui/pull/527#issuecomment-694489708


cjh1 (approved):
Looks great, I love the interaction!

joelvbernier:
NICE.

PR #541, opened 2020-09-18
Overlay fixes for powder offset and eta period

psavery:
The changes are summarized below:

* Turn off keyboard tracking for powder offsets
* Update overlays when eta extents change
* Map angles to eta period for detector borders

See full commit messages for more details. Here is a before and after for the overlays and detector border fixes.

### Before:
before

### After:
after


cjh1 (approved):
LGTM

PR #546, opened 2020-09-22
Polar masks redraw

### LINKED ISSUE #535: Redraw of polarview unnecessary in mask application?
It seems that when we apply masks, it is actually redrawing the whole polarview.  This is unnecessary, and negatively impacts the UX as it takes a while.  Can't we cast the polar image as a masked array and simple update the mask with the logical and of the active mask layers?  This should play nice with the summing for the lineout.

bnmajor:
Speeds up mask creation in polar view - fixes #535


bnmajor:
Before:
before


After:
after

cjh1 (approved):
This speeds things up nicely in my testing. There are quite a few changes here, so let get it in so we can discover any issues.

PR #553, opened 2020-09-22
Draw all detectors for polar view powder overlays

psavery:
The other detectors were intentionally left out before since the
lines were assumed to all be vertical. Add them back in, since the
lines are not always vertical.

Here's an example image. It looks like there are still some vertical lines being drawn for some of the overlays.

image


joelvbernier:
Besides the dreaded dollar sign, I think I might also clip the rings to the respective detector extents...  Have to see how it looks at least.

joelvbernier (approved):

Issue #474, opened 2020-08-20
Enhancements to hand-drawn mask options


joelvbernier:
After using this a bit, I am requesting the following enhancements:

- [x] remove the picker window; this is not needed for picking mask region points
- [x] it seems to want to reset a zoomed view in the canvas (at least polar); is this necessary?
- [ ] want to add the ability to draw masks in the raw view
- [ ] store all masked regions in **raw pixel coordinates**; that way the selections will update in a more intuitive way when changing instrument parameters.  Even if the instrument parameters are off, the regions will still map consistently to the offending features in the raw data.

psavery:
For point 2, yes, it looks like the x-axis is being reset on the polar view when the user is zoomed in and any edits are made to the image. I'll look into why.

joelvbernier:
@psavery I assume the picker window slows things down a bit; for such coarse point picking, we can default to picks on the raw canvas without the zoom window.  I also wanted to modify the behavior of the picker window itself.

joelvbernier:
@psavery bumping this up in priority.  With regards to task 1, I still think just picking points on the image for the polygon masks is sufficient and likely faster (don't need the zoom).  There is also the desire to pick in the zoom window, as requested in #485 -- after playing with it more, this would be more intuitive for the users.

bnmajor:
@joelvbernier In working with the masking in each view, masks created in raw view can be applied to polar view without any issues, but applying masks created in polar mode to raw mode has some issues if masks overlap detectors or points are selected outside of a panel (which seems to be exactly what you were explaining in #556). What behavior would you like to see for now? Should raw masks be applied to polar but not vice versa (until we come up with a way to handle the case of selecting invalid points), or should masks only be applied to the view they were created in to keep behavior consistent?

cjh1:
@joelvbernier Can we get the answers to @bnmajor questions?

psavery:
@bnmajor Is this ready to be closed?

bnmajor:
I think this is related to #595 - I can look over that PR to make sure it's ready to merge, then this should be ready to close.

bnmajor:
Actually, looks like issue #556 goes with #595. I think this can be closed

PR #560, opened 2020-09-22
Raw drawn masks

bnmajor:
This has three parts:

- Does not use the zoom canvas when the hand drawn masks are created in polar view (this is a part of #474)
- Fixes a bug that caused lists of masks to be overwritten with the same key - affected laue masks and when multiple masks are drawn by hand
- Fixes a bug that reset visibility on masks when toggling between views

### DIFF
--- a/hexrd/ui/create_polar_mask.py
+++ b/hexrd/ui/create_polar_mask.py
@@ -14,12 +14,12 @@ def convert_raw_to_polar(det, line):
     return [np.degrees(tth)]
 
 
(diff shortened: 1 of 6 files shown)


bnmajor:
@psavery This should be a fix for the laue mask issue

psavery (approved):
This fixes the Laue mask issues for me, but I didn't try the other parts.

cjh1 (approved):
LGTM

Issue #554, opened 2020-09-22
Material selector for WPPF


saransh13:
Add an option in WPPF dialogue box to select which materials to use from the list of active materials.
The active materials should only have materials with "Powder" overlay option selected. The materials with "Laue" overlay can be ignored.

psavery:
It is currently only using the active material (the one that is currently selected in the materials panel). Get the user to select the materials.

psavery:
@saransh13 Do you think it automatically using all visible powder overlays is acceptable? Or do we want the selector?

saransh13:
I think we need the selector definitely. The default selection can have all materials with powder overlays, but an option to turn certain materials on/off is definitely required. In terms of passing this data to `LeBail` class, a simple list of these material objects should work.

PR #565, opened 2020-09-23
Add ability to select materials for WPPF

psavery:
This adds a button the WPPF options dialog that will pop up a dialog
where the user can select which materials to use for WPPF.

Only materials with powder overlays can be used. Materials that are
both visible and have powder overlays are selected by default. Materials
that have powder overlays but are not visible can be selected in the
dialog.

These materials then get passed on to WPPF. Previously, only the currently
selected material was passed to WPPF. Multiple materials can now be passed.

wppf_material_select

Fixes: #554


cjh1 (approved):
LGTM

PR #566, opened 2020-09-23
Add offset to avoid passing WPPF value <= 0

cjh1:
We add the offset and the remove it from the result.

### DIFF
--- a/hexrd/ui/calibration/wppf_runner.py
+++ b/hexrd/ui/calibration/wppf_runner.py
@@ -18,6 +18,7 @@ class WppfRunner(QObject):
     def __init__(self, parent=None):
         super().__init__(parent)
 
+        self._spec_offset = 0
         self.thread_pool = QThreadPool(self)
         self.progress_dialog = ProgressDialog(parent)
         self.progress_text.connect(self.progress_dialog.setLabelText)
@@ -91,6 +92,13 @@ class WppfRunner(QObject):
             # Re-format it to match the expected input format
             expt_spectrum = np.array(list(zip(*expt_spectrum)))
 
+        # Add offset to ensure we pass WPPF values greater than zero
+        min_value = expt_spectrum[:, 1].min()
+        max_value = expt_spectrum[:, 1].max()
+        # Generate the offset -min value + 0.5% of max value
+        self._spec_offset = -min_value + max_value * 0.005
+        expt_spectrum[:, 1] = np.add(expt_spectrum[:, 1], self._spec_offset)
+
         phases = [HexrdConfig().material(x) for x in dialog.selected_materials]
         kwargs = {
             'expt_spectrum': expt_spectrum,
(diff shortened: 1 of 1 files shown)


psavery:
Although I think the code looks good to me, and it fixes the error we were encountering with the tardis example that had negative pixels, I noticed that this change gets us noticeably different results for WPPF, and I think we need to look into it.

For our GE example, with all the default WPPF options, and 100 refinement steps, using CeO2 with the first 11 overlays, on the master branch, I get this:

before

Last iteration is: `Finished iteration. Rwp: 5.428 % goodness of fit: 1512556.952`

It looks like it matches up reasonably well.

However, after I switch to this branch, and perform the same steps as above, I get this instead:

after

Last iteration is: `Finished iteration. Rwp: 43.558 % goodness of fit: 90698285.791`

We should probably figure out what is causing such a great difference.

Any ideas, @saransh13 ?

psavery:
Note that the WPPF dialog options get saved in the QSettings, and I had to comment out this [`load_settings()`](https://github.com/HEXRD/hexrdgui/blob/b31832f4f2e3a6717ef7450c0304dfd46af5d4e2/hexrd/ui/calibration/wppf_options_dialog.py#L46) to get it to load the default settings.

cjh1:
@psavery Thanks for testing this out. When I testing it out this TARDIS data it didn't look quite right, but I assumed it was because I had my overlays messed up.

cjh1:
I will take another look at my logic and verify things on my end

saransh13:
@psavery Caan you post this dataset on the google drive? It seems to be working for the Ceria Dataset I have at hand.

psavery:
> @psavery Caan you post this dataset on the google drive? It seems to be working for the Ceria Dataset I have at hand.

@saransh13 Sure! It's in the shared google drive under `examples/calibration/ge`. The config file is `Hydra_Aug19.yml`, and the image files are `ge*.tiff`.

It's just this branch that has the issue. Master is fine.

saransh13:
@psavery I also did a PR in `hexrd` which catches the `nans` and `<=0.0` values in the intensity and ignores those points for performing the `WPPF` fits. That brach seems to work with this `Hydra` dataset without the need to apply offsets. In any case, even if you do apply offsets, then that PR will do nothing. 
Could merge that until we hunt down whats going on with this branch.

cjh1:
@saransh13 If WPPF can now handle negative values, we can just close this?

psavery:
@saransh13 Great! That sounds good. We can try that out.

I'm not sure where NaNs in the image are coming from. I know there have been NaNs in some of the actual image files we have worked with, but I am not sure if we are adding any or not.

One option we also have with regard to NaNs for the azimuthal integral data is to use [`np.nansum()`](https://numpy.org/doc/stable/reference/generated/numpy.nansum.html), which will sum columns and ignore NaNs. But this is only valuable if we care about other pixels in that column.

cjh1:
Closing this now, as the issue is now fixed in HEXRD

Issue #561, opened 2020-09-22
Specify refinements for Laue overlays


psavery:
Add the ability to specify whether or not various Laue overlay options should be refined in the calibration.

joelvbernier:
I'd like to see a table view of the parameters akin to the WPPF panel.

joelvbernier:
Also...  we need to expose the lattice parameter(s) for the case of powder diffraction overlays.  @saransh13, probably best to go through PlaneData class as PlaneData.lparms is already symmetrically reduced.  There is also logic in the material panel to cast the proper parameters names to the reduced list associated with the lattice type, a, b, c, &alpha;, &beta;, &gamma;.

joelvbernier:
Looks like [spacegroup](https://github.com/HEXRD/hexrdgui/blob/b31832f4f2e3a6717ef7450c0304dfd46af5d4e2/hexrd/ui/material_editor_widget.py#L124) knows how to parse the space group dependent number of reduced lattice parameters

psavery:
@joelvbernier Do you mean we need refinements for the lattice parameters?

The lattice parameter values themselves are currently exposed on the materials panel.

![Screenshot from 2020-09-23 20-51-20](https://user-images.githubusercontent.com/9558430/94088490-9ace4d80-fdde-11ea-9c41-ce7b8f2f67c0.png)

joelvbernier:
@psavery -- we need access to the lattice parameter only for the powder overlays.  It might be faster to just update the lattice parameter for the PlaneData object used to draw the overlays (and passed to the optimization) rather than modifying the material itself.  Not necessary for the Laue (Laue is insensitive to unit cell volume...)

psavery:
Here's what the refinement editor looks like in the powder overlay editor in #568 

![image](https://user-images.githubusercontent.com/9558430/94337731-ce65cf00-ffba-11ea-9b56-ba1740407ed0.png)

PR #568, opened 2020-09-24
Add ability to specify crystal refinements

psavery:
This adds a "refinement" tab to the calibration crystal editor to
allow the user to specify which crystal parameters can be refined.

It is currently being used only in the Laue overlay editor.

The refinement statuses are saved in the overlays themselves. They
are currently not being used for anything, but they will be used when
the spot picked calibration will be performed.

All parameters default to being refinable.

laue_refinements

Fixes: #561

### DIFF
--- a/hexrd/ui/cal_tree_view.py
+++ b/hexrd/ui/cal_tree_view.py
@@ -13,8 +13,8 @@ from hexrd.ui.tree_views.value_column_delegate import ValueColumnDelegate
 from hexrd.ui import constants
 
 # Global constants
-REFINED = 0
-FIXED = 1
+FIXED = 0
+REFINABLE = 1
 
 KEY_COL = BaseTreeItemModel.KEY_COL
 VALUE_COL = BaseTreeItemModel.VALUE_COL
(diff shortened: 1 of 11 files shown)


psavery:
A note on this: these checkboxes currently indicate "refinable", but the checkboxes in the tree view indicate "fixed". So they are currently opposite each other. Note that hexrd uses a "refinable" flag as well.

joelvbernier:
How involved would it be to swap the detector parameter checkboxes?  Perhaps we should to that and change the lab everywhere (including WPPF) to simply "Refine" -- checked meaning `True`

psavery:
> How involved would it be to swap the detector parameter checkboxes? Perhaps we should to that and change the lab everywhere (including WPPF) to simply "Refine" -- checked meaning `True`

We can do that - we would need to go to the tree view and change it to "Refine" instead of "Fixed". Or, we can just change it here to "Fixed" so it is consistent throughout the GUI. Which do you prefer?

psavery:
This PR is also now swapping "fixed" with "refinable" in the calibration tree view.

Additionally, there is now a refinements editor in the powder overlay editor:
image

cjh1 (approved):
LGTM

Issue #524, opened 2020-09-16
Add calibration with line pick for each overlay


psavery:
This will go through the overlays, one at a time, and have the user pick lines for each one. For Laue spots, we will need the user pick them in the correct order, so we may need to highlight the width for the overlay to be picked. We will also allow for skipping over certain Laue spots.

psavery:
@joelvbernier Is this going to replace the current line picker in "Run"->"Calibration Line Picker", or will it be a new menu item?

joelvbernier:
Yes; there will be a single workflow for calibration.  We will essentially build the problem by picking a list of material/overlay types.  Then we need to select the data points that will go into each.  We can probably include a toggle to the effect of "include in calibration" in the existing overlay manager.  To flag which materials will get used.

psavery:
Maybe another option is that we could make it so if the overlay is visible, it gets included in the calibration? I can still add another checkbox for "include in calibration" if that is not preferred.

psavery:
I went with "if the overlay is visible, then include it in the calibration", FYI.

PR #575, opened 2020-09-24
Add general calibration runner

psavery:
This generalizes the line picked calibration into a general calibration
runner, with several helpful features.

First of all, the calibration runner loops through all of the visible
overlays, and has the user pick points for each one. While the user
is picking points for an overlay, all of the other overlays are hidden,
and the line/spot of interest is highlighted.

For powder overlays, the user can move on to the next line by
right-clicking. For laue overlays, right-clicking is disabled and
left click automatically moves on to the next spot. If a spot in the
laue overlays needs to be skipped, the user may do so by middle clicking,
which inserts NaN's instead of regular values.

The user may click on the "Ok" button on the dialog to move on to the next
overlay. Alternatively, if there are no more points available for picking,
the dialog will automatically close itself.

Currently, nothing is being done with the picked data. But it will be fed
into a calibration function soon.

For powder overlays, the detector association may not be correct if the
user did not click on the right detector. To make things simpler, we should
just automatically determine which detector each point is on and associate
them in that way.

calibration_runner

Fixes: #524

### DIFF
--- a/hexrd/ui/calibration/calibration_runner.py
+++ b/hexrd/ui/calibration/calibration_runner.py
@@ -125,7 +125,22 @@ class CalibrationRunner:
         self.pick_next_line()
 
     def finish(self):
-        print(f'{self.all_overlay_picks=}')
+        # Temporary
+        import json
+
+        pick_results = []
+        for i, val in self.all_overlay_picks.items():
+            overlay = self.overlays[i]
+            pick_results.append({
+                'material': overlay['material'],
+                'type': overlay['type'].value,
+                'picks': val
+            })
+
+        out_file = 'calibration_picks.json'
+        print(f'Writing out picks to {out_file}')
+        with open(out_file, 'w') as wf:
+            json.dump(pick_results, wf)
 
     def set_exclusive_overlay_visibility(self, overlay):
         self.overlay_visibilities = [overlay is x for x in self.overlays]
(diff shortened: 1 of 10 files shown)


psavery:
@joelvbernier You can try this out now. It will dump out the calibration results into a `calibration_picks.json` file when you finish picking. The structure of the file looks like this:

```
[
    {
        "material": "CeO2",
        "type": "powder",
        "picks": {
            "ge1": [
                [
                    [x11, y11],
                    [x12, y12],
                    ...
                ],
                [
                    [x21, y21],
                    [x22, y22],
                    ...
                ],
                ...
            ],
            "ge2": [...]
        }
    },
    {
        "material": "ruby",
        "type": "laue",
        "picks": {
            "ge1": [
                [x1, y1],
                [x2, y2],
                ...
            ],
            "ge2": [...]
        }
    }
]
```

In the powder picks, each detector contains a list of lists of points. It is a list of lines essentially.

For the laue picks, each detector contains a list of points. It is just a list of Laue picks.

You can load this file into python like this:
```python
>>> import json
>>> with open('calibration_picks.json', 'r') as rf:
...     picks = json.load(rf)
```

Let me know if you have any issues.

joelvbernier:
Patrick, I am trying this right now

joelvbernier:
@psavery -- looks like this feature branch hasn't been pushed upstream?

psavery:
@joelvbernier No, it's just in this branch right now.

cjh1 (approved):

Issue #573, opened 2020-09-24
Overlay energy not updating when they become visible


psavery:
If you make a new overlay in the overlay manager for a material that is not the currently selected one in the materials panel, and you make it visible, its energy is not updated immediately with the current beam energy. This makes it so the overlay is drawn incorrectly.

We should make sure that as soon as an overlay becomes visible, its associated material gets an update to its energy. 

The reason there are lazy updates in the material energy is because it requires creating a new plane data object, which can be time-consuming.

PR #576, opened 2020-09-24
Update energy for all visible materials

psavery:
When a material becomes visible due to some changes in the overlay
manager, ensure its energy is updated. This prevents overlays
from having the wrong energy and being drawn incorrectly.

Fixes: #573


joelvbernier:
Good catch.

joelvbernier (approved):

cjh1 (approved):

Issue #570, opened 2020-09-24
Max 2θ in polar view not saved in between sessions


psavery:
If you load a config file, the polar resolution values are automatically computed and set. For instance, loading
the simulated tardis config and data yields this:

![image](https://user-images.githubusercontent.com/9558430/94160696-e3bfe980-fe52-11ea-8cee-4fb4aef50cc0.png)

However, if you exit the program normally (the settings will be saved), and then restart the program, the max
 2θ gets reset to 20, like here:

![image](https://user-images.githubusercontent.com/9558430/94160490-a8bdb600-fe52-11ea-8792-1e6cb4eb941f.png)

This is causing a few problems. First, the full dataset is not being displayed, but only a small subset of it. Secondly
the max 2θ that is being used in the materials panel is set based upon the polar resolution max 2θ. For this tardis example, because
the beam energy is very small, this results in no HKL values appearing in the materials table. This is likely what
caused the problem in https://github.com/HEXRD/hexrdgui/issues/500#issuecomment-698020628

PR #577, opened 2020-09-24
Save image config to QSettings

psavery:
This will result in the image config being preserved in between
hexrdgui runs. This is especially useful for cases where the
polar max two theta is severely truncated in between runs compared
to what it was before, which also results in a small max two theta
in the materials panel, which results in no active hkl values.

This also moves updating the snip1d enable states to a function
so we can also call it when the GUI gets updated.

Fixes: #570


cjh1 (approved):
LGTM

PR #578, opened 2020-09-24
Template line settings

### LINKED ISSUE #550: Allow color and line style of template outline to be selected
We can re use the selectors from else where.

bnmajor:
Fixes #550 

line-style


cjh1 (approved):
This looks great, I think in a follow up we probably at some point what to save the settings.

Issue #580, opened 2020-09-24
Update default material HDF5 file in resources


saransh13:
New material file includes `LiF, MgO, diamond, sapphire, quartz, ceria`.

joelvbernier:
Add LaB6, Silicon, and Forsterite (ambient Pnma) too please.

saransh13:
Sure.

Issue #497, opened 2020-09-04
Do lazy generation of PlaneData


psavery:
In #496, when the default materials are loaded, plane data gets generated on all of the materials during hexrd start-up.

This can be an expensive operation, so it delays start-up time.

At start-up, we should only generate plane data of the active material (the one currently selected in the materials panel), and materials that are visible. We can then generate plane data of other materials when they either become the active material or they become visible.

joelvbernier:
Agreed.

PR #581, opened 2020-09-24
Update default material file

saransh13:
Updated the default material file as suggested in issue #580 
The key for the default material was also updated from `ceo2` to `CeO2` to reflect changes in the `material.h5` file.


psavery:
So, one thing to note here is that currently, in the GUI, when the program starts, it generates new `PlaneData` objects for all of the materials, as this is part of the [Material object initialization](https://github.com/HEXRD/hexrd/blob/f1a7cbc198817e9e996a6dad31aeca09c6fab524/hexrd/material.py#L174), and it currently creates a material object for all the materials at start-up.

This `PlaneData` object generation is expensive, and the more materials we have in the file, the slower the GUI start-up time will be. It's good to have many materials in the HDF5 file. But we might need to prioritize some kind of lazy initialization of the material objects. This is being tracked in #497

saransh13:
Got it. I can think of strategies to do a quicker generation of `planeData`. Most of the time spent in the `Material` class generation is actually in the initialization of the `unitcell` class, which in turn is used to initialize `planeData`

psavery:
@saransh13 Okay. We might want to delay the unit cell class generation as well, then. I'll think of something.

cjh1:
Is there a copy of `materials.h5` in hexrd and hexrdgui?

psavery:
Looks like it is only in hexrdgui right now.

cjh1:
@joelvbernier @saransh13 Is this eventually going to live in hexrd?

saransh13:
@joelvbernier will prob make the final call, but having it in `heXRD` makes more sense to me.

cjh1:
@saransh13 Does this now include LiF, MgO, diamond, sapphire, quartz, ceria, as requested in #580 ?

saransh13:
@cjh1 yes, includes all of them, but @joelvbernier requested a few more additions (LaB6, Silicon, and Forsterite). I'll be pushing those changes today and then we'll be ready to merge.

saransh13:
I think we're ready to merge this now. The startup time was not wildly different.

psavery:
> I think we're ready to merge this now. The startup time was not wildly different.

Great!

cjh1 (approved):

PR #583, opened 2020-09-24
Overlay fixes

joelvbernier:
Fixed the powder overlays (no issue, just email thread).  Attached are the new simulation images with a BCC (sgnum 229, lparam 2.51Å) powder phase added at a 0.1625mm z-offset.  @saransh13 -- images here if you want them.
Screenshot from 2020-09-24 16-40-45

[Tardis_BCC_and_LiF_sim.zip](https://github.com/HEXRD/hexrdgui/files/5279772/Tardis_BCC_and_LiF_sim.zip)

### DIFF
--- a/hexrd/ui/overlays/powder_diffraction.py
+++ b/hexrd/ui/overlays/powder_diffraction.py
@@ -119,10 +119,6 @@ class PowderLineOverlay:
 
             # !!! must apply offset
             xys_full = panel.angles_to_cart(ang_crds, tvec_c=self.tvec)
-
-            # !!! distortion
-            if panel.distortion is not None:
-                xys_full = panel.distortion.apply_inverse(xys_full)
                 
             # clip to detector panel
             xys, on_panel = panel.clip_to_panel(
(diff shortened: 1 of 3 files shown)


psavery:
Looking good!

joelvbernier:
LGTM once checks finish.

hexrd/ui/overlays/powder_diffraction.py
```diff
@@ -137,20 +152,14 @@ def generate_ring_points(self, tths, etas, panel, display_mode):
...
-                # clip to detector panel
-                xys, on_panel = panel.clip_to_panel(
-                    xys_full, buffer_edges=False
-                )
+            elif display_mode in [ViewType.raw, ViewType.cartesian]:
 
                 if display_mode == ViewType.raw:
+                    # !!! distortion
+                    if panel.distortion is not None:
+                        xys_full = panel.distortion.apply_inverse(xys_full)
```
psavery, line 161: @joelvbernier So, is distortion only applied to raw and polar views? Looks like it isn't applied to Cartesian here.

psavery:
Looks like this for the ge example:

ge_example

Is it because one detector is half on the top and half on the bottom?

hexrd/ui/overlays/powder_diffraction.py
joelvbernier: Good catch -- should only be applied for the raw.

psavery:
Since the overlays are now clipped to the detectors in the polar view, I pushed an update so that if the detector transform is modified in the polar view, that it updates the overlays as well.

My ge example problem is now fixed, and this looks great to me!

psavery (approved):
LGTM

joelvbernier:
Took some doing, but this should work for all reasonable cases.

PR #595, opened 2020-09-28
Hand mask in raw

bnmajor:
This branch supports masking by hand in the raw view the same way that the polar view currently supports masking by hand. Right now masks must be completed on the current detector (either by right-clicking or selecting `Okay` from the dialog) in order to begin masking on a new detector.

Additionally, masking for Laue spots is not currently supported in the raw view, as the converted polar-to-raw coordinates do not seem to be correct (the masking does not line up with the overlays).

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -594,13 +594,12 @@ class MainWindow(QObject):
             HexrdConfig().polar_masks_changed.emit()
         elif self.image_mode == ViewType.raw:
             for det, line in zip(dets, line_data):
-                for line in line_data:
-                    name = unique_name(
-                        HexrdConfig().raw_mask_coords, 'raw_mask_0')
-                    HexrdConfig().raw_mask_coords[name] = line.copy()
-                    HexrdConfig().visible_masks.append(name)
-                    create_raw_mask(name, [(det, line)])
-                HexrdConfig().raw_masks_changed.emit()
+                name = unique_name(
+                    HexrdConfig().raw_mask_coords, 'raw_mask_0')
+                HexrdConfig().raw_mask_coords[name] = [(det, line.copy())]
+                HexrdConfig().visible_masks.append(name)
+                create_raw_mask(name, [(det, line)])
+            HexrdConfig().raw_masks_changed.emit()
         self.new_mask_added.emit(self.image_mode)
 
     def on_action_edit_apply_laue_mask_to_polar_triggered(self):
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -559,14 +559,23 @@ class MainWindow(QObject):
         self._apply_drawn_mask_line_picker.finished.connect(
             self.run_apply_polar_mask)
 
-    def run_apply_polar_mask(self, line_data):
-        for line in line_data:
-            name = create_unique_name(
-                HexrdConfig().polar_masks_line_data, 'polar_mask_0')
-            HexrdConfig().polar_masks_line_data[name] = line.copy()
-            HexrdConfig().visible_masks.append(name)
-            create_polar_mask([line.copy()], name)
-        HexrdConfig().polar_masks_changed.emit()
+    def run_apply_polar_mask(self, dets, line_data):
+        if self.image_mode == ViewType.raw:
+            for det, line in zip(dets, line_data):
+                name = create_unique_name(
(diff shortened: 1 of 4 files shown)


joelvbernier:
@bnmajor -- the Laue mask coordinates can come directly form the laue overlays (_i.e._, the raw view output)

joelvbernier:
@bnmajor -- looks like some conflicts here.  Shall we shelve this until after the demo?

cjh1:
@joelvbernier Yes, the plan was to wait.

joelvbernier:
We should bump this one back up -- we can use this functionality to create masks that can be used as panel buffers.

joelvbernier:
Yo @bnmajor -- let's bump this back up.  I think that simply feeling the raw row/col vertices into `skimage.draw.polygon` gives us exactly what we need (I've been doing that manually IPython outside the GUI).  The results of the call to `polygon` yield a boolean array the same shape as the image with the interior points marked `True`; therefore the contribution to the mask (`panel_buffer` attribute on each detector) is just the negation of that array.

To do:

- [ ] Verify that the vertex list has to be ccw to mark points in the interior, and update GUI help (tooltip?) accordingly
- [ ] Use `logical_and` to intersect result with existing `panel_buffer` attributes, including handling if they are either not defined or specified with a "frame width"; in the former case initialize the `panel_buffer` as `np.ones((panel.rows, panel.cols), dtype=bool)` and intersect with `~polygon_mask`, and in the latter we will have to convert the frame width spec to a boolean mask.
- [ ] Probably needs a widget?  Or do you think we can use the existing raw view for the drawing?

Happy coding...

bnmajor:
@joelvbernier I've rebased this off of master and the hand drawn masking seems to be working as expected, but if you have a chance to play with this a little it would be good to make sure this matches your expectations/needs. Also if there any datasets or specific use cases that this should be tested with please let me know and I'll test them out

bnmajor:
@joelvbernier This is just waiting on your feedback to determine if this handles all of the use cases you had in mind. Master already supports masking with rectangles/ovals in raw mode, this just adds the option to hand-draw masks in the raw view as well.

bnmajor:
@joelvbernier This branch is now up to date with master and supports importing and exporting masks that have been hand drawn in the raw view

hexrd/ui/main_window.py
```diff
@@ -563,24 +563,33 @@ def on_action_edit_euler_angle_convention(self):
...
+        self._apply_drawn_mask_line_picker = (
             HandDrawnMaskDialog(canvas, self.ui))
-        self._apply_polar_mask_line_picker.start()
-        self._apply_polar_mask_line_picker.finished.connect(
+        self._apply_drawn_mask_line_picker.start()
+        self._apply_drawn_mask_line_picker.finished.connect(
             self.run_apply_polar_mask)
 
-    def run_apply_polar_mask(self, line_data):
+    def run_apply_polar_mask(self, dets, line_data):
```
psavery, line 575: Can we rename this to `run_apply_hand_drawn_mask`?

hexrd/ui/main_window.py
```diff
@@ -563,24 +563,33 @@ def on_action_edit_euler_angle_convention(self):
...
-        for line in line_data:
-            name = unique_name({**raw_lines, **polar_lines}, 'polar_mask_0')
-            HexrdConfig().polar_masks_line_data[name] = [line.copy()]
-            HexrdConfig().visible_masks.append(name)
-            create_polar_mask([line.copy()], name)
-        HexrdConfig().polar_masks_changed.emit()
+        if self.image_mode == ViewType.raw:
+            for det, line in zip(dets, line_data):
+                name = unique_name({**raw_lines, **polar_lines}, 'raw_mask_0')
+                HexrdConfig().raw_masks_line_data[name] = [(det, line.copy())]
```
psavery, line 581: I guess we need to update some of these variables that changed via #1044?

psavery (commented):
I'll test this out later

hexrd/ui/main_window.py
bnmajor: Yes, renamed

hexrd/ui/main_window.py
bnmajor: Yes, I've merged master in and cleaned up the resulting conflicts.

psavery (approved):
LGTM. Worked very well when I tested it!

Issue #588, opened 2020-09-25
Update lattice parameter for Materials once WPPF finishes running


saransh13:
There's two thing that we'd want to improve for the WPPF form. 
1. The first one is to create the form with all parameters in the `WPPF.LeBail.params` class. This is an instance of `WPPF.Parameters` class. Currently, it it initialized I imagine from the parameters in the `yaml` file and misses the lattice parameters in the form. The lattice parameters are not specified in that file but read from the `Material`.
2. Once WPPF is done running, we need to take the refined lattice parameters from the `WPPF.LeBail.params` class and update the lattice parameters of the material class.

PR #596, opened 2020-09-28
Add extra WPPF params and update materials from output

psavery:
This adds the extra refinement options for the materials' lattice parameters to the WPPF options table,
and it also updates the material objects in hexrdgui from the output of WPPF.

wppf_extra_params

Fixes: #588


cjh1 (approved):
LGTM

PR #602, opened 2020-09-29
Only allow unique powder params to be refinable

psavery:
It doesn't make sense to display non-unique lattice parameters
as being refinable in the powder overlay editor.

Only display unique powder parameters.

Sapphire, for example:

image


cjh1 (approved):
LGTM

Issue #582, opened 2020-09-24
Structure tab issues


joelvbernier:
If I try to add a new material, I'm noticing a few things.

1. The site labels are not editable; they seem to grab a string and advancing number from the last selected material
2. Site atom types are not editable; no option to remove a site if you add one accidentally.  In fact, if I try to change the atom site, it just keeps adding additional ones.  Maybe I am not understanding the interface?  Must I deselect in the periodic table?
3. If I delete a material, the structure tab is stuck to what ever the last edits were instead of some defaults.

![add_new_material-2020-09-24_15 47 13](https://user-images.githubusercontent.com/1154130/94209208-965a7100-fe80-11ea-9bc6-ed555f12a72d.gif)

psavery:
For 1, try two things:

1. Try double-clicking the site label. It works for me on Linux, but Saransh said it doesn't work on Mac.
2. You can also just select the row and start typing, and it should start editing it. 

See example below.

![edit_sites](https://user-images.githubusercontent.com/9558430/94210395-98cac400-fe9c-11ea-8abc-f4442c816afb.gif)

For 2, yeah, the only way to removes atoms is to deselect them from the periodic table. If they are in yellow, it means they are selected. I can add a remove button sometime soon to make it more intuitive.

For 3, this is something I need to fix.

joelvbernier:
Sounds good @psavery

psavery:
Let me know if you can edit the site names by selecting them and typing.

joelvbernier:
@psavery -- editing site labels also doesn't work on windows.

psavery:
Yes, I can use generic site names like `Site1`, `Site2`, etc.

PR #603, opened 2020-09-29
Create QLineEdit directly to avoid issue on Windows

cjh1:
For some reason on Windows if we call super().createEditor(user_type, parent) it doesn't create a line edit, so create it directly.


cjh1:
Fixes #582

psavery (approved):
LGTM

Issue #557, opened 2020-09-22
Allow log scale to be applied to intensities in the polar view 


cjh1:

joelvbernier:
The lineout in the lower axes, that is.

psavery:
I think we'll need to just perform a log on the y-axis of the azimuthal data generated [here](https://github.com/HEXRD/hexrdgui/blob/b31832f4f2e3a6717ef7450c0304dfd46af5d4e2/hexrd/ui/image_canvas.py#L521) and [here](https://github.com/HEXRD/hexrdgui/blob/b31832f4f2e3a6717ef7450c0304dfd46af5d4e2/hexrd/ui/image_canvas.py#L601).

cjh1:
Just log scale the y axes

PR #605, opened 2020-09-29
Apply log scale to y axes of azimuthal integral plot

cjh1:
Fixes: #557

### DIFF
--- a/hexrd/ui/image_canvas.py
+++ b/hexrd/ui/image_canvas.py
@@ -8,6 +8,7 @@ from matplotlib.backends.backend_qt5agg import FigureCanvas
 
 from matplotlib.figure import Figure
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


cjh1:
Fixes #557

hexrd/ui/image_canvas.py
```diff
@@ -649,6 +649,9 @@ def update_azimuthal_integral_plot(self):
         axis.relim()
         axis.autoscale_view(scalex=False)
 
+    def set_azimuthal_integral_plot_y_scale(self, scale):
+        self.azimuthal_integral_axis.set_yscale(scale)
```
psavery, line 653: I think you should check to see if `self.azimuthal_integral_axis` is `None` first. It will be `None` if the image mode is Cartesian or Raw.

psavery (approved):
LGTM

PR #606, opened 2020-09-29
Use "Site1", "Site2", etc. for site names

psavery:
Instead of automatically naming sites based upon the elements
they contain, name them "Site1", "Site2", etc.

Fixes: #582


cjh1 (approved):
LGTM

PR #612, opened 2020-09-29
Obtain and save more information in calibration runner

psavery:
Here is the information that is being saved now after the calibration picks are complete:

1. All overlay options get written out to "calibration_picks.json" as well, under the "options" key. Overlay options include powder overlay offsets, and Laue crystal parameters, and more information too such as the laue widths, min/max energy, etc.
2. All overlay refinements get written out to "calibration_picks.json" as well, under the "refinements" key. Refinements consist of a list of (label, bool) pairs, where they should be sorted in the proper order. The powder refinements should only contain the required lattice parameters, in the right order. And, the laue should have the refineable options in the same order as the crystal parameters.
3. All materials used (including their plane data) are dumped out as "<material_name>.pkl" files, and the instrument gets dumped out as "instrument.pkl".
5. The general refinement flags are dumped as well into "refinement_flags.json".
6. The Laue hkls are saved as well in "calibration_picks.json", under the "hkls" key. It is a dict where the key is the detector, and they should be in the same order as the picks. However, the "hkls" might be a little longer than the picks if the user finished the picks early by pressing the "Ok" button.

### DIFF
--- a/hexrd/ui/calibration/calibration_runner.py
+++ b/hexrd/ui/calibration/calibration_runner.py
@@ -141,6 +141,7 @@ class CalibrationRunner:
                 'material': overlay['material'],
                 'type': overlay['type'].value,
                 'options': overlay['options'],
+                'refinements': overlay['refinements'],
                 'picks': val
             })
 
--- a/hexrd/ui/calibration/calibration_runner.py
+++ b/hexrd/ui/calibration/calibration_runner.py
@@ -128,19 +128,26 @@ class CalibrationRunner:
         # Temporary
         import json
 
+        class NumpyEncoder(json.JSONEncoder):
+            def default(self, obj):
+                if isinstance(obj, np.ndarray):
+                    return obj.tolist()
+                return json.JSONEncoder.default(self, obj)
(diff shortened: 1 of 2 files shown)


psavery:
@joelvbernier I think this should be everything except for the hkls for the powder. I haven't figured out yet how to determine in [`powder_diffraction.py`](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/overlays/powder_diffraction.py) which hkls are actually drawn for each detector (if a powder line doesn't intersect a detector, it isn't included in the data, and we should therefore not include the hkl either).

psavery:
Here's an example of loading one of the pickle files from the command line:
```python
>>> import pickle
>>> with open('instrument.pkl', 'rb') as rf:
...     instrument = pickle.load(rf)
```

That should contain the instrument object.

psavery:
Note that when everything is hooked up properly, we won't be dumping out to files anymore, we'll just be passing the objects directly to the functions.

joelvbernier (approved):

Issue #607, opened 2020-09-29
Polar view not redrawing after changing instrument parameters in Instrument Tree or Form


joelvbernier:
Just noticed this.  The overlays change, but the polar image doesn't get updated if you change instrument parameters in the Tree or Form views, just the Slider View.  Simple fix I am sure.

psavery:
Which parameters were you changing? The beam vector? It looks like I can reproduce this issue for the beam vector...

joelvbernier:
Any.

psavery:
Yeah, I get the same experience here. Looks like if you have "Show Live Updates" enabled, everything causes an update for me except the beam vector.

joelvbernier:
Why is that treated differently that the other parameters with regards to the update?

psavery:
Not sure, I'll take a look

psavery:
So, the raw view and the cartesian view don't need to be redrawn when the beam vector changes, but only the overlays need to be updated. We were accidentally extending that to the polar view as well for the form and tree views. #614 should now always update the polar view no matter where the beam vector gets changed.

PR #614, opened 2020-09-29
Always update polar view when beam vector changes

psavery:
The polar view was only being updated for changes to the beam vector
in the slider widget, but not for changes to the beam vector in the
form view or tree view.

It now updates for all three views. There is also a timer attached to
it so that if the slider view is being used, it doesn't trigger too
many updates.

Fixes: #607


cjh1 (approved):
LGTM

Issue #592, opened 2020-09-25
Overlays in raw mode not updating with change of instrument parameters


joelvbernier:
Could be a Mac specific issue?  I am running in a clean env with our special python 3.8.4 on 10.15.7

joelvbernier:
@saransh13 confirmed the same behavior on 10.14

psavery:
This might be a mac-specific issue. I don't think I can reproduce it on Linux.

@saransh13 @joelvbernier If you are still experiencing it, can one of you send a detailed example and/or a gif of how to reproduce it?

joelvbernier:
Happening on windows too.  Overlays do no update while in the raw view if you change any of the instrument parameter sliders.  Polar view updates, but raw only updates (_i.e._, the overlays) if you re-select the raw view tab after switching to polar or cartesian (where applicable) the switching back.

saransh13:
![out](https://user-images.githubusercontent.com/15834451/94586709-67ece500-0236-11eb-9a32-7d868f6b84c3.gif)

PR #615, opened 2020-09-29
Update overlays in raw view mode on detector edit

psavery:
When the detector transforms get modified, update the overlays
in the raw view.

This is a little slow to use in the slider view, because it is
currently re-generating overlays for all of the raw images (not just
the one that got changed). But maybe we can fix that in the future...

Fixes: #592


cjh1 (approved):
LGTM

PR #616, opened 2020-09-29
Give the zoom canvas an expanding size policy

psavery:
By default, get the zoom canvas to expand the most that it can.

This makes resizing the line picker dialog actually useful.


cjh1 (approved):
LGTM

Issue #485, opened 2020-08-24
Picker window enhancements


joelvbernier:
After playing with this a bit, I think some minor modifications might enhance the picker window utility.  I think that it will be most useful in a "two-click" usage case, where you want to do the picking on a zoom canvas for more accuracy.  This is how the venerable Fit2d software worked (albeit in a less fancy package):

![fit2d_2-click-00 00 39 779-00 01 03 267](https://user-images.githubusercontent.com/1154130/91100122-77ff1c80-e619-11ea-9408-f5961e1004f2.gif)

Maybe we offer the one-click/two-click option to toggle between picking points on the raw canvas and using the existing zoom window to refine picks (as above).

Issue #242, opened 2020-02-03
Two-click mode for Calibration Line Picker


psavery:
Add an option in the calibration line picker for a two-click mode, in which you click once on the main image, then it locks the position of the zoom canvas, and then you click a second time in the zoomed canvas. This allows for finer point picking.

cjh1:
@joelvbernier Is this still something we would like to add? It couple of days work.

PR #617, opened 2020-09-29
Add two click option to line picker dialog

psavery:
If enabled, the two click option will have the user first click on
the main canvas, at which point the zoom canvas will be frozen. The
user must then click on the zoom canvas to add the point.

It allows for finer point picking.

two_pick_mode

Fixes: #242
Fixes: #485


cjh1 (approved):
Works well! Thanks!

Issue #618, opened 2020-09-29
Labels for masking options are swapped


joelvbernier:
In the Edit &#8594; Masks menu, the labels for options "Apply Mask to Polar" and "Apply Polygon Mask" are a bit misleading.  The former is in fact the hand-drawn polygonal mask, and the latter is the mask by shape (ellipse or rectangle).

I suggest the following renames in the ui form:

- **Apply Mask to Polar &#8594; Apply Polygon to Polar**
- **Apply Polygon Mask &#8594; Apply Ellipse/Rectangle to Polar**

Sound good?

bnmajor:
@joelvbernier Sounds good. `Apply Polygon Mask` is available in raw view as well though, would something like `Apply Ellipse/Rectangle Mask` be okay? Or should masking by polygon be toggled off in raw view for now?

bnmajor:
#619 fixes this using `Apply Ellipse/Rectangle Mask` instead of `Apply Ellipse/Rectangle to Polar` but this can be tweaked if needed, very quick fix.

PR #620, opened 2020-09-30
Masking bugs

bnmajor:
In testing the label changes in #618 I discovered two bugs, one of which broke the masking by hand in polar view. The updates to the `LineBuilder` class for the two-click line calibration broke the way masking by hand worked. For a quick a simple fix this adds the old `LineBuilder` class back in, but this can be fixed in the near future.

Below is a gif of all of the masking options in both polar and raw view being used successfully now, including when views are toggled.
masking


cjh1 (approved):
LGTM, @bnmajor Can you raise a issue to remove the duplicate code/refactor in the future.

bnmajor:
Yes, added an issue in #622

psavery:
FYI, the line builder is a tad bit different now, as of https://github.com/HEXRD/hexrdgui/pull/617

It was much simpler for me to work with in this way. But it doesn't do any more mpl_connect's or anything like that. The owning object forwards the button events to it.

psavery:
Actually, I see that you mentioned that in the PR message now. Sorry, I didn't realize the line builder was used anywhere else.

PR #621, opened 2020-09-30
Keep zoom canvas cmap and norm up to date

psavery:
Keep the zoom canvas cmap and norm in sync with the main
canvas cmap and norm, so that if the user modifies the main
canvas cmap and norm, the zoom canvas gets updated also.

Hopefully, this won't slow down the performance much...


cjh1 (approved):
LGTM

PR #625, opened 2020-09-30
Add y integration and extended cursor to zoom canvas

psavery:
This adds a y integration plot to the zoom canvas on the right side of it, to assist with picking Laue spots. It also adds a cursor when the mouse is over the canvas, and the cursor is extended to the two integration axes.

The crosshairs are also removed when the mouse is over the zoom canvas.

new_zoom_features


cjh1 (approved):
Awesome interaction! Couldn't it as the power is out here!

PR #635, opened 2020-10-02
Readme updates

bnmajor:
Add separate conda instructions for each platform and clarify the differences between the release options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,10 @@ set CONDA_BUILD=1
 pip install --no-deps -U -e hexrdgui
 ```
 
+If you are running in Windows PowerShell or other environments where the stdout
+and stderr is not appearing in the console you can run the python module directly
+`python hexrdgui/hexrd/ui/main.py`, you should then see stdout and stderr.
+
 # Install
 
 ## conda (release)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,15 +45,30 @@ pip install --no-deps -U -e hexrd
...
+### Windows
+```bash
+# First, make sure python3.8+ is installed.
+# If it is not, run the following command:
+conda install -c anaconda python=3.8
+# Install deps using conda package
+conda install -c HEXRD -c cjh1 -c anaconda -c conda-forge hexrdgui
+# Now using pip to link repo's into environment for development
+pip install --no-deps -U -e hexrd
+BUILD_CONDA_BUILD=1
```
cjh1, line 57: We need `set` here

README.md
```diff
@@ -32,15 +45,30 @@ pip install --no-deps -U -e hexrd
 CONDA_BUILD=1 pip install --no-deps -U -e hexrdgui
```
cjh1, line 45: We can remove the first python install, only need one!

README.md
bnmajor: Sorry about that, typed the word I was looking at on accident... Fixed.

README.md
bnmajor: Oops, overlooked that. Fixed.

cjh1 (approved):
Thank you!

PR #638, opened 2020-10-08
Update pick_based_calibration.py

joelvbernier:
switch to stored hkls, fixes length mismatch issues when skipping picks.  Tested with Laue and powder.


joelvbernier:
Seems to be working for me.  LGTM

cjh1 (approved):

PR #644, opened 2020-10-13
Use right-click to insert NaNs for Laue spots

psavery:
Rather than forcing the user to middle-click (which some users cannot do, if
they do not have a middle mouse button), have right-click perform the
functionality.

For powder lines, left-click adds a point, and right-click moves on to the next
line.

For Laue spots, left-click adds a point, and now right-click insert NaNs.

Middle-click no longer does anything.


cjh1 (approved):
LGTM

PR #645, opened 2020-10-20
Add ruby back to the default materials.h5 file

psavery:
This was copied from the ruby that was in there before (added in 005b6b488c30d346a3e53d5bfc8e9b3c890ab87a).

It is needed for some of the indexing tests.


cjh1 (approved):
LGTM

PR #647, opened 2020-10-21
Update install to get fabio for HEXRD channel

cjh1:
@joelvbernier This updates the instructions to use a update version of fabio with the TIFF fix.


psavery (approved):
LGTM. I also gave Linux a try and it worked for me.

Issue #429, opened 2020-07-21
Visualize the grain fitting results


cjh1:
@joelvbernier Is going to provide some code examples.

joelvbernier:
Ok -- first let me tabulate the anatomy of the `grains.out` file.  There are 21 columns; they represent:
```
0    grain ID
1    completeness
2    chi^2
3    exp_map_c[0]      
4    exp_map_c[1] 
5    exp_map_c[2]       
6    t_vec_c[0]
7    t_vec_c[1]
8    t_vec_c[2]
9    inv(V_s)[0,0]
10   inv(V_s)[1,1]
11   inv(V_s)[2,2]
12   inv(V_s)[1,2]*sqrt(2)
13   inv(V_s)[0,2]*sqrt(2)
14   inv(V_s)[0,2]*sqrt(2)
15   ln(V_s)[0,0]
16   ln(V_s)[1,1]
17   ln(V_s)[2,2]
18   ln(V_s)[1,2]
19   ln(V_s)[0,2]
20   ln(V_s)[0,1]
```
The 12 parameters in the slice `[:, 3:15]` are the grain parameters that are used by the various models and methods, 3 orientation parameters, 3 center-of-mass coordinates (x, y, z), and 6 strain parameters as the scaled components of the inverse stretch tensor in the sample frame (not particularly useful by itself...).  For convenience, I convert those last 6 grain parameters to strain tensor components in the sample frame in the last 6 columns.

For a start, we will want to use the COM coords for a 3-d scatter plot (cols `[:, 6:9]`), where the colormap is selectable from the following scalars:
```
[:, 0]      completeness ratio
[:, 1]      goodness of fit
[:, 3:6]    grain orientation *
[:, 15]     xx strain
[:, 16]     yy strain
[:, 17]     zz strain
[:, 18]     yz strain
[:, 19]     xz strain
[:, 20]     xy strain
[:, 15:21]  equivalent strain *
```
the `*` quantities require calculation.  I'll have to get you a colormap generator for orientation -- it depends on the symmetry group of the material (@saransh13: do you have code to do this?).  The equivalent strain is a simple calculation:
```
from hexrd.matrixutil import vecMVToSymm

ngrains = len grains_table  # loadtxt on grains.out
eqv_strain = np.zeros(ngrains)
for i, grain in enumerate(grains_table):
    emat = vecMVToSymm(grain[15:21], scale=False)
[...]

PR #660, opened 2020-11-17
Fix equivalent stress computation

psavery:
This is being done in response to https://github.com/HEXRD/hexrdgui/issues/429#issuecomment-725693239

2/3 is now 3/2. Now I just need to figure out how to make `m` the deviator of the stress...


cjh1 (approved):

psavery:
The deviator stress is now also being computed in response to https://github.com/HEXRD/hexrdgui/issues/429#issuecomment-725693239 and https://github.com/HEXRD/hexrdgui/issues/429#issuecomment-729683718

cjh1 (approved):
LGTM

Issue #665, opened 2020-12-08
Add units for stiffness tensor


cjh1:
gigapascal (GPa). This can be added a a tooltip.

psavery:
IIRC, for compliance that will be added in #666, the units will be inverse terapascal (TPa⁻¹). We will update the tooltip when the tensor type gets changed.

PR #684, opened 2020-12-14
Add elastic compliance as an editing option

psavery:
This is the inverse of the stiffness tensor.

It has been added, along with valid/invalid matrix states, since
the elastic tensor is invalid if it is not invertible.

Units have also been added.

example

This PR is complete, but I think the we need to examine the units in hexrd [here](https://github.com/HEXRD/hexrd/blob/ca88d475f72be177141a3ff028df90cd771ea59d/hexrd/unitcell.py#L1293-L1302) to make sure this is correct. I think the units of the stiffness and compliance tensors are currently in GPa and GPa⁻¹ (respectively), when they should be GPa and TPa⁻¹.

Fixes: #665 
Fixes: #666 

For correct compliance units in TPa⁻¹, this PR depends on hexrd/hexrd#141.

### DIFF
--- a/hexrd/ui/material_properties_editor.py
+++ b/hexrd/ui/material_properties_editor.py
@@ -1,5 +1,6 @@
 import copy
 import numpy as np
+from numpy.linalg import LinAlgError, inv
 
 from PySide2.QtCore import QSignalBlocker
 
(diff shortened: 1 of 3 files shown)


psavery:
@pep8speaks Doesn't recognize the walrus operator yet. This is actually pep8 compliant.

cjh1 (approved):
LGTM

Issue #674, opened 2020-12-08
Add ability to change color map in fit grains result view


cjh1:
We should be able to reuse code from where we do this elsewhere.

bnmajor:
Added in #686 - Closing, can re-open if needed

PR #686, opened 2020-12-21
Add ability to change color map in fit grains

bnmajor:
Addresses #674

### DIFF
--- a/hexrd/ui/indexing/fit_grains_results_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_results_dialog.py
@@ -57,6 +57,7 @@ class FitGrainsResultsDialog(QObject):
         self.ui.splitter.setStretchFactor(1, 10)
 
         self.setup_tableview()
+        self.load_cmaps()
 
         # Add column for equivalent strain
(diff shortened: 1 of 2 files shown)


psavery:
This looks great, and works well! But I think there's one thing that might make it a little nicer. Notice the combo options is a big list that covers a lot of the screen:

big_list

For the color map editor in the main window, it is a small list, that only displays 5 items at a time:

small_list

It might be nice to make it a small list as well for the one in the grain fitting results dialog. I think we did this by editing the styleSheet and maxVisibleItems in Qt designer for the combo box. You can see the options we set for the main window one here:

Screenshot from 2020-12-22 11-32-12

bnmajor:
@psavery It looks like the `maxVisibleItems` is currently set to 10, but is still showing that massive list (a quick test lowering that value seemed to make no difference). I could be missing something else though, I'll check it out.

psavery:
@bnmajor Yeah, I don't know why we have to edit the style sheet as well. But we do!

psavery (approved):
LGTM

Issue #671, opened 2020-12-08
Add option to change the glyph size in the fit grains results view


cjh1:

bnmajor:
Added in #687 - Closing, can re-open if needed

PR #687, opened 2020-12-21
Allow user to resize glyphs

bnmajor:
Addresses #671

### DIFF
--- a/hexrd/ui/indexing/fit_grains_results_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_results_dialog.py
@@ -57,7 +57,7 @@ class FitGrainsResultsDialog(QObject):
         self.ui.splitter.setStretchFactor(1, 10)
 
         self.setup_tableview()
-        self.set_glyph_size(reset=False)
+        self.reset_glyph_size(update_plot=False)
 
         # Add column for equivalent strain
         eqv_strain = np.zeros(self.num_grains)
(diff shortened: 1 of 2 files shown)


bnmajor:
This currently uses a slider to adjust the size, but could easily be updated to use a spinbox instead if that makes more sense.

glyph_size

cjh1 (approved):

hexrd/ui/indexing/fit_grains_results_dialog.py
```diff
@@ -442,6 +445,12 @@ def update_range_constraints(self):
             w1.setMaximum(w2.value())
             w2.setMinimum(w1.value())
 
+    def set_glyph_size(self, reset=True):
```
psavery, line 448: Maybe this should be called `reset_glyph_size`? With the `reset` argument being called `update_plot`?

psavery:
Can we also add tab stops for the new widgets?

psavery:
I tested it out and works well for me.

hexrd/ui/indexing/fit_grains_results_dialog.py
bnmajor: Sounds good, makes it clearer. I've updated the name of the function and parameter.

bnmajor:
> Can we also add tab stops for the new widgets?

Updated the tab stops - Thanks for pointing that out, @psavery!

psavery (approved):
Thanks @bnmajor! Looks good to me.

Issue #682, opened 2020-12-08
Provide view option to display/edit fit grain options in a tree view


cjh1:

psavery:
Indexing options also, I think, as a separate tree view.

PR #689, opened 2020-12-22
Edit indexing and fit grains configs via tree view

psavery:
This adds editing for indexing and fit grains configs via a tree
view. These are located under the "View" menu item.

Right now, you can edit all of the values in the right column,
which lets you edit most of the config options. However, you cannot
edit the keys. For the seed search method, you need to be able to
change the key in order to change the method (blob log, blob dog, and label), 
and this in turn changes children keys. This is not supported currently, and you
can only currently edit values.

Also, we should potentially remove the `hkl_seeds` key from the displayed config, because that
is set by the selected hkl values in the materials table. There's also no way to add or remove
items from lists in the fit grains config editor.

Screenshot from 2020-12-22 10-59-01

example

Fixes: #682


psavery:
I added functionality to do the following:

1. Hide items in the config (this is being done for the hkl_seeds, because they are set automatically)
2. Insert/remove items from a list via context menu (needed for list items in the fit grains config editor)
3. Change combo keys (needed for seed search method in indexing config editor)

You can see the results in the gifs attached.

list_editing

seed_search_method_change

I think this is mostly everything we needed in a tree view editor. And I think it's ready for review.

cjh1 (approved):
LGTM

joelvbernier:
Very nice!  Although there is a requirement that the tolerances (tth, eta, one) all have the same length… there are asserts to catch it in the code.

Issue #677, opened 2020-12-08
Allow sort of fit grains table by t_vec_XXX columns


cjh1:

psavery:
We might just need to add these columns [here](https://github.com/HEXRD/hexrdgui/blob/8c006bf33432c084cd5cf6de7b5d50927a590e9c/hexrd/ui/indexing/fit_grains_results_dialog.py#L346-L352) and [here](https://github.com/HEXRD/hexrdgui/blob/8c006bf33432c084cd5cf6de7b5d50927a590e9c/hexrd/ui/indexing/fit_grains_results_dialog.py#L168-L175).

PR #693, opened 2020-12-26
Allow sort of fit grains table by t_vec_X columns

psavery:
This allows sorting in the first three columns and the
t_vec_X columns. Sorting is not allowed in the other columns.

Fixes: #677


cjh1 (approved):
LGTM

Issue #675, opened 2020-12-08
Add option to export stress


cjh1:

psavery:
This can be a button that is only enabled when `Convert Strain to Stress` is checked.

PR #694, opened 2020-12-26
Add export stresses button in fit grains results

psavery:
This button is only enabled if strain is being converted to
stress. This button will save out the stresses to a compressed
npz file, under the "stresses" key. It is saved as Mandel-Voigt
vectors, and the shape is (n, 6), where n is the number of grains.

Fixes: #675

### DIFF
--- a/hexrd/ui/indexing/fit_grains_results_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_results_dialog.py
@@ -1,6 +1,7 @@
 import copy
 from functools import partial
 import os
+from pathlib import Path
 import sys
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


hexrd/ui/indexing/fit_grains_results_dialog.py
```diff
@@ -168,6 +169,25 @@ def on_export_button_pressed(self):
...
+            raise Exception('Tensor type must be stress')
+
+        selected_file, selected_filter = QFileDialog.getSaveFileName(
+            self.ui, 'Export Fit-Grains Stresses', HexrdConfig().working_dir,
+            'Npz Files (*.npz)')
+
+        if not selected_file:
+            return
+
+        stresses = self.converted_data[:, 15:21]
```
cjh1, line 183: It would be useful to add a comment or define constants for 15 and 21, to make it clear what we a slicing here.

cjh1 (approved):
LGTM

hexrd/ui/indexing/fit_grains_results_dialog.py
psavery: Sounds good, I'll do that soon.

Issue #685, opened 2020-12-17
Export cartesian view


saransh13:
Can we add an export option for the cartesian view? I've found it to be the most intuitive for people and its useful for generating figures for papers and abstracts. Two options I can think of are
1. Make it part of the polar plot export button and ask user if they want to write out the cartesian view to the same HDF file.
2. Make a separate button for just the cartesian view.

Thoughts? Personally I prefer 1 just to have everything in one file.

psavery:
Sure, we can add this!

I think it might be easiest to change the polar plot export button to a general "Export Plot" button that will export whichever view is currently displayed, Cartesian or Polar.

Internally, only one view (raw, cartesian, or polar) is generated and available at a time. It might require a little re-working to be able to export multiple views simultaneously (since we'd need to generate them first). And the Cartesian view is not always available, depending on the angles of the detectors, for instance.

But only exporting whichever view is currently displayed should be easy.

bnmajor:
When the polar view is saved, the `tth coordinates`, `eta coordinates`, and `extent` [are saved](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/calibration/polar_plot.py#L69) in addition to the image - is there any data for the cartesian view that should be included when saved?

bnmajor:
Closing for now, the option to export cartesian view has been added in #695. We can re-open if we'd like to expand on this to write out additional keys.

PR #695, opened 2021-01-05
Support exporting cartesian view

bnmajor:
Addresses #685

Still needs to add in any additional data that should be included with the cartesian image as mentioned [here](https://github.com/HEXRD/hexrdgui/issues/685#issuecomment-754682547)


psavery (approved):
I read through this, and tested it, and LGTM.

We can always add more keys in the future if we want more things to be saved.

bnmajor:
@psavery Sounds good, thank you!

Issue #652, opened 2020-10-27
Add menu to run fit grains directly


cjh1:

psavery:
@joelvbernier So the grain fitting currently uses the following output from the indexing:

1. Eta omega maps (to generate `min_samples, mean_rpg`).
2. Completeness 
3. `qfib`

When we run fit grains without indexing, where do we get these different pieces of info from?

I see that completeness is in the grains.out file. Does qfib come from a file we are currently not writing out in the GUI, like `accepted_orientations*`? I'll get the user to select all the files they need to.

joelvbernier:
A few things regarding grain fitting: this procedure (`hexrd fit-grains`) does **not** use the eta omega maps, completeness, or qfib, those are purely for the indexing step.  In general, what is required to fit a grain is are

- the orientation
- centroid
- strain
- imageseries.omega.OmegaImageSeries class
- PlaneData class

The CLI is set up to get this orientation from the `config` class, plus flags.  If the fitting has not been performed, or the `--clean` flag is issued, the `accepted_orientations_*.dat` file is used for the grain orientations along with trivial guesses for the centroids and strains (this is the default).  Alternatively, if a `fit_grains:estimate` exists, this information is obtained from the specified  `grains.out` file.

The process is two-step: first, `HEDMInstrument.pull_spots()` is executed to extract the _measured_ G-vector positions from the image stack (this is the data that goes into the `spot_XXXXXX.out` files),  Then a non-linear least-squares fit is performed between the simulated and measured positions.  The results are captured in the `grains.out` file.

There are additional options in the `fit_grains` block to control the data extraction/fitting process, namely specification of the angular tolerance windows to use in the imageseries around each simulated G-vector coordinate, and a "refit" tolerance `[m, n]`, which, if specified, performs the fitting again but excludes G-vectors whose measured and simulated coordinates differ by more than `m` pixels in `(x, y)` and `n` frames (_i.e._, an &omega; tolerance).  There is also a path for loading an overlap table, which is currently not part of the "standard" analysis, but I want to get it incorporated.

Does that help clarify things?

psavery:
[...]

PR #696, opened 2021-01-07
Allow fit grains to be ran directly

psavery:
This allows bypassing the indexing and running fit grains directly.

When the user runs fit grains directly, they must provide the grains
table in one of three ways. A select dialog will appear where the user
can choose from:

1. Indexing: use the grains table from the most recent indexing run
             (Only available if indexing has been performed)
2. Estimate: use the grains table from a grains.out file
3. Orientation: generate a grains table from an accepted_orientations.dat file

Once the user has selected one of these, the grain fit options dialog
will appear, and the user can select the grain fitting options. Afterwards,
grain fitting is ran and the results are viewed.

Internally, this re-organized the indexing runner into separate IndexingRunner
and FitGrainsRunner classes, so that the FitGrainsRunner can be used both
at the end of indexing, and also directly through the main window.

Some other cleanup has been performed, including making the tolerances
editable in the fit grains options dialog.

fit_grains_directly

Fixes: #652

### DIFF
--- a/hexrd/ui/indexing/fit_grains_options_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_options_dialog.py
@@ -1,5 +1,5 @@
 from PySide2.QtCore import (
-    QItemSelectionModel, QObject, QSignalBlocker, Qt, Signal, Slot)
+    QItemSelectionModel, QObject, QSignalBlocker, Qt, Signal)
 from PySide2.QtWidgets import QDialogButtonBox, QHeaderView
 
 from hexrd.ui import enter_key_filter
@@ -71,13 +71,11 @@ class FitGrainsOptionsDialog(QObject):
     def show(self):
         self.ui.show()
 
-    @Slot()
     def on_accepted(self):
         # Save the selected options on the config
         self.update_config()
         self.accepted.emit()
 
-    @Slot()
     def on_tolerances_add_row(self):
         new_row_num = self.tolerances_model.rowCount()
         self.tolerances_model.add_row()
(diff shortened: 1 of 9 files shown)


hexrd/ui/indexing/fit_grains_options_dialog.py
```diff
@@ -71,13 +71,11 @@ def all_widgets(self):
     def show(self):
         self.ui.show()
 
-    @Slot()
```
cjh1, line 74: Why remove these?

cjh1:
@psavery How do I run this, my "Fit Grains" menu item is disable?

hexrd/ui/indexing/fit_grains_options_dialog.py
psavery: They aren't necessary. If there is a benefit in having them, let me know.

psavery:
So, to run it, you'll need to have an imageseries loaded (such as the one we use for indexing), and the view need to be in raw mode.

hexrd/ui/indexing/fit_grains_options_dialog.py
cjh1: When you have a slot parameter, don't they help enforce the type?

psavery: Yeah, it looks like it. I guess we can use them if we want to enforce a type.

joelvbernier:
I'll have to try this out today.

cjh1 (approved):

Issue #681, opened 2020-12-08
Look at adding the "tight axes" when saving plots


cjh1:

psavery:
Right now, all of the code that is ran when you press the "Save" button is in matplotlib. So we will also need to come up with a way to override the matplotlib function. Maybe by overriding the function in our [child class](https://github.com/HEXRD/hexrdgui/blob/8c006bf33432c084cd5cf6de7b5d50927a590e9c/hexrd/ui/navigation_toolbar.py#L6).

PR #697, opened 2021-01-07
Save with a tight bbox by default in matplotlib

psavery:
To our custom navigation toolbar, add an argument to the initializer
that indicates whether figures should be saved with a tight bbox
(defaults to `True`).

This will override the rcParams at the time of saving, but they are
restored afterward.

Here are some examples:

Not tight
=======
not_tight

Tight
=====
tight

Fixes: #681


cjh1 (approved):
Nice, glad you managed to fine `rc_context`!

Issue #678, opened 2020-12-08
Add toggle option to use cylindrical reference frame


cjh1:

psavery:
@joelvbernier Can we get the math for this again?

joelvbernier:
The tvec_c components are in order (**X**, **Y**, **Z**);  we will report out coordinates (&rho;, &phi;, **Y**).

- In our case, **Y** is the cylinder axis and will remain unchanged in the transform.
- the radius &rho; is calculated as `np.sqrt(x*x + z*z)`
- the azimuth &phi; is calculated as `np.arctan2(z, x)`

PR #699, opened 2021-01-08
Add cylindrical reference display to fit grains

psavery:
This allows the user to choose a cylindrical reference frame instead
of the standard cartesian reference for the 3D plot.

In the cylindrical reference frame, the coordinates are (ρ, φ, Y),
where ρ = np.sqrt(x*x + z*z), and φ = np.arctan2(z, x).

cylind

Fixes: #678

### DIFF
--- a/hexrd/ui/indexing/fit_grains_results_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_results_dialog.py
@@ -25,6 +25,12 @@ from hexrd.ui.navigation_toolbar import NavigationToolbar
 from hexrd.ui.ui_loader import UiLoader
 
 
+COORDS_SLICE = slice(6, 9)
+ELASTIC_SLICE = slice(15, 21)
+ELASTIC_OFF_DIAGONAL_SLICE = slice(18, 21)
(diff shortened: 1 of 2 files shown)


psavery:
This will probably conflict with #698, but it should be easy to fix.

psavery:
@joelvbernier Should ρ and φ be displayed in degrees instead of radians?

hexrd/ui/indexing/fit_grains_results_dialog.py
```diff
@@ -72,7 +78,7 @@ def __init__(self, data, material=None, parent=None):
         eqv_strain = np.zeros(self.num_grains)
         hydrostatic_strain = np.zeros(self.num_grains)
         for i, grain in enumerate(self.data):
-            epsilon = vecMVToSymm(grain[15:21], scale=False)
+            epsilon = vecMVToSymm(grain[ELASTIC_SLICE], scale=False)
```
cjh1, line 81: Nice

cjh1 (approved):
LGTM, thanks for pulling out the constants

PR #700, opened 2021-01-08
Fix fit grains plot coordinates

psavery:
It was plotting:
```python
(grain1[0], grain2[0], grain3[0])
(grain1[1], grain2[1], grain3[1])
(grain1[2], grain2[2], grain3[2])
```

When it should have been plotting:
```python
(grain1[0], grain1[1], grain1[2])
(grain2[0], grain2[1], grain2[2])
(grain3[0], grain3[1], grain3[2])
```

This bug was introduced in 87545810f98dce86678aa83c268f76a3bbb6bec8

The transpose fixes it. You can see the change with the data below:

Screenshot from 2021-01-08 11-31-05

Before
=====
bad

After
====
good


cjh1 (approved):
LGTM

Issue #680, opened 2020-12-08
Allow user to select a point on the 3D plot and highlight the grain in the table


cjh1:
We will need to check if matplotlib will let use do this.

psavery:
To leave some notes here:

I am able to do most of this by first adding as a kwarg to `scatter3D()` `'picker'=True`, then
connecting to pick events via `canvas.mpl_connect('pick_event', self.point_picked)`.

When I get the pick event, there is an index associated with it, which appears like it is
supposed to match the index of the original data used to create the plot. However,
the index does not match for some reason. This could be a bug in matplotlib, I'm
not sure.

Issue #679, opened 2020-12-08
Allow user to click on row in grain table to highlight grain in 3D plot


cjh1:

psavery:
We might also want a way to show/hide grians.

joelvbernier:
I think we'll want to be able to show/hide by applying sort/filter on values in the grain results table as well as direct user selection.

PR #701, opened 2021-01-08
Highlight grain row when point is picked

psavery:
This almost works. A row is highlighted when a point is picked,
and this is based upon the `event.ind` that gets passed to the
event callback when a point is picked. In theory, this index
should match up with the index of the data that was used to
generate the point (see [here](https://matplotlib.org/3.3.3/gallery/event_handling/pick_event_demo.html) for an example). However,
in our case, the index does not seem to match the data. I'm
not sure why. It could be a bug on matplotlib's side.

Anyways, if we can fix that, this will be mostly ready to go.
The one other thing we should check is what happens if multiple
points are selected when the user clicks (which one gets chosen?).

Fixes: #679
Fixes: #680


joelvbernier:
Ooops, looks like some conflicts.

PR #708, opened 2021-01-13
Update constants.py

joelvbernier:
Update default colormap for the benefit of our colorblind friends...


cjh1 (approved):

cjh1:
@joelvbernier Was there a particular problem with 'plasma', my understand that modern matplotlib colormap, such as plasma were designed to be colorblind friendly?

psavery:
Yeah, it might be nice to have a default colormap that is both colored and looks good in grayscale (for colorblind users).

joelvbernier:
@cjh1 -- I think the issue mainly comes when drawing overlays on non-grayscale colormaps.  I was working with a grad student of Darren's and he was having a hard time seeing everything.  Can't go wrong with grayscale as default :-)

Issue #711, opened 2021-01-20
Site editor in structure tab not accepting coord modifications


joelvbernier:
Changes in fractional coordinates are not accepted.  I loaded a material archive and attempted to modify atom coordinates. They get overwritten by the as-loaded values when you toggle back and forth between another site.   See screen cap below.  
![Screen Recording 2021-01-19 at 08 31 41 83 PM](https://user-images.githubusercontent.com/1154130/105127402-b21ffd80-5a95-11eb-8183-437822b9cbaf.gif)

joelvbernier:
Edit: same behavior if you try to make a new material; you are stuck with the coords of the most recent active material and cannot change them.  Also, there is some very strange mouse behavior going on; notice I _am not_ holding Ctrl or Shift, and left mouse is selecting multiple sites.
![Screen Recording 2021-01-19 at 08 29 49 52 PM](https://user-images.githubusercontent.com/1154130/105128117-3aeb6900-5a97-11eb-9283-086734aa68e2.gif)

psavery:
@joelvbernier I think this might be happening because of #513. Out of the default materials, ruby and forsterite are the two that have a "U" that is a symmetric tensor rather than a single value, and we hadn't yet built in a way to deal with a U that has a symmetric tensor. If you make a new material, it copies over the old material, so it may also have a symmetric tensor U.

Can you give this a try with any of the other materials, other than ruby and forsterite, to see if modifying the fractional coordinates works?

joelvbernier:
@psavery: I played with Si (which uses isotropic B) and get the same weirdness.  Not sure why the mouse left click is behaving like Ctrl+click if I add sites and try to switch between them.  I also, edits to any site (if I add more than the one) edit all the sites.

@saransh13, can you help Patrick with this?  Neither the 0.8 nor the current master GUI has a functional structures tab.

saransh13:
@psavery let me know how I can help. I can't reproduce this on OSX.

joelvbernier:
Confirmed the same behavior on Linux, using the ruby as an example.

joelvbernier:
Am I wrong in stating that our current interface has a fundamental problem w.r.t. inputting U tensor values?  We only have a scalar spin box...

psavery:
Yeah, it looks like trying to view a site that had a tensor U put the structure editing tab in a bad state, and caused a lot of bugs. #712 prevents us from getting into this bad state. However, we should still add a tensor editor soon (tracked in #513).

Issue #513, opened 2020-09-11
Allow "U" to be specified as a symmetric tensor in the site editor


psavery:
`U` can either be a single value, or a 3x3 symmetric tensor. We should allow a symmetric tensor to be specified as well.

We can perhaps add a checkbox as to whether the `U` is a single value or a tensor. And, if it is a tensor, instead of a single spinbox in the `U` column, we can put an `Edit` button that brings up a small dialog with the matrix to be edited.

![Screenshot from 2020-09-11 16-36-06](https://user-images.githubusercontent.com/9558430/92970590-ec74f080-f44c-11ea-982b-1a10ea65fb7c.png)

If all `U` values need to match in whether they are tensors or single values, we can have one checkbox for all of them. Otherwise, we can maybe add a column with a checkbox for each `U`. It depends on whether each `U` can be a tensor or single value.

Currently, one of the default materials, `PbSO4`, has symmetric tensors for `U`, and it produces an error if you try to edit a site in the structure tab:

```python
Traceback (most recent call last):
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/material_structure_editor.py", line 62, in selection_changed
    self.update_site_editor()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/material_structure_editor.py", line 83, in update_site_editor
    self.material_site_editor = MaterialSiteEditor(site, self.ui)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/material_site_editor.py", line 46, in __init__
    self.update_gui()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/material_site_editor.py", line 197, in update_gui
    self.update_table()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/material_site_editor.py", line 212, in update_table
    w = self.create_thermal_factor_spinbox(self.thermal_factor(atom))
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/material_site_editor.py", line 175, in create_thermal_factor_spinbox
    sb.setValue(v)
TypeError: only size-1 arrays can be converted to Python scalars
```

PR #712, opened 2021-01-20
Material structure editor fixes

psavery:
Add temporary fix to ensure all U's are scalars

This helps us avoid putting the material structure editor in a bad state
for now. It performs a mean() on all U values that are tensors.
    
We should add a tensor editor to handle the tensor U in the future.

Use scientific spinboxes for site editor
    
These allow an arbitrary number of decimal places.

Fixes: #711

Alleviates #513, but we still need to add a tensor editor for U.


cjh1 (approved):
LGTM, thank you!

PR #717, opened 2021-01-25
Use scientific spin boxes in form view

### LINKED ISSUE #716: Pixel sizes in Form View not updating properly upon change in Tree View
I was making a detector from scratch in the GUI (single panel, but that's irrelevant) and noticed that there is a disconnect between the pixel size reported in the Form View and Tree View; it seems only the latter allows editing, and the former does not get updated after making changes in the Tree View:

image

Observe:
image

psavery:
This PR fixes the scientific spin box so that it ignores the prefix/suffix (if they are present), and it also changes all of the widgets in the form view to use the scientific spin box. This increases the precision for displaying and entering parameters.

This also fixes #716, which was being caused by the spinboxes now allowing values below 0.1.

Fixes: #716

### DIFF
--- a/hexrd/ui/resources/ui/calibration_config_widget.ui
+++ b/hexrd/ui/resources/ui/calibration_config_widget.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>506</width>
+    <width>560</width>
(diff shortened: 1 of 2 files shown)


cjh1 (approved):
LGTM

PR #718, opened 2021-01-25
Omega precision

### LINKED ISSUE #714: Precision limitation leading to rounding error in omega specs in image loader
I just noticed this one:

I have an imageseries set from chess with 1441 frames spanning (0, 360).  Thus the delta omega reflected in the metadata on the imageseries object is 360./1441 ~ 0.249826509.  However, when I load these imageseries series via the HEDM-style "Load Data" widget, I see the following:

image

The precision limit of 2 decimal places on the delta omega is leading to an erroneous omega max value of 360.25; now, I am not entirely sure how these numbers are getting used, since the imageseries already has metadata attached (@psavery or @bnmajor could you tell me how/where these values get used?  Obviously, if you are loading raw images, this is the only way to input the omega metadata, so I imagine these values are indeed important)

This bad omega information would cause find-orientations and fit-grains to fail, the range *cannot* exceed 360.  In this case, I cannot enter the accurate delta omega manually, and the omega max seems to be slaved to the delta omega value.  We will at the very least require a higher degree of precision in these fields to avoid this rounding error (single precision should suffice).  We should also raise a warning if the range exceeds 360.

bnmajor:
- Set precision in table to 6 decimal places
- Warn users if omega range is greater than 360°

Fixes #714

### DIFF
--- a/hexrd/ui/load_panel.py
+++ b/hexrd/ui/load_panel.py
@@ -477,9 +477,18 @@ class LoadPanel(QObject):
 
         self.ui.file_options.blockSignals(False)
 
-    # Process files
+    def confirm_omega_range(self):
+        omega_range = abs(self.omega_max[0] - self.omega_min[0])
+        if not (r := omega_range <= 360):
+            msg = f'The omega range is greater than 360°.'
+            QMessageBox.warning(self.ui, 'HEXRD', msg)
+        return r
 
+    # Process files
     def read_data(self):
+        if not self.confirm_omega_range():
+            return
+
         data = {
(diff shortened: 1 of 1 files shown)


bnmajor:
The PEP 8 issues are because the walrus operator is being used on [line 483](https://github.com/HEXRD/hexrdgui/blob/31fc313e77bdfba0885b1fc8335fbbc92e91d114/hexrd/ui/load_panel.py#L483)

cjh1 (approved):
LGTM

PR #719, opened 2021-01-25
Load panel state

### LINKED ISSUE #713: Load Data options reverting to defaults upon reloading imageseries
Noticed this bug working with Darren today: if you load and imageseries with a set of processing options (aggregation, transform, dark), then load a new imageseries (for the same instrument), the options are reverting to the default values but the selectors are not changing.  For example, if I loaded an imageseries with max aggregation, loading a new file set will load without aggregation even though the selector still reads max aggregation.  Same with the dark; if I've previously set it to None, then it performs Median despite the fact the selector still shows None. The only way to get it to perform the options the selectors indicate is to toggle the "apply to all detectors" checkbox and re-select the desired options.

The desired behavior is for the current selections to be "sticky" so that you can load different datasets for the same instrument without having to reset all the selectors.

bnmajor:
Fixes #713 as well as a few other issues that were discussed:

- Change load panel defaults: `Aggregation` defaults to `Maximum`, `Image Transform` continues to default to `None`, `Dark Mode` defaults to `None`, and `Apply Selections to All Detectors` defaults to being checked.
- The progress dialog text is not updated to say that the dark images are being aggregated unless they actually are. Previously, the text was being updated to soon and it gave the false impression that dark aggregation was being applied even when it was not.
- Make sure that the selected files are cleared and the table is updated to reflect this when a new instrument configuration is loaded.


cjh1 (approved):
LGTM

PR #725, opened 2021-01-28
Make sure delta is float before rounding

bnmajor:


bnmajor:
@joelvbernier I've added a line that ensures that the delta value is always a float before attempting to round it for display in the table. I was not able to actually reproduce the issue myself though, would you be able to share which instrument/imageseries or steps that you took when you saw this?

bnmajor:
Just to note - this issue was happening when files without any omega data were being loaded because (in that case and that case only) defaults were being created as strings rather than float values.

joelvbernier (approved):

Issue #654, opened 2020-10-30
Consider using "draw_idle()" in parts of the code


psavery:
The matplotlib `FigureCanvas.draw()` function is currently being used in parts of the code to re-render the canvas. This is good, but we might sometimes call `draw()` two or more times before control is returned to the event loop (and only the last draw will count).

If we use `draw_idle()`, an event to draw is posted to the event loop, and repeated calls to `draw_idle()` only result in one `draw()` being called.

This might improve efficiency in some cases, especially where drawing is slow (which is the case if there are a lot of overlays, for instance, or the resolution is too high). It can also allow us to not have to worry about calling `draw()` too many times.

PR #727, opened 2021-01-29
Prevent unnecessary re-rendering of the window

bnmajor:
Repeated calls to the `update_all` function was causing the canvas to be redrawn multiple times during the process of loading images. Prevent these unnecessary calls by temporarily turning off live updates until the image processing has completed.

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -37,6 +37,7 @@ class ImageLoadManager(QObject, metaclass=Singleton):
     update_needed = Signal()
     new_images_loaded = Signal()
     images_transformed = Signal()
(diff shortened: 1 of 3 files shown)


bnmajor:
Testing data out locally showed a pretty good speed up, from approximately 25 seconds to approximately 4 seconds to load.

Before:

slower_ge_detectors


After:

faster_ge_detectors

bnmajor:
These changes seem to have also improved the way the progress bar updates during aggregation:

Before:

agg-progress-old

After:

agg-progress-new

cjh1 (approved):
LGTM, a great improvement!

psavery:
@bnmajor Awesome! Looks great.

A related issue: #654

Issue #721, opened 2021-01-25
Spot labeling in eta omega maps viewer should change with the method


psavery:
Currently, we are [always labeling spots as if the method were "label"](https://github.com/HEXRD/hexrdgui/blob/beadbc312f91c1633c51fe16c260fb1084e4ac2f/hexrd/ui/indexing/ome_maps_viewer_dialog.py#L280-L304).

We need to change the spot labeling to be based upon the chosen method, [as is done in hexrd](https://github.com/HEXRD/hexrd/blob/b79c7fa34e4f9aadb81da4c8425bd2e3f8de887a/hexrd/findorientations.py#L112-L152).

PR #728, opened 2021-01-29
Update spots with chosen method in ome maps viewer

psavery:
Previously, it always used logic for the "label" method when creating
the spots, and it wasn't even fully accurate for the "label" method (it
was missing some of the code logic).

Instead, we should call the "label_spots" function in hexrd so the spots
are generated the exact same way that hexrd does, to give the user an
accurate preview.

Additionally, it now generates spots based upon the method that has
been selected (label, blob dog, or blob log), and the options for each of
those methods. The spots are automatically updated with any change to
the method options or to the method itself.

Depends on: hexrd/hexrd#163
Fixes: #721


psavery:
This is ready when hexrd/hexrd#163 gets merged.

joelvbernier (approved):

PR #730, opened 2021-01-29
Load panel bugs

bnmajor:
Fixes some bugs I found in the load panel:

- The directory label should only display the parent directory. It was being updated to display a full file path on occasion; fix this.
- Connect the clear load panel state signal to the load panel so that the current selections are updated accordingly (this happens if a new instrument config is loaded)
- Do not reset the table unless a new instrument config is loaded (the table was previously being cleared every time a transform was applied and the images were loaded)


joelvbernier (approved):

joelvbernier:
Uh oh @bnmajor -- I think the YAML PR might have caused this conflict?  Can you fix the merge?

PR #732, opened 2021-02-03
Multiple tiffs

bnmajor:
This adds a dialog to the load panel that can be accessed through the new `Select Image Stacks` option.

load_panel

This dialog essentially replaces the `YAML` file used for the `image-files` `Imageseries` extension. The `YAML` file format:

```
image-files:
  directory: GE
  files: "ti7_*.ge2"
options:
  empty-frames: 0
  max-frames: 2
meta:
  omega: "! load-numpy-array example-omegas.npy"
```
image_stack_dialog

The directory and search pattern must be set manually on each detector, which is done by selecting the detector from the drop-down menu. If no search pattern is entered all of the files in the directory will be selected.

The rest of the data is assumed to be the same for each detector and is only set once. This can be changed to be set per detector id needed.

If the `Load from File` option is checked, a `npy` file can be selected to be used as the omega data.

When a search pattern is entered the first matching file is opened as an `Imageseries` in order to determine the number of frames per file. This value is used to calculate the delta and is assumed to be the same for every file. It is also used to restrict the ranges for `Max File Frames`, `Max Total Frames`, and `Delta`.

The state is remembered between uses and is only reset if the detectors change.


bnmajor:
The remaining PEP8 issues on this branch are the result of the walrus operators [here](https://github.com/HEXRD/hexrdgui/blob/d97ba572354c17a154ee6d9a47d5ed0ff51a441d/hexrd/ui/image_stack_dialog.py#L99) and [here](https://github.com/HEXRD/hexrdgui/blob/d97ba572354c17a154ee6d9a47d5ed0ff51a441d/hexrd/ui/load_panel.py#L515)

joelvbernier:
@bnmajor -- will the file selector support generic glob templates?  We often will have directories with many scans in them, all with the same stem and different file numbers.  It would also be very useful if we could also propagate the selection across all detectors.  I'll have to think about how we want to specify the filename templates, but `glob` is probably the way to go.

joelvbernier:
@bnmajor -- perhaps we can have 2 options for the file specification:
1. just have the ability to open a file selector (so we can use Ctrl and Shift + click; most useful for single detector instruments)
2. have the templated version that can be propagated across all detectors
In both cases, we will want to echo back to the user the number of files selected per detector as a sanity check.  The omegas will raise an exception if we try to attach them to an imageseries with an inconsistent number of frames.

bnmajor:
@joelvbernier Sounds good, I'll update the options. In the case of the templated search will that always apply to all detectors? Or do we need to keep the option to set it per detector as well?

joelvbernier:
good question... I'd say it would almost always apply to all, but there can be cases where the detector types are different they write different numbers and formats of files...

joelvbernier:
One request:  Allow for the definition of the omegas using a list of "wedges".  We do have instances where the ranges my not be contiguous across the full frame list.  Luckily `imageseries.omega` has a class for generating the omega metadata that we should be using.  See example below
```
nsteps = 240
totalframes = 4*nsteps
omwedges = omega.OmegaWedges(totalframes)
omwedges.addwedge(0, 60, nsteps)
omwedges.addwedge(120, 180, nsteps)
omwedges.addwedge(180, 240, nsteps)
omwedges.addwedge(300, 360, nsteps)

ims.metadata['omega'] = omwedges.omegas
```
Since we will know the total number of frames from the user input in the form (or can get it), we won't have to calculate that.  We can have the user add wedges via (start ω, stop ω, steps) interactors.  The metadata object on the imageseries get set as `ims.metadata['omega'] = omwedges.omegas`

joelvbernier:
I'll say that this also circumvents the need to input delta omegas to a zillion sig figs ;-)

bnmajor:
@joelvbernier This has been updated with the changes we discussed. This PR also fixes #760 in order to keep the load panel in sync with the behavior of the dialog.

joelvbernier:
Ok -- I like the general layout, but I have 2 requests:

1. the widget doesn't let me cancel; just keeps re-raising if I try "cancel", killing via the frame's close button, or typing "ok" and click through the error dialog.
2. we may want to allow the user to optionally specify all detectors with a format string exactly like you did in #773

bnmajor:
@joelvbernier I've fixed the bug that was preventing canceling the dialog and have added the option to select directories for each detector by directory stem and follows the same logic as the save dialog.

Right now, if the stem:

- Is a directory: the detector name will just be appended to the path
- Is not a directory: the detector name will be appended with an underscore

So given the following file structure:
```
home
  user
    images
      ge1
        001.tiff
        ...
        00N.tiff
      detector_ge1
        001.tiff
        ...
        00N.tiff
```
The stem `/home/user/images` will search for `/home/user/images/ge1` and the stem `/home/user/images/detector` will search for `/home/user/images/detector_ge1`. Let me know if this seems correct.

joelvbernier:
Just tested and it looks like it is working now.  There is one spot I called out in the review where we should make a quick modification to avoid a deprecation warning from `yaml.load()`, but we can fix that later.

joelvbernier (approved):

joelvbernier:
Ah -- I did find a bug after the fact, so don't delete your feature branch quite yet...  

If I try to load a single image (through the widget) after successfully loading a big series of files, it looks like the omega wedges are not getting reset, which leads to the following
```
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\image_load_manager.py", line 199, in process_ims
    self.add_omega_metadata(HexrdConfig().imageseries_dict)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\image_load_manager.py", line 343, in add_omega_metadata
    ims_dict[key].metadata['omega'] = omw.omegas
  File "C:\Users\berni\miniconda3\envs\hexrdgui\lib\site-packages\hexrd\imageseries\omega.py", line 145, in omegas
    raise OmegaSeriesError(msg)
hexrd.imageseries.omega.OmegaSeriesError: 'number of frames (1) does not match number of wedge frames (240)'
```

cjh1:
This has been fixed in #775

Issue #735, opened 2021-02-03
Need interface to specify map indices for seeded search 


joelvbernier:
In the HEDM config YAML, the indices under `find_orientation:seed_search_hkl_seeds` allows specification of a subset of maps to use for the seeded search.  We typically only use the first 2 or 3 maps, not all.  There is currently no attribute for setting this list in the GUI -- is it running `generate_orientation_fibers` on all the maps?

psavery:
@joelvbernier Yes it is currently using all the maps. I will add an option to specify this.

Issue #646, opened 2020-10-20
Ensure hkls of loaded eta omega maps match that in the GUI


psavery:
Currently, for the indexing, the user can load eta omega map files that do not have the same HKLs (or maybe not even the same material) as that in the GUI.

With this mismatch, the indexing runs fine, but a key error will occur when the user tries to run grain fitting, that looks like this:
```python
Traceback (most recent call last):
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/indexing/run.py", line 171, in run_fit_grains
    min_samples, mean_rpg = create_clustering_parameters(config,
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/findorientations.py", line 581, in create_clustering_parameters
    seed_hkl_ids = [
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/findorientations.py", line 582, in <listcomp>
    plane_data.hklDataList[active_hkls[i]]['hklID']
IndexError: list index out of range
```

It might be nice to ensure the HKLs of the loaded eta omega maps match, because this is otherwise a little bit of an obscure error for the user, and they may not realize what the problem is.

psavery:
This is probably less important as of #832, because the user explicitly selects the material they are using. I guess it will be their responsibility to make sure the material used to generate the eta omega maps and the material they are using for indexing/grain fitting are the same.

PR #742, opened 2021-02-04
Add ability to select hkls in ome maps viewer

psavery:
Previously, we were running the indexing with all hkls. This
allows the users to select the hkls they wish to use.

example

Fixes: #735 

@joelvbernier What do you think of this?


psavery:
@joelvbernier Please also note that, for loading eta omega maps from a file, there is currently no validation that ensures that the material and active hkls (ones selected in the materials table) match those in the GUI. This is being tracked in #646. A mismatch can cause an exception, or worse, results that are invalid.

Do you think we should add some validation for this kind of mistake?

joelvbernier (approved):

PR #743, opened 2021-02-04
typo in distortion application

joelvbernier:
I noticed for a user's GE detector example that the overlays in the raw and cartesian mode were way off after calibration, while the polar view showed great alignment.  I found a simple typo where the distortion function was being applied to the wrong variable in the powder diffraction overlays.  Unfortunately, the Cartesian view mode will require a more in-depth fix.


psavery (approved):
LGTM

PR #745, opened 2021-02-09
Always remove previous hexrdgui in wix

psavery:
This is an attempt to always remove previous hexrdgui installations via
wix. The previous behavior is that old files are not removed but only
over-written with new files.


psavery:
From my testing, it looks to me that previously installed files are already removed by wix before it installs the new ones. So changes may not be needed.

If you manually create a file/folder in the directory, it won't remove it when it removes the old version and creates the new one. But it does look like it removes previously installed files, so we are hopefully good...

PR #749, opened 2021-02-11
cast warped images as masked array

joelvbernier:
running fix; this one worked, but the sum was hosed on account on the nans will continue working it.

### DIFF
--- a/hexrd/ui/calibration/polarview.py
+++ b/hexrd/ui/calibration/polarview.py
@@ -215,9 +215,13 @@ class PolarView:
                 beamVec=panel.bvec)
         xypts[on_plane] = valid_xys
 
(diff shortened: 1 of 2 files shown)


joelvbernier:
Tested.  Looks good; LGTM

psavery (approved):
LGTM - although we need to make the 'Show SNIP' behavior when 'Apply SNIP?' is unchecked match the behavior of applying the SNIP. This will be done separately.

psavery:
> LGTM - although we need to make the 'Show SNIP' behavior when 'Apply SNIP?' is unchecked match the behavior of applying the SNIP. This will be done separately.

This is now being tracked in #750

PR #754, opened 2021-02-17
modified autoscaling behavior

joelvbernier:
simple change to autoscaling behavior.  Looks to give the desired behavior.


psavery (approved):
LGTM

Issue #752, opened 2021-02-17
Warning when mouse first moves over masked intensity


psavery:
This warning gets produced the first time a user moves their mouse over a masked intensity:
```python
/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/main_window.py:888: FutureWarning: Format strings passed to MaskedConstant are ignored, but in future may error or produce different behavior
  labels.append('value = {:8.3f}'.format(info['intensity']))
```

It looks like it is caused by trying to format a masked value. For instance, I get the same warning
if I do this:
```python
>>> import numpy as np
>>> array = np.array([1, 2, 3, 4], dtype=np.float32)
>>> masked = np.ma.masked_array(array, mask=True)
>>> f'{masked[0]:0.2f}'
<stdin>:1: FutureWarning: Format strings passed to MaskedConstant are ignored, but in future may error or produce different behavior
'--'
```

joelvbernier:
Yeah, I saw that.  It looks like it is getting ‘—‘.  I believe we can actually use the fix_invalid method to have it report the fill_value instead

psavery:
Okay. It looks like this is a masked array created by matplotlib. Our NaNs are coming out masked because matplotlib is using `np.ma.masked_invalid()` to set the data. They have the fill value set at the default (`1e+20` for floats).

Currently, for these masked values, the intensity in the status bar is showing `--` (which is what comes out if you `print()` a masked value).

For our NaNs in the image, what do we want them displayed as in the status bar? One of these maybe?

1. `--`
2. `NaN`
3. 0

I think we can also just check if the intensity value is `np.ma.core.MaskedConstant`, and if it is, display one of the 3 above instead of trying to format it. That would avoid the formatting warning.

Issue #641, opened 2020-10-08
TypeError on mouse hover over raw image canvas


joelvbernier:
Notice this being spit to stdout
```
File "c:\users\berni\onedrive\documents\github\hexrdgui\hexrd\ui\main_window.py", line 832, in new_mouse_position
    labels.append('value = {:8.3f}'.format(info['intensity']))
TypeError: unsupported format string passed to MaskedArray.__format__
```

bnmajor:
@joelvbernier I was unable to reproduce - Is this still an issue you are seeing? If so can you share your workflow leading up to the error?

psavery:
It looks to me that this is the same as #752, except maybe it is an error on Windows instead of a warning (like it is on Linux).

It looks like the problem occurs when you move the mouse over a masked value in the image. Matplotlib creates masked values from `nan` and `inf` values. So if there are any `nan` or `inf` values in the raw view, this error would occur when the mouse moves over that spot.

PR #755, opened 2021-02-18
Avoid trying to format MaskedArray instances

psavery:
Matplotlib is creating a MaskedArray from the data, where nans and infs
are masked. We tried to format the intensity with a float formatting string, and
this was producing a warning on Linux and an error on Windows.

Check to see if the intensity is an instance of np.ma.MaskedArray
(as a side note: np.ma.core.MaskedConstant inherits from np.ma.MaskedArray,
so this check also works for np.ma.core.MaskedConstant). If it is,
then show another label and don't try to format it.

Fixes: #641 
Fixes: #752

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -878,17 +878,22 @@ class MainWindow(QObject):
 
     def new_mouse_position(self, info):
         labels = []
-        labels.append('x = {:8.3f}'.format(info['x_data']))
-        labels.append('y = {:8.3f}'.format(info['y_data']))
+        labels.append(f'x = {info["x_data"]:8.3f}')
+        labels.append(f'y = {info["y_data"]:8.3f}')
         delimiter = ',  '
 
         intensity = info['intensity']
         if intensity is not None:
-            labels.append('value = {:8.3f}'.format(info['intensity']))
-            labels.append('tth = {:8.3f}'.format(info['tth']))
-            labels.append('eta = {:8.3f}'.format(info['eta']))
-            labels.append('dsp = {:8.3f}'.format(info['dsp']))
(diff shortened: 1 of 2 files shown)


hexrd/ui/main_window.py
```diff
@@ -878,17 +878,22 @@ def view_fit_grains_config(self):
...
 
         intensity = info['intensity']
         if intensity is not None:
-            labels.append('value = {:8.3f}'.format(info['intensity']))
-            labels.append('tth = {:8.3f}'.format(info['tth']))
-            labels.append('eta = {:8.3f}'.format(info['eta']))
-            labels.append('dsp = {:8.3f}'.format(info['dsp']))
-            labels.append('hkl = ' + info['hkl'])
+            if isinstance(intensity, np.ma.MaskedArray):
+                masked_label = '--'
```
psavery, line 888: @joelvbernier If you'd rather this be something other than `--`, let me know. Matplotlib masks both `nan` and `inf`, so it could be either.

If it's preferable, we could also try to grab the unmasked value [here](https://github.com/HEXRD/hexrdgui/blob/36831d1fc8360ee26dae12b6cd8199b0356817f2/hexrd/ui/image_tab_widget.py#L266), and then display what it actually is, either `nan` or `inf`.

joelvbernier: `nan` is probably the best.

joelvbernier: LGTM

psavery: Okay, it's fixed. It now actually displays whatever the underlying data is: `nan` or `inf`.

joelvbernier (approved):

PR #756, opened 2021-02-19
Update setup.py

joelvbernier:
edit specs for fabio package.


cjh1 (approved):
LGTM, installs fabio 0.11 as expected ...

PR #762, opened 2021-02-23
Use instrinsic zxz as euler angle convention for LLNL import

### LINKED ISSUE #759: Force Euler angle convention to intrinsic zxz for LLNL workflow.
I noticed while switching workflows in the GUI that the Load Data widget in the LLNL workflow was not computing the tilts properly -- the cause seemed to be the fact that the Euler Angle conventions had been set to "None", but the widget is expecting it to be "Intrinsic ZXZ".  As such, we should force this selection upon switching the workflow to LLNL (or at least when using the Load Data widget in that setting).

bnmajor:
Fixes #759 

@joelvbernier It looks like the Euler angle was only being corrected in the case of TARDIS. It should now be updated to "Intrinsic ZXZ" for all instruments (this is updated any time you start using the Import Data Panel).

### DIFF
--- a/hexrd/ui/import_data_panel.py
+++ b/hexrd/ui/import_data_panel.py
@@ -103,9 +103,6 @@ class ImportDataPanel(QObject):
                                 enabled=True)
 
     def set_convention(self):
-        if self.instrument != 'TARDIS':
-            return
-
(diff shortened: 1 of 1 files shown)


psavery (approved):
LGTM

PR #764, opened 2021-02-23
Update README.md

joelvbernier:
Fixes: #765

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # Installing
 
 ## conda (release)
+Requires Python 3.8+
+
 To install the latest stable release
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -51,8 +51,8 @@ conda install -c anaconda python=3.8
 # Install deps using conda package
 conda install -c HEXRD -c anaconda -c conda-forge hexrdgui
 # Now using pip to link repo's into environment for development
-pip install --no-deps -U -e hexrd
-CONDA_BUILD=1 pip install --no-deps -U -e hexrdgui
+CONDA_BUILD=1 pip install --no-build-isolation --no-deps -U -e hexrd
+CONDA_BUILD=1 pip install --no-build-isolation --no-deps -U -e hexrdgui
 ```
 
 ### Mac OS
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,32 @@
-# Develop
+# Installing
 
-Requires Python 3.8+
+## conda (release)
+To install the latest stable release
+
+```bash
+conda install -c hexrd -c anaconda -c conda-forge hexrdgui
```
psavery, line 7: We might want to clarify somewhere around here that python 3.8+ is required.

psavery:
@joelvbernier I pushed a commit here that fixes the fabio error, and I also updated the instructions to add the `--no-build-isolation` flag which fixes an issue I encountered as part of #765.

With these changes, and with hexrd/hexrd#183, I am able to follow the instructions exactly and get a conda development
environment set up on macos.

README.md
cjh1: Yes, lets not remove the fact that we need python 3.8, unless we want to support multiple versions ...

psavery: I just fixed that.

psavery (approved):
LGTM

setup.py
```diff
@@ -16,8 +15,10 @@
 # dist info so setuptools can't find it, even though its there, which results in
 # pkg_resources.DistributionNotFound, even though the package is available. So we
 # only added it if we aren't building with conda.
+# fabio from the hexrd channel has a similar problem.
```
cjh1, line 18: I have removed fabio from our channel, so I think this comment can me removed.

psavery: I'll give it a try

psavery: Yep, I just gave it a try, and this change is no longer needed. I'll revert this commit.

psavery: @cjh1 That commit was removed, and now we are back to README only changes.

cjh1 (approved):

joelvbernier:
Just to be clear, is Python 3.8 ok on Mac OS now?  Or do we still need our special python 3.7.6 for Qt?

cjh1:
Our special python is 3.8. I haven't check recently but last time I looked we still needed it as the changes hadn't been back ported yet, I will take a look.

psavery:
Yeah, I think once we start requiring python 3.9, we might not need the special python.

cjh1:
OK, so I download 3.8.8 from conda-forge and it is indeed built with 10.15, so I think we are now safe to install from conda-forge, yay!

cjh1:
But we need to update the readme to say the on the mac you need to install python from conda-forge, I will push a PR.

joelvbernier:
Ah bollocks @cjh1 -- looks like the `conda-forge` python 3.8.8 is not working on Mac OS 11 (Big Sur); our old `HEXRD` python 3.8.4 is, however. Probabaly should fix the README (again) ASAP.

cjh1:
@joelvbernier in what way does it not work? I can update the readme .

cjh1:
@joelvbernier Is it just failing to start?

PR #766, opened 2021-02-24
WPPF update

saransh13:
Theres a few changes to `wppf_options_dialog`. Listing them below.

1. The files for the default `LeBail` and `Rietveld` parameters are now two separate file (name of files should be self explanatory for which corresponds to which).
2. Added a function named `add_materialparameters` [here](https://github.com/HEXRD/hexrdgui/blob/d9d8d73359997479046280932c2ead48fa2aded1/hexrd/ui/calibration/wppf_options_dialog.py#L424) which can add all the parameters which are not present in file. This function can now replace the existing method now.
3. The wavelength is now passed to `LeBail` and `Rietveld` class identically. This was slightly different. The format for `LeBail` was `wavelength = {name : valWUnit type}`, which has now been changed to `wavelength = {name : [valWUnit type, float]}`. The additional `float` is a weight factor. This required a small change to the `wppf_options_dialog' [here](https://github.com/HEXRD/hexrdgui/blob/d9d8d73359997479046280932c2ead48fa2aded1/hexrd/ui/calibration/wppf_options_dialog.py#L571)


hexrd/ui/calibration/wppf_options_dialog.py
```diff
@@ -420,6 +421,136 @@ def create_wppf_params_object(self):
...
+                        for j in range(6):
+                            nn = p+'_'+elem + \
+                                str(atom_label[i])+'_'+_nameU[j]
+                            params.add(
+                                nn, value=U[i, j],
+                                lb=-1e-3,
+                                ub=np.inf,
+                                vary=False)
+                    else:
+                        nn = p+'_'+elem+str(atom_label[i])+'_dw'
```
psavery, line 544: Note that in python >= 3.6, you can use f-strings like this:
```python
nn = f'{p}_{elem}{atom_label[i]}_dw'
```

and it might be a little simpler.

hexrd/ui/calibration/wppf_options_dialog.py
```diff
@@ -420,6 +421,136 @@ def create_wppf_params_object(self):
...
+
+                    """
+                    first 3 are lengths, next three are angles
+                    """
+                    if(rid[i] <= 2):
+                        params.add(nn, value=l, lb=l-0.05,
+                                   ub=l+0.05, vary=False)
+                    else:
+                        params.add(nn, value=l, lb=l-1.,
+                                   ub=l+1., vary=False)
```
psavery, line 483: The above section of code looks nearly identical to the section in `if method == 'LeBail':`. It might be better to extract this out of the `if` blocks to avoid code duplication.

hexrd/ui/calibration/wppf_options_dialog.py
```diff
@@ -420,6 +421,136 @@ def create_wppf_params_object(self):
...
+                    nn = p+'_'+n
+
+                    """
+                    first 3 are lengths, next three are angles
+                    """
+                    if(rid[i] <= 2):
+                        params.add(nn, value=l, lb=l-0.05,
+                                   ub=l+0.05, vary=False)
+                    else:
+                        params.add(nn, value=l, lb=l-1.,
```
psavery, line 482: Is `params` defined somewhere, or should it be `self.params`?

hexrd/ui/calibration/wppf_options_dialog.py
```diff
@@ -420,6 +421,136 @@ def create_wppf_params_object(self):
             params.add(**kwargs)
         return params
 
+    def add_material_parameters(self):
```
psavery, line 424: I think it'd be ideal if this were a function called by both `hexrd.WPPF` to add parameters, and this class as well. That way, if anything changes in the code, we won't have to change it in two places.

But we can leave it this way if you'd prefer.

psavery (commented):
Overall, I think it looks fine. See individual comments.

We want to fix the pep8 issues.

Also, I know we'll need to get the parameters to change when
the user changes methods. And, we can get rid of the temporary
WPPF object that is created once we start using the `add_material_parameters`
function.

I'll have more time to work on this once we get access to the LLNL funding.

hexrd/ui/calibration/wppf_options_dialog.py
saransh13: ok got it.

hexrd/ui/calibration/wppf_options_dialog.py
saransh13: ok yeah, should be `self.params`

hexrd/ui/calibration/wppf_options_dialog.py
saransh13: I think thats a good idea. It'll require some changes to the `WPPF` module in `hexrd`. I'll have a look at this.

hexrd/ui/calibration/wppf_options_dialog.py
joelvbernier: yes, endorse!

hexrd/ui/calibration/wppf_options_dialog.py
joelvbernier: perhaps put the common functions in a `def _lattice_param_helper()` or something to avoid code duplication.

joelvbernier:
Are we still waiting for a hexrd PR to be merged @saransh13 ?

saransh13:
@joelvbernier already submitted the associated `hexrdgui` PR for this.

joelvbernier (approved):

PR #773, opened 2021-02-25
Allow user to save all detectors at once

### LINKED ISSUE #758: Batch imageseries export
Some users have asked for the following updates/checks to the current imageseries export action under "File⇒Save⇒ImageSeries"
- [x] Rename menu item "Images"
- [x] The ability to batch save the imageseries for all detectors using a user-speifed file stem, to which the detector name will be appended (join with '_')
- [x] The option to save the current image (_e.g._ a single frame) or a multiframe imageseries that ignores any aggregation.  The latter case is mainly for the HEDM workflow, as a user might want to load an imageseries with aggregation to get a global view, but then save the associated `OmegaImageSeries` object with the proper processing options (sans aggregation) and omega metadata attached for external analysis.
I realized that this may require a new widget...

bnmajor:
Fixes #758

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -37,6 +37,7 @@ from hexrd.ui.mask_manager_dialog import MaskManagerDialog
 from hexrd.ui.mask_regions_dialog import MaskRegionsDialog
 from hexrd.ui.materials_panel import MaterialsPanel
 from hexrd.ui.powder_calibration_dialog import PowderCalibrationDialog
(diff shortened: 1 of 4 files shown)


joelvbernier (approved):
Just tested with PXRDIP, and it looks great!

bnmajor:
@joelvbernier There is a new PR for this since this branch has already been merged in - see #775

joelvbernier:
Saw it, thanks

PR #775, opened 2021-03-04
Reset omegas for each load

bnmajor:
Omega wedges were not reset between loading images (via stack dialog or load panel). Fixes bug.


bnmajor:
@joelvbernier This PR fixes the bug you were seeing

psavery (approved):
LGTM

joelvbernier:
Ah, ok.  Thanks.

PR #777, opened 2021-03-04
Remove structure factor bug work-around

psavery:
This is no longer needed because hexrd/hexrd#194 now generates
all structure factors instead of just the ones that aren't excluded.


cjh1 (approved):
LGTM

PR #778, opened 2021-03-04
Materials panel overhaul

### LINKED ISSUE #598: Remove sglite usage
This will be done after the demo, most likely.

We are currently using `hexrd.spacegroup` just a little bit in the material editor widget, to get things like the hall symbol, required lattice parameters, etc. We should move away from using that module to other alternatives in the repo, because sglite is going to be removed soon.

psavery:
This PR is going to re-design some parts of the materials panel, as per user requests.

The first step was to move some widgets out of the tab widget. You can see an example attached.
Screenshot from 2021-03-04 16-32-16

Some things left to do:
- [x] Move "Min d-spacing" out of "Space Group Info" into untabbed region as well
- [x] Use scientific spin boxes for all floats
- [x] Have user select lattice type, then space group, then space group setting
~~- [x] Make slider view for lattice parameters~~
- [x] Fix #598
- [x] Add c/a ratio editor for space groups 75 -  194


psavery:
So, apparently, we aren't using sglite after all. We are only using dictionaries in the `hexrd.spacegroup` module...

psavery:
@joelvbernier This currently just moves some of the pieces out of the "General" tab for the materials, and clarifies the "Computed" min d-spacing (used for hexrd computations including Laue overlays) vs. the "Displayed" min d-spacing (used to display powder overlays).

Screenshot from 2021-08-30 11-45-58

Although I didn't check off all the items, I think it's ready for merging, and we can check off more items later.

psavery:
@joelvbernier The merge conflicts were fixed.

joelvbernier:
Regarding @saransh13 comment above, can we add an optional widget to specify c/a for hexagonal/trigonal/rhombohedral lattice types?  A helper that will modify the c parameter based on input a and the ratio.

saransh13:
If space group is in the range 75-194 (inclusive), then we should let the users specify c/a and c (or a) should be set based on the value of a (or c)

PR #786, opened 2021-03-05
Remove some instrument_config_loaded signals

psavery:
Unfortunately, instrument_config_loaded signals cause cartesian and
polar parameters to be auto generated. Auto generating the polar
parameters in particular is very time consuming, and causes the GUI to
freeze.

Thus, we should only emit "instrument_config_loaded" when an instrument
config was actually loaded by the user via "File"->"Open"->"Configuration".

This commit attempts to remove all instrument_config_loaded signals except
for that one. Some explanations will be present in the PR.

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -207,7 +207,6 @@ class ImageLoadManager(QObject, metaclass=Singleton):
         if not self.update_status:
             self.update_needed.emit()
         if self.transformed_images:
-            HexrdConfig().instrument_config_loaded.emit()
             HexrdConfig().deep_rerender_needed.emit()
 
     def get_dark_aggr_op(self, ims, idx):
--- a/hexrd/ui/import_data_panel.py
+++ b/hexrd/ui/import_data_panel.py
@@ -53,7 +53,6 @@ class ImportDataPanel(QObject):
         self.ui.button_box.rejected.connect(self.clear)
         self.ui.save.clicked.connect(self.save_file)
         self.ui.complete.clicked.connect(self.completed)
-        self.new_config_loaded.connect(HexrdConfig().instrument_config_loaded)
         self.ui.bb_height.valueChanged.connect(self.update_bbox_height)
         self.ui.bb_width.valueChanged.connect(self.update_bbox_width)
         self.ui.line_style.currentIndexChanged.connect(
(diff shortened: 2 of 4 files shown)


hexrd/ui/image_load_manager.py
```diff
@@ -207,7 +207,6 @@ def finish_processing_ims(self):
         if not self.update_status:
             self.update_needed.emit()
         if self.transformed_images:
-            HexrdConfig().instrument_config_loaded.emit()
```
psavery, line 210: It looked like this had 2 effects:
1. Auto-generating cartesian and polar parameters
2. Calling `LoadPanel.config_changed()`, which only really does something if detectors were changed.

So, *hopefully* removing this won't have any negative side effects.

hexrd/ui/import_data_panel.py
```diff
@@ -53,7 +53,6 @@ def setup_connections(self):
         self.ui.button_box.rejected.connect(self.clear)
         self.ui.save.clicked.connect(self.save_file)
         self.ui.complete.clicked.connect(self.completed)
-        self.new_config_loaded.connect(HexrdConfig().instrument_config_loaded)
```
psavery, line 56: This had the same effects as https://github.com/HEXRD/hexrdgui/pull/786/files#r588765330

Hopefully, removing it won't have any side effects.

hexrd/ui/import_data_panel.py
```diff
@@ -121,7 +120,6 @@ def set_detector_defaults(self, det):
         for key, value in self.detector_defaults[det].items():
             HexrdConfig().set_instrument_config_val(
                 ['detectors', det, 'transform', key, 'value'], value)
-            self.detector_defaults[det]
```
psavery, line 124: I removed this here because it didn't seem to do anything.

hexrd/ui/load_panel.py
```diff
@@ -77,8 +77,7 @@ def setup_gui(self):
         self.ui.file_options.resizeColumnsToContents()
 
     def setup_connections(self):
-        HexrdConfig().detectors_changed.connect(self.detectors_changed)
-        HexrdConfig().instrument_config_loaded.connect(self.config_changed)
```
psavery, line 81: Since `LoadPanel.config_changed()` only does something if the detectors were changed, we are now hooking it up directly to `HexrdConfig().detectors_changed` instead, which should be emitted any time the detectors were changed.

 Also, since `LoadPanel.config_changed()` calls `LoadPanel.detectors_changed()`, only one connection was needed here.

bnmajor (approved):
Worked great for me, thank you!

PR #787, opened 2021-03-05
Add hkl_seeds to default indexing config

psavery:
For a first-time user, this will add the hkl_seeds to their indexing
config dictionary, which will prevent a key error in
`OmeMapsViewerDialog.setup_hkls_table()`.

For users of previous versions, this will also automatically add
the hkl_seeds to their indexing config dictionary (if it is missing),
which will also prevent the same key error.


bnmajor (approved):
LGTM, indexing worked perfectly. Thanks!

Issue #780, opened 2021-03-04
Normalize structure factor to be between 0 and 100


psavery:
In the materials table, the structure factor numbers are larger than they need to be. Since they have arbitrary scaling, rescale them so the minimum is 0 and the max is 100.

![image](https://user-images.githubusercontent.com/9558430/110042591-9dc55680-7d0b-11eb-9636-bf6fe7dc4f4c.png)

PR #789, opened 2021-03-06
Rescale structure factors to be between 0 and 100

psavery:
Since structure factors have arbitrary scaling, rescale them to a
range that's easier on the eyes.

The HKL list that is rescaled is the one displayed in the materials
table. Essentially, exclusions are ignored, but tThMax is considered.

example

Fixes: #780


cjh1 (approved):
LGTM

Issue #772, opened 2021-02-25
Update density when atom sites get modified


psavery:
The density is under "Properties" in the materials panel.

See https://github.com/HEXRD/hexrdgui/issues/767#issuecomment-785407192

PR #790, opened 2021-03-06
Update density when structure is modified

psavery:
example

Fixes: #772


psavery:
@saransh13 @joelvbernier Do you guys think we should use fewer sig figs for displaying the density?

joelvbernier (approved):

joelvbernier:
@saransh13 -- is there a summed total occupancy, say when you have a substitutional alloy and you want to put two atom types at the same site with fractional occupancies?

saransh13:
You mean a field in `Materials` for total occupancy of each site? The GUI imposes restrictions so that the total occupancy for each site can't be > 1. I don't think theres a variable in the class which stores that info right now.

Issue #774, opened 2021-03-03
Sync error when updating material


joelvbernier:
Just noticed that if I update a material interaction that forces a reinitialization (_e.g._, min d-spacing) and for which I have an activate overlay, the following things happen:

1. the exclusions get clobbered (they don't have to, but we'll have to code that fix up)
2. the `tThWidth` gets reset to the default value of 0.25°, but _does not_ update the value set in the associated attribute interaction in the overlay manager.  It _should_ adopt the value set in the overlay manager if it is specified for that material.

This may be related to other sync errors in the material package that @saransh13 is working on.

joelvbernier:
Fixing this probably will be related to issue #767

joelvbernier:
Also likely related: the structure factors are all coming back as 1.0:
![image](https://user-images.githubusercontent.com/1154130/109890295-eb728e00-7c3b-11eb-8e8d-0de9fa578945.png)

psavery:
> Just noticed that if I update a material interaction that forces a reinitialization (_e.g._, min d-spacing) and for which I have an activate overlay, the following things happen:
> 
> 1. the exclusions get clobbered (they don't have to, but we'll have to code that fix up)
> 2. the `tThWidth` gets reset to the default value of 0.25°, but _does not_ update the value set in the associated attribute interaction in the overlay manager.  It _should_ adopt the value set in the overlay manager if it is specified for that material.
> 
> This may be related to other sync errors in the material package that @saransh13 is working on.

So unfortunately, calling `_newPdata()` on the material currently erases all of that info.

I've actually used a [utility function in hexrdgui](https://github.com/HEXRD/hexrdgui/blob/c4b906171a7cafdea6c9786ac8d963921a5f275f/hexrd/ui/utils.py#L90-L102) to preserve three things on new plane data creation: exclusions, tThWidth, and tThMax.

It might make sense to do this kind of thing in hexrd (maybe inside `_newPdata()`), though. Because currently, when `_newPdata()` gets called in hexrd, the exclusions, tThWidth, and tThMax get erased. Some properties such as the [`dmin setter`](https://github.com/HEXRD/hexrd/blob/9e967b145c953cb2dc37a2e3a32f4e33736919ea/hexrd/material.py#L714-L725) do this, which is why setting the `dmin` in hexrdgui erases that info.

[...]

PR #791, opened 2021-03-07
Remove make_new_pdata from hexrdgui

psavery:
It seems that often (perhaps always), creating a new PlaneData object
from scratch is not necessary, and is a performance hit.

Try removing all cases of re-creating a PlaneData object from scratch,
and instead, we should ensure that when the relevant hexrd Material
attributes get modified, that the underlying PlaneData object will
get properly updated as well (including anything that needs to be
re-computed).

This currently seems to give a nice performance boost for modifying
the energy via the slider widget, and seeing the overlays move.

However, we're going to need to change some of the properties in
hexrd to update everything that is needed, and we should ensure
that everything in the GUI is updating as expected.

### DIFF
--- a/hexrd/ui/calibration/calibration_runner.py
+++ b/hexrd/ui/calibration/calibration_runner.py
@@ -6,7 +6,7 @@ from hexrd.ui.constants import OverlayType
 from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.line_picker_dialog import LinePickerDialog
 from hexrd.ui.overlays import default_overlay_refinements
-from hexrd.ui.utils import convert_tilt_convention, make_new_pdata
+from hexrd.ui.utils import convert_tilt_convention
 
 
 class CalibrationRunner:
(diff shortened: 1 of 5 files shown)


psavery:
An update: it was discovered that new PlaneData objects must be created when the HKLs change, because PlaneData objects currently have immutable HKLs.

However, rather than creating the new PlaneData objects in hexrdgui, we are moving the creation over to hexrd, in hexrd/hexrd#204, so that the Material object itself can discern when is appropriate to create a new PlaneData object,
and also how to do it (such as preserving old PlaneData object attributes).

Thus, we are going to continue with this PR, and remove `make_new_pdata` from hexrdgui. This, along with hexrd/hexrd#204, will fix #774.

psavery:
This one is ready for merging now. The PlaneData is automatically updated now in hexrd, so it doesn't need to be updated in hexrdgui anymore.

Can I get a review, @saransh13?

saransh13 (approved):
LGTM.

Issue #741, opened 2021-02-03
New WPPF warning


joelvbernier:
Now when I select WPPF, I get a warning in a dialog; not sure why.  This is occurring just in bringing up the WPPF widget, _not_ in executing the fit.  Initialization error @saransh13?
![Screen+Recording+2021-02-03+at+01 10 44 86+PM](https://user-images.githubusercontent.com/1154130/106810689-2b852780-6622-11eb-8625-5414d471bef9.gif)

psavery:
@saransh13 FYI, when the dialog is first initialized, it creates a LeBail object so that the dialog can see the material options that the LeBail object adds (such as `gold_a`), and add them to the table also. It only initializes the `LeBail` class, it doesn't call functions on it.

I'm guessing that error is coming from the `LeBail` initialization?

saransh13:
@psavery I think we can skip the LeBail initialization and initialize the extra parameters from the list of `Material` class objects itself. Theres also some modifications to be made for reading in the parameters for the `Rietveld` method. The `Rietveld` method has a slightly different list of parameters compared to the `LeBail` class. Need some help with that though.

psavery:
@saransh13 Okay. [Here](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/resources/calibration/wppf_params.yml) is a yaml file that contains the list of defaults for the LeBail table. We will probably make another yaml file for Rietveld, and when the user switches methods, load the defaults from the respective yaml file and populate the table.

saransh13:
@psavery I have a WPPF-update branch which updates the default parameter list and adds the rietveld.yml file containing the default parameters for that class.

psavery:
Sure! Let me know when you have the method to generate the extra parameters from a list of Material objects.

saransh13:
ok the function `add_material_parameters` [here](https://github.com/HEXRD/hexrdgui/blob/WPPF-update/hexrd/ui/calibration/wppf_options_dialog.py) should work. Warning that I haven't actually tested it. But if there are any syntax error, it should be minor.

saransh13:
ok the function `add_material_parameters`line 424 [here](https://github.com/HEXRD/hexrdgui/blob/WPPF-update/hexrd/ui/calibration/wppf_options_dialog.py#L424) should work. Just a warning that I haven't actually tested it. But if there are any syntax error, it should be minor.

saransh13:
[...]

PR #794, opened 2021-03-08
Re-work how extra wppf parameters are added

psavery:
This PR is primarily for re-working how the extra WPPF parameters are added. Instead of initializing a WPPF object and grabbing the extra parameters from it, the extra parameters are generated separately.

However, this PR also:

- Adds a toggle option for displaying the WPPF plot.
- Makes the Rietveld method functional (it now works - at least on a basic level).

example

Fixes: #741


joelvbernier:
@saransh13 -- I'd like to see an option where the user can evaluate the current model, potentially even a mode where you can live update the parameters (like Maud) to futz with an initial guess.  Not sure if we should make that part of this PR or make a new one.

joelvbernier:
@psavery and @saransh13 -- do we need to merge this to maintain the functionality of WPPF with the current hexrd master?  We can push the eval option I previously mentioned to another PR in that case, otherwise, let's add that feature to this one.

psavery:
@joelvbernier Well, this fixes the issue of #741. But otherwise, it's not urgent to merge.

joelvbernier:
gonna merge; the extra features I was alluding to require changes in hexrd, so we will handle that separately.

joelvbernier (approved):

PR #796, opened 2021-03-09
diamond atom position fix

saransh13:
fixing the atom position for diamond in the default material file. The data in there was for setting 2 but space group is setting 1 by default.


psavery:
Since git can't show changes on h5 files, I didn't check them. But I trust you.

PR #800, opened 2021-03-09
Handle null buffer case

### LINKED ISSUE #799: Panel buffer widget error
If the panel buffer in an instrument file is `null` (yaml yields `None`), then the GUI throws the following exception when trying to use the configure button:
```
Range to be used:  12.433911180496207  ->  721.0535961304395
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration_config_widget.py", line 388, in _on_configure_buffer
    dialog = PanelBufferDialog(detector, self.ui)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration\panel_buffer_dialog.py", line 34, in __init__
    self.update_gui()
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration\panel_buffer_dialog.py", line 141, in update_gui
    self.ui.border_x_spinbox.setValue(buffer[0])
TypeError: 'NoneType' object is not subscriptable
```
@psavery, did you set up that dialog, or was it @bnmajor?

cjh1:
Fixes #799

### DIFF
--- a/hexrd/ui/calibration/panel_buffer_dialog.py
+++ b/hexrd/ui/calibration/panel_buffer_dialog.py
@@ -133,7 +133,7 @@ class PanelBufferDialog(QObject):
 
             if isinstance(buffer, np.ndarray):
                 self.mode = CONFIG_MODE_NUMPY
-            else:
+            elif isinstance(buffer, list):
                 self.mode = CONFIG_MODE_BORDER
                 if np.isscalar(buffer):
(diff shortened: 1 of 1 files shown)


cjh1:
@joelvbernier This fixes the issue you were seeing.

psavery (approved):
LGTM

PR #801, opened 2021-03-10
Add Windows specific guard to rename

cjh1:
On the other platforms the directory is created by the QFileDialog. Apparently not on Windows, so add a guard for the rename logic.


cjh1:
@joelvbernier This should resolve your directory saving issue on Windows.

Issue #767, opened 2021-02-24
Material instantiation error


joelvbernier:
When adding a new material in the GUI, the symmetry-appropriate lattice parameter values are not being enforced.  It appears that when a new material is added, it is taking lattice parameters and structure values from the most recently selected material.  For example, if I select sapphire (trigonal) and make a new cubic (fcc) material, I get the following:
![image](https://user-images.githubusercontent.com/1154130/109069906-b1871200-76a6-11eb-94d4-5591ed5307fd.png)
Note that this leads to erroneous reflection indices in the materials table, which implies that the material object is actually getting the infeasible lattice parameters (note that for cubic they should be [a a a 90 90 90] for cubic).  **_Even after editing the free lattice parameter, the reflection table is still incorrect_**.  You have to de/reselect the space group to have it properly generate the reflection table.

There are clearly a few things going on with respect to properly updating the material class, but I propose that we implement a simpler set of default values for the lattice parameters and structure for adding a new material.  
- [ ] make the default lattice parameter(s) can be 5Å.
- [ ] the angular defaults come directly from the space group lattice type,  with the exception of monoclinic and triclinic.  in those cases, we can use 90 degrees for the free angles.
- [ ] make the default structure a single site with a Carbon atom at (0, 0, 0).

@saransh13: can you work with @psavery to make sure the material class is getting instantiated properly?

joelvbernier:
Related:  the density calculation in the properties tab is not getting updated when changing atom types.

psavery:
@saransh13 @joelvbernier What if we do one of the following?

1. Get the default `Material()` created by hexrd to use these options, and just use that when a new material is added. We'd have to modify the defaults [here](https://github.com/HEXRD/hexrd/blob/de82c86204316d2c5a618eab5048b4c0c095285e/hexrd/material.py#L84-L116) to do that. Looks like it is currently using 3.61 angstroms, and creating a single Nickel atom.

2. Create a `default_material.h5` file either in hexrd or hexrdgui, and have the defaults for our material inside there. When the user adds a new material, load the material from that file and use it.

joelvbernier:
I vote for 1!

psavery:
@joelvbernier Okay. Should we change [any of the defaults](https://github.com/HEXRD/hexrd/blob/de82c86204316d2c5a618eab5048b4c0c095285e/hexrd/material.py#L84-L116)? It's currently:
- One Ni atom
- Fm-3m (225) space group
[...]

PR #802, opened 2021-03-10
Use default Material() when a material is added

psavery:
Previously, we would perform a deep copy of the active material
to create a new material when a material was added.

Instead, just create a default `Material()` object and use that.

This PR also cleans up some of the updating in the materials panel to
make it easier and more intuitive to keep the panel up to date with
changes.

It also fixes renaming updates in the overlay manager as well as in
the materials table.

Fixes: #767


psavery:
I removed the quick fix, as it is no longer needed due to hexrd/hexrd#207 being merged, and hexrd/hexrd#205 being fixed.

@joelvbernier @saransh13 Thi sis ready

joelvbernier (approved):

Issue #798, opened 2021-03-09
Allow users to save aggregated images


bnmajor:
If the images are currently aggregated we should default to saving them that way but leave the user with the option to toggle the selection and save the unaggreated images instead (which is the current behavior).

bnmajor:
@joelvbernier I'm thinking we can just add the option to the bottom of the dialog, let me know if that's okay or if it's looking too crowded
![save_agg_images](https://user-images.githubusercontent.com/51238406/110531948-3849df00-80ea-11eb-8943-0d45f5453d06.png)

PR #804, opened 2021-03-11
Save agg images

bnmajor:
This addresses #798 

This is still a WIP at the moment because I am seeing weird behavior when saving out unaggreated images as HDF5. All other options work well.


bnmajor:
Note: the issue that I was seeing came from saving large data on a small machine - The two Ruby files I was loading for the GE instrument are about 2Gb a piece and took about a minute to save a an hdf5 file. We should look into adding a progress dialog (or something similar) for cases like this. Otherwise it just seems like the GUI is hanging.

joelvbernier (approved):

Issue #805, opened 2021-03-11
Automatically add polar masks for 2D panel buffers


psavery:
For the polar view, check if each detector has a 2D array for its panel buffer, and if it does, then create a mask from it for the polar view automatically.

joelvbernier:
Per our Slack thread, probably want to make this togglable since it may adversely affect the snip background removal.

psavery:
NaNs are now automatically added to the panel buffer regions in the plot because hexrd/hexrd#208 was merged.

psavery:
Closing, since they are automatically added...

PR #808, opened 2021-03-11
Apply 2D panel buffer as a mask to the polar view

psavery:
This checks if the panel buffer is a 2D numpy array, and if it is,
it will apply it as a mask to the polar view.

This seems to correctly generate NaNs in the output. However, the
image doesn't look so good. Maybe the mask is messing up the math
somewhere...

bad_image

Fixes: #805

### DIFF
--- a/hexrd/ui/calibration/polarview.py
+++ b/hexrd/ui/calibration/polarview.py
@@ -200,6 +200,12 @@ class PolarView:
         panel = self.detectors[det]
         img = self.images_dict[det]
 
+        if isinstance(panel.panel_buffer, np.ndarray):
+            if img.shape == panel.panel_buffer.shape:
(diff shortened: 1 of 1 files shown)


joelvbernier:
I guess we need to reconcile with the output of `hexrd.xrdutil.PolarView`, because that is looking ok with the panel buffers on:
image

joelvbernier:
@psavery, I ran the above interactively with the calib example you have in `run_calibration.py`

psavery:
@joelvbernier Oh nevermind with this PR. I thought I needed to something in hexrdgui to get the nans in the panel buffer region. But apparently not. They are now automatically added in since hexrd/hexrd#208 was merged. 

image

I now get good images, with NaNs in the panel buffer region as expected.

Issue #810, opened 2021-03-12
Save azimuthal lineouts to HDF5 file as well.


saransh13:
In addition to saving the polar view, a user requested that the 1-D azimuthal integation be saved as well. So in the file --> export --> current view, the azimuthal integation needs to be written as a nx2 vector. The first column in the two-theta values (from tth_coordinates) and the second column is the azimuthal integarted intensities.

PR #812, opened 2021-03-12
When exporting the polar plot, include integration

psavery:
This exports the azimuthal integration as well when exporting
the polar plot.

The data is currently formatted to be in two columns, where the first
column is the two theta, and the second column is the intensity.

The two theta is currently in degrees, which doesn't match the
`tth_coordinates` and `eta_coordinates` units of radians. Should
we use radians for the azimuthal integration as well, @saransh?

```python
>>> data['azimuthal_integration']
array([[   9.48379266,   49.80703472],
       [  10.05258401,   58.68515111],
       [  10.62137536,   63.75402841],

       ...

       [  69.20688447,   32.74983483],
       [  69.77567582,   36.18811307],
       [  70.34446717,   43.57711938],
       [  70.91325852,  161.0254772 ],
       [  71.48204988,   54.09543593],
       [  72.05084123,   46.3929216 ]])
```

Fixes: #810

### DIFF
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -65,12 +65,18 @@ class InstrumentViewer:
         self.img = self.pv.img
 
     def write_image(self, filename='polar_image.npz'):
+        azimuthal_integration = HexrdConfig().last_azimuthal_integral_data
+
+        # Re-format the data so that it is in 2 columns
+        azimuthal_integration = np.array(azimuthal_integration).T
+
         # Prepare the data to write out
         data = {
             'tth_coordinates': self.angular_grid[1],
             'eta_coordinates': self.angular_grid[0],
             'intensities': self.img,
-            'extent': np.radians(self._extent)
+            'extent': np.radians(self._extent),
+            'azimuthal_integration': azimuthal_integration,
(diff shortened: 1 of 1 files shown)


joelvbernier (approved):

psavery:
I'm a little concerned about the fact that the `eta_coordinates` and `tth_coordinates` are in radians, but the `azimuthal_integration` uses degrees. Should we make them the same to avoid confusion? What do you think, @saransh13?

psavery:
> I'm a little concerned about the fact that the `eta_coordinates` and `tth_coordinates` are in radians, but the `azimuthal_integration` uses degrees. Should we make them the same to avoid confusion? What do you think, @saransh13?

@joelvbernier What do you think about this? Currently, polar plot exported `eta_coordinates` and `tth_coordinates` are in radians, but the `azimuthal_integration` x units are in degrees. Should we make them the same to avoid confusion?

saransh13:
azimuthal_integration is mainly to prepare plots for publications etc.,
where we exclusively use degrees. If anything, I would probably update the
units of polar plot to degrees.

psavery:
@saransh13 Sounds good, I will change those to be in degrees.

Issue #814, opened 2021-03-12
Add support for BBXRD instrument


bnmajor:

bnmajor:
Note: Z rotation needed is -112.91

PR #815, opened 2021-03-12
Add option to load BBXRD instrument

bnmajor:
Addresses #814 

@joelvbernier We'll need a default config file and sample hdf4 file with the raw images to test this properly, but a quick test using the BBXRD boundaries on the PXRDIP images shows that the boundaries look correct at least so this should pretty much work as-is for the instrument (with these few changes to actually add it as an option).


bnmajor:
This is incredibly old. Closing for now and can re-open a new branch when/if we need it.

PR #816, opened 2021-03-16
Display warning for empty frames dark mode if none set

bnmajor:


bnmajor:
@joelvbernier I believe this fixes all of the issues that we discussed in the meeting yesterday

joelvbernier (approved):

PR #818, opened 2021-03-17
temp fix for AttributeError when trying to set unitcell property

joelvbernier:


psavery (approved):
LGTM. I'll put in a better fix for this soon.

Issue #779, opened 2021-03-04
Material structure tab overhaul


psavery:
Some changes for the material structure tab:

- [ ] Remove site selection table. Only add/remove atoms. Display all atoms in the atoms table.
- [ ] Add remove duplicate atoms button (see hexrd/hexrd#196; tracked in #769)
- [ ] Have +/- buttons for adding/removing atoms be "New" and "Delete"
- [x] Add an "apply" button that is grayed out when no changes have been made, and only apply material structure tab changes when the user presses the "apply" button. Don't make changes live.
- [ ] Default to no atoms
- [ ] Allow anisotropic editing of U (tracked in #513)

cjh1:
Last checkbox will be taken care of #769

PR #819, opened 2021-03-17
Add apply/reset buttons to MaterialStructureEditor

psavery:
When the material is modified, the updates no longer get pushed to
hexrd immediately. Instead, the apply/reset buttons become enabled,
and the user can either click "reset" to reset the settings to what
hexrd has, or "apply" to push the changes to hexrd.

https://user-images.githubusercontent.com/9558430/111512624-ef171200-871d-11eb-9e61-97d4da3f34e3.mp4

Addresses one of the items in #779

### DIFF
--- a/hexrd/ui/material_structure_editor.py
+++ b/hexrd/ui/material_structure_editor.py
@@ -66,10 +66,13 @@ class MaterialStructureEditor(QObject):
         self.ui.table.selectionModel().selectionChanged.connect(
             self.selection_changed)
 
+        self.ui.apply.pressed.connect(self.update_material)
+        self.ui.reset.pressed.connect(self.update_gui)
+
     def site_modified(self):
         site = copy.deepcopy(self.material_site_editor.site)
         self.sites[self.selected_row] = site
-        self.update_material()
+        self.material_edited()
 
     def ensure_scalar_U(self):
         # When we allow editing a tensor U, we can remove this function.
@@ -83,7 +86,7 @@ class MaterialStructureEditor(QObject):
         if tensor_encountered:
             name = self.material.name
             print(f'Warning: converting U tensors in {name} to scalars')
-            self.update_material()
+            self.material_edited()
 
     def name_changed(self, row, column):
         new_name = self.ui.table.item(row, column).text()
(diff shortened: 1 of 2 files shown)


psavery:
@saransh13 Can you check over 5b49fb8f1eb0590db3b6c6d4e8f18fe31c1b23c4 and make sure it looks good? That's the commit where we are setting the properties rather than re-creating the unit cell. I'm also assuming the density gets re-computed.

hexrd/ui/material_structure_editor.py
```diff
@@ -259,28 +259,9 @@ def update_material(self):
...
-            mat.unitcell.stiffness = old_unitcell.stiffness
-
-        # Compute the density
-        mat.unitcell.CalcDensity()
-
-        # Update the structure factor of the PData
-        mat.update_structure_factor()
+        mat.atominfo = np.array(info_array)
+        mat.atomtype = np.array(type_array)
+        mat.U = np.array(U_array)
```
saransh13, line 264: I think we can perform a check at this point to see if the values were changed or not. If those things are not checked, then the `unitcell` update is performed thrice, once for each set of values.

saransh13 (commented):
Other than the one comment, LGTM.

hexrd/ui/material_structure_editor.py
psavery: Should we do that here, or in the hexrd setters?

[Here](https://github.com/HEXRD/hexrd/blob/74131509ceb5bb83c2b835d60ffa701b97618cf5/hexrd/unitcell.py#L1413-L1417), for instance, could maybe be:

```python
    @U.setter
    def U(self, Uarr):
        if hasattr(self, '_U') and np.array_equal(self._U, Uarr):
            return

        self._U = Uarr
        self.aniU = False
        if(Uarr.ndim > 1):
            self.aniU = True
            self.calcBetaij()
```

That would cover any other cases where we set `U` as well in hexrdgui.

saransh13: Yeah, you're right. `hexrd` would be a better place for that.

joelvbernier:
I am gun shy about merging this one due to our slew of issues with material of late. If you are absolutely sure that this does not break anything then let's go ahead. Otherwise this can wait.￼

psavery:
This is ready by the way @joelvbernier

joelvbernier (approved):

PR #824, opened 2021-03-18
Specify tvec_c kwarg for polar powder overlays

psavery:
The [second arg in cart_to_angles is not tvec_c](https://github.com/HEXRD/hexrd/blob/74131509ceb5bb83c2b835d60ffa701b97618cf5/hexrd/instrument.py#L2457). So we need to
specify `tvec_c` as a keyword argument.

### DIFF
--- a/hexrd/ui/overlays/powder_diffraction.py
+++ b/hexrd/ui/overlays/powder_diffraction.py
@@ -139,7 +139,8 @@ class PowderLineOverlay:
 
             if display_mode == ViewType.polar:
                 # !!! apply offset correction
-                ang_crds, _ = panel.cart_to_angles(xys, self.instrument.tvec)
+                ang_crds, _ = panel.cart_to_angles(xys,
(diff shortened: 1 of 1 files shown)


hexrd/ui/overlays/powder_diffraction.py
```diff
@@ -139,7 +139,8 @@ def generate_ring_points(self, tths, etas, panel, display_mode):
 
             if display_mode == ViewType.polar:
                 # !!! apply offset correction
-                ang_crds, _ = panel.cart_to_angles(xys, self.instrument.tvec)
+                ang_crds, _ = panel.cart_to_angles(xys,
+                                                   tvec_c=self.instrument.tvec)
```
joelvbernier, line 143: This is not exactly correct, but more correct than it was.  There will be other changes here soon.

joelvbernier (approved):

PR #825, opened 2021-03-18
Only use 'image-files' for raw images

bnmajor:
This fixes a bug introduced in [this commit](https://github.com/HEXRD/hexrdgui/commit/7416b6aafd1b27782847217dcc2aa34e38e04a36).

The original approach was flawed, but was intended to handle the case where a single raw image is loaded with empty frames. This change uses the correct load path for each file type and instead offers the option of passing in a dict of options that may need to be set on (a) raw image(s). If the options are there they are added to the yaml file and used in creating the imageseries.


joelvbernier (approved):

psavery (commented):
LGTM

PR #830, opened 2021-03-19
Data should always be a dict, even if empty

### LINKED ISSUE #827: TARDIS Image load error
Traceback attached.
Screen Shot 2021-03-18 at 5 50 08 PM

bnmajor:
Fixes #827


psavery (approved):
LGTM

Issue #829, opened 2021-03-19
Add checkbox for threshold in eta omega maps generation options


psavery:
It runs much faster without a threshold. Add a checkbox to make it optional, and pass in `None` if it is not enabled.

PR #832, opened 2021-03-19
Improve HEDM Indexing and Grain Fitting UI

psavery:
To do:

- [x] Make thresholding optional for eta omega map generation
- [x] Make reflections table general so it can be used in other places
- [x] Have users explicitly pick material and hkls to use for eta omega map generation
- [x] In the maps viewer, make it clearer that the color map minimum is used for a threshold
- [x] In the fit grains options dialog, have the user explicitly pick hkls as well. Display the material being used.

Fixes: #829

### DIFF
--- a/hexrd/ui/indexing/create_config.py
+++ b/hexrd/ui/indexing/create_config.py
@@ -15,13 +15,19 @@ def create_indexing_config():
 
     # Make a copy to modify
     indexing_config = copy.deepcopy(HexrdConfig().indexing_config)
-    material = HexrdConfig().active_material
     omaps = indexing_config['find_orientations']['orientation_maps']
+    available_materials = list(HexrdConfig().materials.keys())
+    selected_material = omaps.get('_selected_material')
+
+    if selected_material not in available_materials:
+        raise Exception(f'Selected material {selected_material} not available')
+
+    material = HexrdConfig().material(selected_material)
     omaps['active_hkls'] = list(range(len(material.planeData.getHKLs())))
 
     # Set the active material on the config
     tmp = indexing_config.setdefault('material', {})
-    tmp['active'] = HexrdConfig().active_material_name
+    tmp['active'] = material.name
 
     # Create the root config from the indexing config dict
     config = RootConfig(indexing_config)
--- a/hexrd/ui/indexing/create_config.py
+++ b/hexrd/ui/indexing/create_config.py
@@ -1,5 +1,7 @@
 import copy
 
+import numpy as np
+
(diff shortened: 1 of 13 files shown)


psavery:
Here is the current status of my testing with the HEDM:

Compared to the CLI for our test case with the GE example. If I run fit grains directly from the accepted orientations file produced from the CLI, I get the same fit grains answer as the CLI to 4 decimal places.

If I run the GUI from the beginning using npz for one attempt and ge files (the ge attempt includes dark background subtraction via median and thresholding in the eta omega map generation), I get nearly the same results for npz compared to ge (they are off in tvec by a little bit), but the results look a little different compared to the CLI. You can see the results from the different runs attached.

[gui_grains_from_accepted_orientations.txt](https://github.com/HEXRD/hexrdgui/files/6174549/gui_grains_from_accepted_orientations.txt)
[gui_grains_from_beginning_npz.txt](https://github.com/HEXRD/hexrdgui/files/6174550/gui_grains_from_beginning_npz.txt)
[gui_grains_from_beginning_ge.txt](https://github.com/HEXRD/hexrdgui/files/6174551/gui_grains_from_beginning_ge.txt)

The tests above were performed skipping the conversion of tensor U to scalar U (which needs to be fixed at some point, referenced here: #513). The conversion of tensor U to scalar U modifies the structure factors a little bit, and it can result
in us changing which hkls we include (based on the structure factor cutoff). But a little test I performed seemed to show that
this difference doesn't have much of an effect on the fit grains output. I guess one or two hkls difference might not make a
huge difference, considering we were still using about 73 hkls.

The test files are attached below.
[NIST_ruby_single_GE_CLI_results.zip](https://github.com/HEXRD/hexrdgui/files/6174565/NIST_ruby_single_GE_CLI_results.zip)

psavery:
The issues mentioned in https://github.com/HEXRD/hexrdgui/pull/832#issuecomment-803201638 seem to have been caused by the hkls getting memoized [here](https://github.com/HEXRD/hexrd/blob/74131509ceb5bb83c2b835d60ffa701b97618cf5/hexrd/xrdutil.py#L1035-L1039). I added a quick fix for now via 056dc2771ea9ceab95de6e55ee05870b933a5251.

Now, loading from npz and running from scratch gives the same result as the CLI to 3 decimal places. Loading the GE files and running from scratch gives the same result as the CLI to 2 decimal places.

I think this is ready to go. It looks like the HEDM workflow is mostly working now.

joelvbernier (approved):

Issue #826, opened 2021-03-19
Switching materials doesn't change the label


saransh13:
Clicking on the newly created material doesn't change the label of the material. Everything else seems to work fine though. The properties/peak locations/lattice parameters etc. seem correct. Looks like the order of the materials in the table after a new material is created is not correct and leading to those bugs.
![output](https://user-images.githubusercontent.com/15834451/111713753-742d2480-880d-11eb-9959-b98f87a49318.gif)

Console error looks like the following. Maybe has something to do with the above error?
![Screen Shot 2021-03-18 at 5 32 45 PM](https://user-images.githubusercontent.com/15834451/111714953-0afae080-8810-11eb-88a0-cc51858b39a6.png)

psavery:
Unfortunately, it's kind of hard to see the gif. I can't read a lot of things like the space group.

Is that the full traceback as well? I was just wondering if there's anything below that `<listcomp>` line. Also, is there an exception occurring in `unicell.ChooseSymmetric`?

On Linux, it seems that changing to a new material changes the label. It might either be a Mac issue, or it might be caused by that exception.

saransh13:
Oh sorry about that. Using a higher res gif here. The `unicell.ChooseSymmetric` was error was due to an empty list being passed to the function. I'll post a fix for that, not sure about the other error though. I couldn't reproduce them this morning, but the labels were still glitching.

These are the steps I followed:
1. open the gui and create a new material with default values.
2. create another material with some random space group and lattice constants.
3. Try to go back to the first material that I created.
![output2](https://user-images.githubusercontent.com/15834451/111814389-a2efdd00-8897-11eb-83b1-ce837e94b8b3.gif)

psavery:
Looks like I was able to reproduce this on Linux as well. So it's a cross-platform issue.

Thank you for the higher resolution gif, Saransh. I was able to see exactly what you did and reproduce it :slightly_smiling_face:

psavery:
I believe I was able to fix this in #834.

I was unable to reproduce the exception, however. Let me know if you continue to encounter it and if you come up with some reproducible steps. It might be something separate...

PR #834, opened 2021-03-21
Avoid sorting keys in the materials panel

psavery:
This was causing the panel to be confused as to which key belonged
to which material.

Keep the keys in the same order to avoid this confusion.

Fixes: #826


psavery:
@saransh13 Can you double-check that this fixes the issue? It did for me.

joelvbernier (approved):

PR #835, opened 2021-03-21
Use ScientificDoubleSpinBox everywhere

psavery:
This also allows for a higher precision, up to 8 decimal places for
the properties modified.

### DIFF
--- a/hexrd/ui/resources/ui/calibration_crystal_editor.ui
+++ b/hexrd/ui/resources/ui/calibration_crystal_editor.ui
@@ -50,7 +50,7 @@
          </attribute>
          <layout class="QGridLayout" name="gridLayout_2">
           <item row="0" column="3">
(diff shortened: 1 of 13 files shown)


psavery:
@joelvbernier Should I just change the double spinboxes everywhere to be scientific spinboxes? I found a few places where we are still using double spin boxes, such as the options dialog for the auto powder calibration, the threshold for the eta omega maps select dialog, the laue crystal editor, image mode widget, fit grains options dialog, some slider widgets, laue overlay editor, overlay style picker, and a few more...

joelvbernier:
Yeah, search and replace :-)

psavery:
@joelvbernier This now changes scientific spin boxes to be used everywhere (except the ones that are already gonna be changed via #842).

I tested most of them, and I think this seems okay.

joelvbernier (approved):

PR #838, opened 2021-03-22
Use only required params for default refinements

psavery:
Before, it was adding all of the parameters to the default refinements.

Now, it should be making better defaults based upon which lattice
parameters are required.

Now, the default overlay refinements function accepts a whole overlay
dict, and it extracts from it the type and the material name in order
to come up with better default refinements.

### DIFF
--- a/hexrd/ui/overlays/__init__.py
+++ b/hexrd/ui/overlays/__init__.py
@@ -1,5 +1,7 @@
 import copy
 
+from hexrd import unitcell
+
 from hexrd.ui.overlays.laue_diffraction import LaueSpotOverlay
 from hexrd.ui.overlays.mono_rotation_series import (
     MonoRotationSeriesSpotOverlay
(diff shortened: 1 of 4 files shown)


hexrd/ui/overlays/__init__.py
```diff
@@ -61,7 +68,12 @@ def default_overlay_refinements(overlay_type):
     if overlay_type not in default_refinements:
         raise Exception(f'Unknown overlay type: {overlay_type}')
 
-    return copy.deepcopy(default_refinements[overlay_type])
+    refinements = copy.deepcopy(default_refinements[overlay_type])
+
+    # Only give it the required indices
+    indices = unitcell._rqpDict[material.unitcell.latticeType][0]
```
joelvbernier, line 74: Shouldn't we grad the parameter list from the `PowderCalibrator.params` (which are in this case, just the reduced lattice params), or `PowderCalibrator.full_params` (which is the full concatenated list, [instr params] + [reduced lattice params])

joelvbernier:
@psavery see above.

hexrd/ui/overlays/__init__.py
joelvbernier: @psavery ...

joelvbernier (changes requested):

joelvbernier (approved):

PR #845, opened 2021-03-23
Enable transform dialog after images are loaded

bnmajor:
Make sure that the transform dialog option is enabled no matter which path is used to load images.

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -196,6 +196,7 @@ class MainWindow(QObject):
         self.new_images_loaded.connect(self.update_hedm_enable_states)
         self.new_images_loaded.connect(self.color_map_editor.reset_range)
         self.new_images_loaded.connect(self.image_mode_widget.reset_masking)
+        self.new_images_loaded.connect(self.images_loaded)
         self.ui.image_tab_widget.update_needed.connect(self.update_all)
         self.ui.image_tab_widget.new_mouse_position.connect(
             self.new_mouse_position)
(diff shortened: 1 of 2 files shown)


hexrd/ui/main_window.py
```diff
@@ -196,6 +196,7 @@ def setup_connections(self):
         self.new_images_loaded.connect(self.update_hedm_enable_states)
         self.new_images_loaded.connect(self.color_map_editor.reset_range)
         self.new_images_loaded.connect(self.image_mode_widget.reset_masking)
+        self.new_images_loaded.connect(self.images_loaded)
```
psavery, line 199: Any chance this line:
```python
        self.new_images_loaded.connect(self.images_loaded)
```

Can be the only connection from `self.new_images_loaded`, and we just call all the functions it was connected to previously in `self.images_loaded()`? That would reduce the number of connections and make it a little more readable, I think...

psavery: Or it can be another function if `self.images_loaded()` is currently being used for other things, although I think the name `images_loaded` makes sense to be called from `self.new_images_loaded`.

bnmajor: The reason this was separate was to avoid enabling transforms when dummy images are loaded, since that also emits the `new_images_loaded` signal. Playing around with it though, "transforming" a dummy image is essentially harmless so if that's okay then yes, this can be condensed

psavery: We can fix this another time.

psavery (approved):
LGTM

PR #867, opened 2021-03-25
Update LLNL import workflow

### LINKED ISSUE #793: Avoid Image Re-load  for Import Data Panel
Change up the workflow a bit so that the user can:
- [x] Select an image to load
- [x] Select the detector
- [x] Draw the template based on selected detector
- [x] Do not display preview of cropped image after each selection
- [x] User should be able to transform the image after the detector boundary is drawn

This differs from the current workflow in that it will not require the image to be selected each time the detector is changed (although that will still be an option for instruments that need it). This means that for instruments like `PXRDIP` the image only needs to be loaded one time for all five detectors.

bnmajor:
This PR addresses the basic changes needed to avoid image reloading for the import data panel.

Fixes #793


bnmajor:
@joelvbernier This is ready to test whenever you're ready

bnmajor:
@joelvbernier @saransh13 All conflicts with master have been fixed on this branch and this is ready to be tested

joelvbernier:
ok!

joelvbernier:
Ok, some feedback:

1. The 90 degree rotation has disappeared and now the loaded image plate orientations are incorrect. 
image
They should all have the notched edges facing up like they used to.  I assume this got axed accidentally when the callback to plot the cropped image was removed...  Actually, looking at the tilts on the actual instrument, they are coming in all wrong as well.
2. I was a bit confused about the pxrdip widget layout; I think we should rename the "Select File" button to "Select Image"
3. Did the option to select the current instrument, or load a different one than the default not make it in?
4. The default image transform should be "Mirror about Horizontal", not "None".
5. Would it be possible to add another static plot object to leave the outlines of the previously selected crops drawn over the image?
6. Are we going to punt on trying to capture the arrow keys for now?

bnmajor:
@joelvbernier 

1. I apologize for losing the final 90 degree rotation, I will add that back in.
2. Sounds good, I'll rename that
3. I would like to add the changes for the instrument selection in a new branch. I do have a WIP for that, but I did not want to introduce too many changes at once. Once this is merged in I will open a few more PRs (instrument selection, remembering boundary position, default PXRDIP transform)
4. See number 3 - the PXRDIP default was not yet changed in this branch
5. I actually did forget about this feature. I can add that in as well, but again I think it may be good to add that in a separate PR.
6. See number 3

My thinking in avoiding putting everything in at once was that I wanted to minimize the possibility of introducing/overlooking bugs, but if you'd rather have all of this together a merge it all at once let me know and we can do that.

joelvbernier:
@psavery and @bnmajor -- could you run a profiler to see why it is taking ~10s to select the instrument in the LLNL import panel?

joelvbernier:
@bnmajor -- I noticed that the default transform is still "None" instead "Mirror about horizontal"

bnmajor:
@joelvbernier Yes - I made some comments above about those changes coming after this PR was merged in. I will add that change to this branch though. Let me know if the others should be added as well, or if they're okay to add in future PRs.

joelvbernier:
Ok, so the rotation is being applied correctly to the images, but the tilts on the default instrument are still getting mangled upon import.  You are not modifying the tilts unless the user rotates a crop mask, right?  Well, one way or another the tilts are getting modified:
image
For your ref, the default tilt parameters in the required Intrinsic ZXZ parameterization are:
```
detector 'IMAGE-PLATE-B'[0.0, 0.0, 0.0]
detector 'IMAGE-PLATE-D'[180.0, 90.0, -180.0]
detector 'IMAGE-PLATE-L'[90.0, 90.0, 180.0]
detector 'IMAGE-PLATE-R'[-90.0, 90.0, -180.0]
detector 'IMAGE-PLATE-U'[-0.0, 90.0, -180.0]
```
I also thought that if you load an instrument configuration that has the same keys as the current that it preserves the loaded imageseries (?), but the imported images are getting clobbered if I reload with an instrument having the same detector keys (I suppose they might be in a different order).

I also tested the TARDIS import (which I thought we were going to leave as it was...), and get the following error when trying to load an HDF5 file:
```
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\import_data_panel.py", line 160, in load_images
    if not self.canvas.raw_axes[0].get_autoscale_on():
IndexError: list index out of range
```

bnmajor:
I'm still looking at what is happening to the tilts... the values that are set on completion here do not match the loaded config in the GUI and I haven't yet figured out why.

The TARDIS import should remain unchanged. I am loading the TARDIS instrument and images on this branch successfully at the moment. I am using the HDF5 images found in the shared drive under `sample_images` > `LLNL_data` > `TARDIS`

joelvbernier:
TARDIS tilts are messed up too.
image

joelvbernier:
If I load a saved config and cropped images:
image

bnmajor:
@joelvbernier The Euler angle convention was being called twice which seems to have been to blame for the tilts being completely wrong. With the latest fix I get the following:

pxrdip_polar
pxrdip_raw
tardis_polar
tardis_raw

Sample output that I produced:
[output_images_and_configs.zip](https://github.com/HEXRD/hexrdgui/files/6247484/output_images_and_configs.zip)

joelvbernier:
That looks better, but looks like there are merge conflicts to resolve.

bnmajor:
Yes, those were caused by the branches that I just merged in. Conflicts should be resolved now. Also the import panel now supports adjusting the boundary position with the arrow keys and will remember the previous boundary position.

joelvbernier (approved):

joelvbernier:
Testing this now and getting the following `AttributeError` when trying to draw detector outlines:
```
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\import_data_panel.py", line 218, in add_template
    self.it.create_shape(
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\interactive_template.py", line 39, in create_shape
    self.update_position(instr, det)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\interactive_template.py", line 55, in update_position
    elif self.instrument == 'PXRDIP':
AttributeError: 'InteractiveTemplate' object has no attribute 'instrument'
```

PR #869, opened 2021-03-29
Save/load .hexrd instrument config files

psavery:
These are HDF5 files that hexrd is now capable of saving/loading.

This is now the default instrument config file type, but yaml files
can still be loaded as well.

Now also, the way the files are saved/loaded internally has been changed.
Before, a dict was just written out to a yaml file. Now, an HEDMInstrument
object is created and used to write out either the yaml/hdf5 file.

Also, save/load of .hexrd directories is no longer supported, since it is
not needed anymore.

A util function was also added to extract out the config dict from an
HEDMInstrument object and perform any necessary conversions.

### DIFF
--- a/hexrd/ui/utils.py
+++ b/hexrd/ui/utils.py
@@ -245,3 +245,39 @@ def exclusions_off(plane_data):
 def has_nan(x):
     # Utility function to check if there are any NaNs in x
     return np.isnan(np.min(x))
+
+
+def instr_to_internal_dict(instr, calibration_dict=None):
+    from hexrd.ui.hexrd_config import HexrdConfig
+
+    # Convert an HEDMInstrument object into an internal dict we can
+    # use for HexrdConfig.
+
+    if calibration_dict is None:
+        calibration_dict = {}
+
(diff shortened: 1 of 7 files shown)


hexrd/ui/utils.py
```diff
@@ -245,3 +245,39 @@ def exclusions_off(plane_data):
...
+    finally:
+        # Restore the panel buffers
+        for key, panel in instr.detectors.items():
+            panel.panel_buffer = panel_buffers[key]
+
+    # Set the panel buffers on the detectors
+    for key, val in panel_buffers.items():
+        config['detectors'][key]['buffer'] = val
+
+    # Convert to the selected tilt convention
```
joelvbernier, line 278: This shouldn't be necessary, but let me check.

joelvbernier: got it!

joelvbernier (approved):

joelvbernier:
LGTM!

psavery:
This fulfilled some of the requirements for hexrd/hexrd#239

PR #871, opened 2021-03-31
Do not enable thresholding by default

### LINKED ISSUE #870: Image series is list when running indexing
I don't know why, but, using the master branches of both hexrd and hexrdgui, when I try to run the indexing on the GE example, I get an error:

```python
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/indexing/run.py", line 87, in ome_maps_selected
    config = create_indexing_config()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/indexing/create_config.py", line 59, in create_indexing_config
    load_omegas_dict = {
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/indexing/create_config.py", line 60, in <dictcomp>
    k: ims for k, ims in ims_dict.items() if 'omega' not in ims.metadata
AttributeError: 'list' object has no attribute 'metadata'
```

If I click "Read Files" a second time, I am then able to run the indexing without any issue.

I'm using [this config](https://github.com/HEXRD/examples/blob/77671d621f42d3852d44eab24e36f78ba47a3a24/NIST_ruby/single_GE/include/ge_detector.yml) and [this data](https://github.com/HEXRD/examples/blob/77671d621f42d3852d44eab24e36f78ba47a3a24/NIST_ruby/single_GE/imageseries/RUBY_0000-fc_GE.npz).

You can see a full example of it happening in the video below.

https://user-images.githubusercontent.com/9558430/113046036-2ac1cb00-9165-11eb-8ccf-f44b36ef2a84.mp4

bnmajor:
Fixes #870

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -377,7 +377,7 @@ class MainWindow(QObject):
         self.update_color_map_bounds()
         self.update_hedm_enable_states()
         self.color_map_editor.reset_range()
-        self.image_mode_widget.reset_masking(enabled)
+        self.image_mode_widget.reset_masking()
 
     def open_aps_imageseries(self):
         # Get the most recent images dir


hexrd/ui/main_window.py
```diff
@@ -377,7 +377,7 @@ def images_loaded(self, enabled=True):
         self.update_color_map_bounds()
         self.update_hedm_enable_states()
         self.color_map_editor.reset_range()
-        self.image_mode_widget.reset_masking(enabled)
+        self.image_mode_widget.reset_masking()
```
psavery, line 380: So this means masking should be disabled every time new images are loaded, then?

bnmajor: Yes, unless there is a reason to default to threshold masking on for every new image?

Also, thinking about this a little more no matter whatever the default should be, I am thinking the masking may not be being handled properly. Right now when a threshold mask is applied a copy of the imageseries is created and handled as an array. Should this be handled more like the aggregated images where we access the copy of the ProcessedImageSeries when thresholding is toggle on? Or should the original imageseries be getting updated as the masking is changed (and if so should it happen with each change or should the user have to "accept" the current settings before the update)?

psavery (approved):
LGTM. Good enough for now. Let's think more, though, about the discussion brought up in https://github.com/HEXRD/hexrdgui/pull/871#discussion_r605006814

PR #873, opened 2021-03-31
Make UiLoader a Singleton

psavery:
We don't need to instantiate a new Uiloader object for every widget we create. Make UiLoader a singleton instead.

This also moves all the singleton classes to a single file and imports them to avoid code duplication.


joelvbernier:
I'll take @psavery and @bnmajor's work on this one.

bnmajor (approved):
LGTM!

Issue #859, opened 2021-03-25
Reduced parameter length mismatch in Powder Calibration if lattice parameter(s) are checked


joelvbernier:
I accidentally left the lattice parameter checked to refine while using the calibrator and got the following `ValueError`
```
Traceback (most recent call last):
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration\auto\powder_runner.py", line 162, in run_calibration
    ic.run_calibration(**kwargs)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration\auto\instrument_calibrator.py", line 53, in run_calibration
    resd0 = obj_func(x0, data_dict)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration\auto\powder_calibrator.py", line 344, in residual
    return self._evaluate(reduced_params, data_dict)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration\auto\powder_calibrator.py", line 302, in _evaluate
    full_params[self.flags] = reduced_params
ValueError: NumPy boolean array indexing assignment cannot assign 2 input values to the 3 output values where the mask is true
```
Time to go check where it is getting the parameters/flags.  Maybe best to just write an `update` method on the `PowderCalibrator' class?

psavery:
Looks like this is the same as #855

psavery:
It looks to me that this is caused by the `InstrumentCalibrator` [passing reduced flags to the residual function](https://github.com/HEXRD/hexrdgui/blob/8abd42a0fca9830a85985b01841f30d5dc3def6e/hexrd/ui/calibration/auto/instrument_calibrator.py#L49-L53), but the [flags in the `PowderCalibrator` class are full length](https://github.com/HEXRD/hexrdgui/blob/8abd42a0fca9830a85985b01841f30d5dc3def6e/hexrd/ui/calibration/auto/powder_calibrator.py#L24-L26).

The reduced length flags does not include lattice parameter refinements.

[...]

Issue #855, opened 2021-03-25
ERROR message from CeO2 calibration for FF HEDM


minety:
<img width="1440" alt="Screen Shot 2021-03-24 at 8 28 35 PM" src="https://user-images.githubusercontent.com/19909710/112401060-13578d80-8ce0-11eb-9cdd-7157e133ee71.png">

Another question, do I need to consider distortion (GE_41RT) when dealing with ge3 files from APS sector1?

Thanks,
Ye Tian

psavery:
@minety We are working on that error now in #859. It happens if any of the lattice parameters are refineable (which, they are by default). You can turn off the lattice parameter refinements via the "Materials" page on the left-hand side, and clicking on "Overlay Manager", then clicking on the row of your overlay (click on the white space of the row next to the "Visible" checkbox), and unchecking the refinements for your lattice parameters, as shown below:
![image](https://user-images.githubusercontent.com/9558430/112496837-117ae200-8d53-11eb-9014-1fe1159b908f.png)

joelvbernier:
@minety -- you'll want to make sure the lattice parameters are not refineable for the CeO2 anyhow, since this would give bad results.  We may change the default behavior to have them all unchecked (seems most reasonable @psavery) in the process of fixing the bug giving the `ValueError`.

As far as the ge3 goes, yes, you should consider the distortion.  I am going to upload a single GE CeO2 example to the [test problem repo](https://github.com/HEXRD/examples) right now.

joelvbernier:
@minety -- this should be fixed!

PR #875, opened 2021-03-31
update instrument calibrator and runner

joelvbernier:
Fixes: #855 
Fixes: #859

### DIFF
--- a/hexrd/ui/calibration/auto/instrument_calibrator.py
+++ b/hexrd/ui/calibration/auto/instrument_calibrator.py
@@ -31,8 +31,7 @@ class InstrumentCalibrator(object):
         for calib in self._calibrators:
             assert calib.instr is self._instr, \
                 "all calibrators must refer to the same instrument"
-            self.full_params += calib.params
-        self.full_params = np.asarray(self.full_params)
+            self.full_params = np.hstack([self.full_params, calib.params])
 
     @property
     def instr(self):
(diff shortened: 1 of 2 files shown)


hexrd/ui/calibration/auto/instrument_calibrator.py
```diff
@@ -64,22 +155,25 @@ def run_calibration(self,
...
+                print('OPTIMIZATION SUCCESSFUL!!!')
             else:
                 print('no improvement in residual!!!')
-                step_successful = False
+            print('initial ssr: '
+                  f'{sum(resd0**2)/float(len(resd0))}')
+            print('final ssr: '
+                  f'{sum(resd1**2)/float(len(resd1))}')
 
+            step_successful = False
```
psavery, line 176: Looks like this is causing the loop to end after 1 iteration, is that correct?

hexrd/ui/calibration/auto/instrument_calibrator.py
```diff
@@ -19,43 +41,112 @@ def instr(self):
...
+        master_data_dict_list = []
+        for calib_class in self.calibrators:
+            master_data_dict_list.append(
+                calib_class._extract_powder_lines(
+                    fit_tth_tol=fit_tth_tol, int_cutoff=int_cutoff
+                )
+            )
+        return master_data_dict_list
+
+    def residual(self, x0, master_data_dict_list):
```
psavery, line 89: Looks like the argument `x0` is not being used. Is that okay?

joelvbernier: Nope!   Fixed.

hexrd/ui/calibration/auto/instrument_calibrator.py
joelvbernier: bad indent; fixed.

psavery (approved):
LGTM

Issue #876, opened 2021-04-01
Error in reading instrument config


joelvbernier:
@psavery -- looks like there might be a bug when trying to read detectors with a distortion function (at least the GE_41RT style).  Seems to at least be in the `get_instrument_config_val` method.  Try to load the[ single_GE config](https://github.com/HEXRD/examples/blob/master/NIST_ruby/single_GE/include/ge_detector.yml) as a test.
```
Traceback (most recent call last):
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\main_window.py", line 624, in update_config_gui
    self.calibration_config_widget.update_gui_from_config()
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration_config_widget.py", line 196, in update_gui_from_config
    self.update_detector_from_config()
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\calibration_config_widget.py", line 235, in update_detector_from_config
    status_val = self.cfg.get_instrument_config_val(full_path)
  File "c:\users\berni\documents\github\hexrdgui\hexrd\ui\hexrd_config.py", line 842, in get_instrument_config_val
    cur_val = cur_val[val]
TypeError: 'int' object is not subscriptable
```

PR #877, opened 2021-04-01
Check for np.ndarray when making status lists

psavery:
It looks like in some cases, the instrument config value lists can be
np.ndarray now. We noticed this specifically for distortion parameters.

Add a check for np.ndarray as well when creating statuses, so that a
list of statuses will be properly created when it receives an np.ndarray.

Fixes: #876


bnmajor (approved):
LGTM

PR #882, opened 2021-04-02
Support fine-tuning boundary position with arrow keys

bnmajor:
Arrow keys can be used to translate or rotate the detector boundaries.


psavery (approved):
LGTM

PR #887, opened 2021-04-09
Detector bounds

bnmajor:
Two minor updates:

- Remove the limitations on resizing detector bounding boxes. The size is now limited by the size of the image rather than the default template size.
- Allow the focus to be set on the canvas again by clicking on it. Before, changing the bounding ox pulled the focus to the spin box and prevented the user from using the keyboard to move the template unless they toggled the translate/rotate option.


bnmajor:
@joelvbernier This should be ready to go - user can now make the bounds as large as the image if they'd like

joelvbernier (approved):

PR #892, opened 2021-04-15
Add basic refinements editor

psavery:
This adds a basic refinements editor under the "Edit"->"Refinements"
menu item. It allows the user to edit both the refinements and values
of refinable parameters that are used for calibration.

It additionally comes with a combo box of actions to automatically set 
refinements. Currently, the only action is "Mirror first detector", which
causes the refinement settings of all of the detectors to match what is
currently set for the first detector.

It is hopefully made in a way so it will be easily extensible to include
additional things such as lower bounds and upper bounds, and other
types of constraints.

https://user-images.githubusercontent.com/9558430/114899027-3f929580-9dd8-11eb-8569-966b7d209409.mp4


joelvbernier:
I am *LIKING* this

joelvbernier (approved):

PR #897, opened 2021-04-21
Allow clearing selected files in image stack dialog

bnmajor:
@joelvbernier This adds a clear option for selected files in the image stack dialog. This does not clear the load panel if images were previously selected. Do we want the load panel table to be cleared as well?

### DIFF
--- a/hexrd/ui/image_stack_dialog.py
+++ b/hexrd/ui/image_stack_dialog.py
@@ -166,6 +166,8 @@ class ImageStackDialog(QObject):
 
     def detector_selection(self, checked):
         self.state['all_detectors'] = checked
+        self.ui.single_detector.setChecked(not checked)
+        self.ui.search_directories.setEnabled(checked)
         self.ui.detector_search.setEnabled(checked)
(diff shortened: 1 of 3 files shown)


joelvbernier:
yes to clearing the load panel table :-)

joelvbernier:
Is this one ready too?

bnmajor:
@joelvbernier I hadn't updated this branch with to clear the load panel table but that has been added now. This also cleans up some bugs I found while testing out the changes. Specifically:
- If previously selected images could not be found the file selector was being opened before the dialog without any explanation and it would be quite unclear to the user both why this was happening and what they were expected to do. Now they are instead provided with a warning dialog that explains that the previous images could not be found and that they should double-check their currently selected file paths.
- Toggling between detectors was duplicating any existing omega data each time the detector was switched. This has been fixed.
- If the last state of the image stack dialog was selecting directories manually for each detector the GUI was not properly updated the next time the dialog was opened - the wrong items were selected/enabled. This has been fixed as well.

This should be ready for review now.

joelvbernier (approved):

PR #905, opened 2021-04-30
Prev det bounds

### LINKED ISSUE #884: Keep previous detector borders
If the import panel image has not been re-loaded for the next detector, keep the previous detector boundary around to make it more obvious which detectors have been already selected.

- Maybe automatically change the default color and/or opacity for previous boundaries to make it clear which is which

bnmajor:
Fixes #884

This branch adds a feature to keep around any previously accepted bounds any time a new detector is selected but the selected image remains unchanged. Currently detector bounds retain their selected color and line width, but the line style is changed to dashed to indicate that it has been accepted. This behavior could easily be changed if there is another way that might be best to represent the accepted bounds.

### DIFF
--- a/hexrd/ui/import_data_panel.py
+++ b/hexrd/ui/import_data_panel.py
@@ -40,7 +40,6 @@ class ImportDataPanel(QObject):
         self.detector_defaults = {}
         self.cmap = cmap
         self.detectors = []
-        self.editing = False
 
         self.set_default_color()
         self.setup_connections()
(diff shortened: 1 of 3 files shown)


bnmajor:
remember_outlines

bnmajor:
@joelvbernier This should be ready to test out

joelvbernier:
Oh???  Looks nice!

joelvbernier (approved):

Issue #820, opened 2021-03-17
Changes to WPPF Dialog


psavery:
Rather than having an ok/cancel button, have these buttons instead:

- [ ] Run - it runs WPPF, keeping the dialog open, and updating the parameters with their new values
- [ ] Compute - calculates the spectrum given the values of the parameters, to see if we are close to the answer or not

These changes especially make sense in light of #795 and #811

From my discussion with @saransh13

saransh13:
@psavery Can you please hold off any work on this for this week? I'm adding more options to choose additional peak shapes (in addition to the default one) to use for fitting. The user will have the option to choose the peak shapes depending on the input. Each of these shapes comes with its own set of default parameters which will need to be added to the WPPF dialog. Once I have everything testing and merged, we can start work on this again. I think should be only a couple of days.

Issue #811, opened 2021-03-12
Save simulated lineouts


saransh13:
Once WPPF finishes running, the user wants to save the simulated lineouts for pubications etc. Add an option to save the simulated pattern once the run finishes. The data is located in `two_theta, intensity = LeBail.spectrum_sim.data` or `two_theta, intensity = Rietveld.spectrum_sim.data` depending on which option was selected in WPPF. HDF5 could be a good format to write out this data.

The function should also write all the parameters of the class in the file as well.

joelvbernier:
Ooops -- @psavery is there another PR to link to this related to WPPF?

Issue #795, opened 2021-03-08
Save WPPF objects after run and optionally re-use them


psavery:
From https://github.com/HEXRD/hexrdgui/issues/741#issuecomment-772843193

Basically, provide a way for the users to choose whether or not to re-use previous WPPF objects when they perform a new WPPF run.

This issue will currently wait for @saransh13 to provide new properties on the WPPF classes so that we can update it like `obj.params = params`.

Issue #782, opened 2021-03-05
Add snip as background method in WPPF


saransh13:
`WPPF` now has functionality to use `hexrd.imageutil.snip1d_quad` for background estimation. We need to add this to the GUI. We also discussed a way to streamline this, so we don't have to do this each time I add something to WPPF. LMK if theres something on the `hexrd` side that needs to be done for that.

psavery:
Sure, we can add the option and widgets for snip.

About streamlining: so, if the background methods didn't require any parameters, it would be easy to automatically keep hexrdgui in sync with it. For instance, in hexrd, there could just be a list of possible background methods:
```python
background_methods = [
    'chebyshev',
    'spline',
    'snip',
]
```

We could just read this list and automatically populate the combo box with the items from the list.

If they all required one parameter of the same type (such as float), that would also be easy, as we could just use
a list like the one above and assume they all have the same number of parameters and the same type.

With varying numbers of parameters and parameter types, things start to get a little more complicated. In hexrd, we'd probably have to define a structure that provides parameter information like the following:

```python
background_methods = [
    'spline': None,
    'chevyshev': [
        {
            'label': 'Chebyshev Polynomial Degree',
            'type': int,
        }
    ],
   'snip': [
        {
            'label': 'Snip Width',
            'type': float,
        },
        {
            'label': 'Snip Num Iterations',
            'type': int,
        }  
    ],
]
```

Then in hexrdgui, we could automatically construct widgets with the label and spinbox type (or other parameter type) to match. That way, you could arbitrarily add background methods and hexrdgui would automatically create the widgets based on what is in hexrd.

If you think you might be adding a lot of background methods, then this might be a good path to go down.

PS: a dict structure like
```python
{
    'label': 'Chebyshev Polynomial Degree',
    'type': int,
}
```

Is nice mostly because it provides more flexibility in case we want to add things in the future. For example, if
you'd want to add min/max constraints, or even tooltips:
```python
{
    'label': 'Chebyshev Polynomial Degree',
    'type': int,
    'min': 0,
    'max': 99999,
    'tooltip': 'The polynomial degree used for Chebyshev.',
}
```

saransh13:
What would be a good location to define the `background_method` dictionary you mentioned above? In the `WPPF` module itself?

psavery:
> What would be a good location to define the `background_method` dictionary you mentioned above? In the `WPPF` module itself?

Yeah, that would be fine.

PR #906, opened 2021-04-30
Overhaul WPPF workflow

psavery:
This involves a lot of design changes to the WPPF workflow in order to
better work with the changes coming to hexrd. This depends on the
"Peak-Profile" branch of hexrd. Some of the changes are outlined below:

1. Instead of ok/cancel, the dialog now has a "Run" button which runs
   refinement cycles. The dialog is kept open, and the WPPF object is
   kept around, so that the user can run again with the same WPPF object
   if they so choose. There is also a "Reset Object" button so that they
   may reset the object.
2. Internally, we are now loading all default parameters by calling
   functions in hexrd's WPPF module. We do not need the default yaml files
   any more.
3. This adds a "Peak Shape" combo box where the user may choose the
   peak shape. The parameters in the table are updated accordingly by
   calling functions in hexrd.
4. The background method parameters are automatically generated from a
   dict that specifies the requirements in hexrd.
5. The user now has the option to save the plot as well to a file, in
   either HDF5 or npz format.
6. The user also has the option to preview the computed spectrum from the
   current parameter values, which brings up a plot of the spectrum.

Depends on the Peak-Profile branch in hexrd

Fixes: #782
Fixes: #795
Fixes: #811
Fixes: #820

Still to do:

image


joelvbernier:
@psavery and @saransh13 -- looks like this is still a WIP?

joelvbernier (approved):

PR #907, opened 2021-04-30
Use draw_idle() everywhere instead of draw()

psavery:
draw_idle() will perform a draw() when control returns to the Qt
event loop. It has the potential for major performance increases
because multiple calls to draw_idle() functions in a slot will
only result in one draw() when control returns to the event loop.

It resulted in major performance boosts for the brightness and
contrast editor, and I'm hoping it will give performance boosts
elsewhere as well. Just need to double check that we didn't call
draw_idle() on some object that doesn't have that function...

Fixes: #654

### DIFF
--- a/hexrd/ui/hand_drawn_mask_dialog.py
+++ b/hexrd/ui/hand_drawn_mask_dialog.py
@@ -66,7 +66,7 @@ class HandDrawnMaskDialog(QObject):
 
         self.canvas.mpl_disconnect(self.bp_id)
         self.bp_id = None
-        self.canvas.draw()
+        self.canvas.draw_idle()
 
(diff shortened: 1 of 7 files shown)


psavery:
@bnmajor Can you check out this branch, and check your affected code (masks, interactive templates, and anything else I might have missed) to make sure they still work and look good?

The only thing I have found that doesn't work with draw_idle() so far is if you have a `Cursor` object with `useblit=True`, because it uses animation to render the lines.

bnmajor:
@psavery This is looking good to me, everything seems to work very nicely.

bnmajor (approved):
LGTM

PR #911, opened 2021-05-04
Install QDialog enter key filter upon Ui load

psavery:
This removes the need to install this filter in every dialog's
__init__ method.

Now, any QDialog generated via the UiLoader will automatically install
the enter key filter. Any QDialog not generated via the UiLoader may
need to still install the enter key filter manually.


bnmajor (approved):
LGTM, very nice simplification!

PR #912, opened 2021-05-04
Support empty frames for hdf5 and npz files

### LINKED ISSUE #910: Handle empty frames for all image types
The `frame-cache` and `hdf5` adapters do not support empty frames but a user may need still need to set them when loading`.hdf5` or `.npz` files. For now we can use a shortcut to get around with the `frame_list` keyword argument for the `ProcessedImageSeries`.

bnmajor:
Fixes #910 

@joelvbernier This should solve the issue of setting empty frames on files that don't use the `image-files` adapter.

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -336,7 +336,13 @@ class ImageLoadManager(QObject, metaclass=QSingleton):
             ims_dict[key].metadata['omega'] = omw.omegas
 
     def get_range(self, ims):
-        return range(len(ims))
+        start = 0
+        if len(ims) != self.data.get('total_frames', [len(ims)])[0]:
+            # In the case of hdf5 and npz files we need to handle empty
+            # frames via the ProcessedImageSeries 'frame_list' arg to slice
+            # the frame list
+            start = self.data.get('empty_frames', 0)
+        return range(start, len(ims))
 
     def get_flip_op(self, oplist, idx):
         if self.data:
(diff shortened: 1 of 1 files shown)


hexrd/ui/image_load_manager.py
```diff
@@ -336,7 +336,13 @@ def add_omega_metadata(self, ims_dict):
...
 
     def get_range(self, ims):
-        return range(len(ims))
+        start = 0
+        if len(ims) != self.data.get('total_frames', [len(ims)])[0]:
+            # In the case of hdf5 and npz files we need to handle empty
+            # frames via the ProcessedImageSeries 'frame_list' arg to slice
+            # the frame list
+            start = self.data.get('empty_frames', 0)
+        return range(start, len(ims))
```
joelvbernier, line 345: So I think we need to replace `len(ims)` with the `steps` values from the omega metadata table.  This will also alow use to trim frames off the end as well...  right?

joelvbernier: we'll come back to this if need be...

joelvbernier: @bnmajor -- what do you think?

bnmajor: @joelvbernier range creates the list of frames to use - in this case that list of frames starts at the value of `empty-frames` and ends at the last frame of the imageseries for the `hdf5` and `npz` files. The only way to specify a different value for the max frames per file or max total frames is through the image stack dialog right now, and the only way to load `hdf5` and `npz` files is through the load panel. Since this check only handles the `hdf5` and `npz` files I think that setting the stop value as the last frame makes sense right now (since the user can no longer change that value in the table).

If we add the option to change the number of steps in the load panel or add support for loading `npz` or `hdf5` files through the image stack dialog then yes, this range will need to be calculated differently. Let me know if I'm misunderstanding the use case you're thinking of

joelvbernier (approved):

PR #913, opened 2021-05-04
Add value() function to progress dialog

psavery:
This mimics the value() function from QProgressDialog...


bnmajor (approved):
Fixes the error I was seeing, thanks!

PR #923, opened 2021-05-06
Use draw() in OmeMapsViewerDialog

psavery:
draw_idle() is not updating the image on Mac for some reason. Use
draw() instead.

draw_idle() should really work, and it gives a nice speed improvement, but I'm
not sure what the problem is. We can go with `draw()` for now.


bnmajor (approved):
LGTM

PR #924, opened 2021-05-12
Get pdb to interpret arrow keys and tab correctly

psavery:
Because we are now re-directing stdout and stderr, pdb can't
interpret things like arrow keys and the tab key correctly.

To fix this, get pdb to always use the default stdout and stderr
when it is running.


bnmajor (approved):
LGTM, works well!

PR #936, opened 2021-05-25
Set panel buffer

### LINKED ISSUE #932: Update panel buffers for PXRDIP workflow
The panel buffers should be set after image crop:
- inside = true, outside = false

Most important is to have this working with the `TARDIS` instrument at minimum

bnmajor:
Fixes #932 

@joelvbernier The only change that this makes is saving the panel buffer in the config from the mask that is created in the LLNL Import Panel - the image is still being masked. If that is indeed the correct way to handle the masking then this should be ready to go

### DIFF
--- a/hexrd/ui/import_data_panel.py
+++ b/hexrd/ui/import_data_panel.py
@@ -362,11 +362,12 @@ class ImportDataPanel(QObject):
             width, height = height, width
         self.it.cropped_image(height, width)
 
-        img = self.it.masked_image
+        img, panel_buffer = self.it.masked_image
 
         self.edited_images[self.detector] = {
             'img': img,
-            'tilt': self.it.rotation
+            'tilt': self.it.rotation,
+            'panel_buffer': panel_buffer,
         }
 
         self.it.completed()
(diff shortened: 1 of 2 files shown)


joelvbernier (approved):
@bnmajor -- how it the code dealing with boundaries that have NaN in then, as far as calculating a mask is concerned?

bnmajor:
@joelvbernier Using numpy's `logical_xor`: https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/interactive_template.py#L161

joelvbernier:
This is really only an issue with the Tardis boundary with the holes so far...  assuming the results of `polygon` are setting the pixels _inside_ the hole to `True`, the XOR should work.  Can you attach an image of the masks to this thread?

bnmajor:
@joelvbernier That is how it is currently working for TARDIS since the outermost boundary is stored first in the files.

What the unmasked and masked images look like:
unmasked_tardis
masked_tardis

Attached are those images (the original hdf5 files and the masked images) as well as the masks dumped to `.npz` files. Let me know if this does not seem correct to you.
[data.zip](https://github.com/HEXRD/hexrdgui/files/6556741/data.zip)

joelvbernier:
Looks correct to me!  Merging!

PR #939, opened 2021-05-26
Apply the same transform to all images for a single detector

### LINKED ISSUE #902: Reflecting transform in load panel
There is still an issue in the load panel when transforms are selected for images. If there are multiple images selected for a detector, the transforms chosen when confirming image/detector association are not reflected in the load panel.

The problem right now is that in the load panel the selected transform is applied to all images for that detector. In the load dialog however, the user is able to set the transform per image.

@joelvbernier How would we like to handle this? We could:

1. Assume that the user will be applying the same transform to all images for a detector and simply reflect the selection of the first image for each detector in the load panel.
2. Force the user to have the same transform for all images associated with a detector. In the load dialog changing one transform would change all that are associated with that detector.
3. Change the transform option in the load panel to read something like `multiple transforms` if multiple images are loaded per detector. It would essentially be temporarily changing the `None` text in this case. It would not be a selectable option in the menu and would be reset to `None` as soon as the images are loaded. Changing the transform in the load panel would keep the same behavior: the selection is a applied to all images.

bnmajor:
Fixes #902

### DIFF
--- a/hexrd/ui/load_images_dialog.py
+++ b/hexrd/ui/load_images_dialog.py
@@ -73,6 +73,8 @@ class LoadImagesDialog:
                 det_cb = QComboBox()
                 det_cb.addItems(list(set(self.detectors)))
                 table.setCellWidget(i, 0, det_cb)
+                table.cellWidget(i, 0).currentTextChanged.connect(
+                    lambda v, i=i: self.selection_changed(v, i))
             else:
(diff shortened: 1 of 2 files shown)


psavery (approved):
I don't fully understand everything going on, but I don't see any language issues.

LGTM

Issue #942, opened 2021-05-27
Sticky matplotlib tool state create problems for LLNL import tool


joelvbernier:
I was in a session where I had toggled the zoom-to-rectangle tool in the graphics canvas; when I brought up the LLNL import tool, the zoom behavior was still enabled when trying to drag the detector outlines, which is problematic (obvi).  Looks like we should reset the state of the zoom and pan tools in the matplotlib ToolManager when switching to the LLNL import tool.

PR #943, opened 2021-05-27
De-select toolbar options if toolbar is toggled off

bnmajor:
Make sure that zoom and pan have been toggled off before hiding the toolbar.

This showed up as a bug in the import panel (once the import was started the boundary could not be moved if pan or zoom had been left selected) but this could also cause confusion in other cases if the user toggles the toolbar off from the menu.

Always de-select options to be safe.


joelvbernier (approved):

bnmajor:
Fixes #942

PR #948, opened 2021-06-03
Downgrade scikit-learn to avoid Windows error

psavery:
Downgrade scikit-learn to 0.24.1 to avoid installation error during
`conda install anaconda` for the github actions packaging.

This error only happens for scikit-learn that is installed from
anaconda, and for scikit-learn 0.24.2, and for windows-latest on
GitHub actions.


bnmajor (approved):
LGTM, thanks Patrick!

PR #953, opened 2021-06-09
Save spots

### LINKED ISSUE #944: Add an option to the FitGrainsOptionsDialog to write out spots files
The user would choose a directory, and hexrd will automatically write the files to that directory.

I believe these are the changes required:

1. Setting [this option to `True`](https://github.com/HEXRD/hexrdgui/blob/2b92ca14b3cc954ceff7fc83a887795c9cbfa18e/hexrd/ui/indexing/run.py#L284) if the user wants to save the spots files
2. Setting up [the `analysis_dir`](https://github.com/HEXRD/hexrd/blob/e1f9b320d24b807ef7d6e1da70217244a6bf72c0/hexrd/config/root.py#L27-L29) on the config object to return the directory that the user selected. I **believe** that to do this, in the indexing config, we will need to set the [`working_dir` key](https://github.com/HEXRD/hexrdgui/blob/2b92ca14b3cc954ceff7fc83a887795c9cbfa18e/hexrd/ui/resources/indexing/default_indexing_config.yml#L7) to be the path to the parent directory of the directory the user selected, and set the [`analysis_name`](https://github.com/HEXRD/hexrdgui/blob/2b92ca14b3cc954ceff7fc83a887795c9cbfa18e/hexrd/ui/resources/indexing/default_indexing_config.yml#L1) to be the name of the directory the user selected. The defaults are in that yaml file linked (not setting `working_dir` looks like it defaults to the current working directory). The user would choose the directory in the [`FitGrainOptionsDialog`](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/indexing/fit_grains_options_dialog.py), and those keys linked earlier should be set on [this indexing config object in this function](https://github.com/HEXRD/hexrdgui/blob/2b92ca14b3cc954ceff7fc83a887795c9cbfa18e/hexrd/ui/indexing/fit_grains_options_dialog.py#L173) based upon what they chose. It should hopefully work, then. I think it would also be a good idea to save an underscore key to the indexing config such as `_write_spots` (similar to the underscore key being saved [here](https://github.com/HEXRD/hexrdgui/blob/2b92ca14b3cc954ceff7fc83a887795c9cbfa18e/hexrd/ui/indexing/fit_grains_options_dialog.py#L174)) that indicates whether or not to write the spots files. This can be checked [here](https://github.com/HEXRD/hexrdgui/blob/2b92ca14b3cc954ceff7fc83a887795c9cbfa18e/hexrd/ui/indexing/run.py#L284) when determining whether to write out the spots files or not.
3. The `indexing_config` object gets saved and loaded from the `QSettings`. So, the GUI should be updated with whatever is set in the `indexing_config` object around [here](https://github.com/HEXRD/hexrdgui/blob/2b92ca14b3cc954ceff7fc83a887795c9cbfa18e/hexrd/ui/indexing/fit_grains_options_dialog.py#L207). That way, the options the user previously chose will be the default the next time they run it, including whether or not to write out the spots files, and which directory to use.

bnmajor:
Fixes #944 

@joelvbernier This adds the option to write out the spots files by providing a checkbox and option to set the directory where they should be saved. Previous selections are remembered between uses.

### DIFF
--- a/hexrd/ui/indexing/fit_grains_options_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_options_dialog.py
@@ -175,11 +175,10 @@ class FitGrainsOptionsDialog(QObject):
         self.tolerances_model.copy_to_config(config)
 
         indexing_config = HexrdConfig().indexing_config
-        indexing_config['analysis_name'] = Path(self.working_dir).stem
-        indexing_config['working_dir'] = str(Path(self.working_dir).parent)
+        indexing_config['analysis_name'] = Path(self.spots_path).stem
+        indexing_config['working_dir'] = str(Path(self.spots_path).parent)
         indexing_config['_selected_material'] = self.selected_material
         indexing_config['_write_spots'] = self.ui.write_out_spots.isChecked()
-        indexing_config['_spots_path'] = self.spots_path
 
     def update_gui_from_config(self, config):
         blocked = [QSignalBlocker(x) for x in self.all_widgets()]
@@ -214,11 +213,11 @@ class FitGrainsOptionsDialog(QObject):
 
         indexing_config = HexrdConfig().indexing_config
         self.selected_material = indexing_config.get('_selected_material')
-        working_dir = indexing_config.get('working_dir', str(Path(HexrdConfig().working_dir).parent))
-        analysis_name = indexing_config.get('analysis_name', Path(HexrdConfig().working_dir).stem)
-        self.working_dir = str(Path(working_dir) / analysis_name)
-        self.spots_path = indexing_config.get(
-            '_spots_path', HexrdConfig().working_dir)
+        working_dir = indexing_config.get(
+            'working_dir', str(Path(HexrdConfig().working_dir).parent))
+        analysis_name = indexing_config.get(
+            'analysis_name', Path(HexrdConfig().working_dir).stem)
+        self.spots_path = str(Path(working_dir) / analysis_name)
         write_spots = indexing_config.get('_write_spots', False)
         self.ui.write_out_spots.setChecked(write_spots)
 
--- a/hexrd/ui/indexing/fit_grains_options_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_options_dialog.py
@@ -213,7 +213,8 @@ class FitGrainsOptionsDialog(QObject):
 
         indexing_config = HexrdConfig().indexing_config
         self.selected_material = indexing_config.get('_selected_material')
-        self.working_dir = indexing_config.get('working_dir', HexrdConfig().working_dir)
+        self.working_dir = indexing_config.get(
+            'working_dir', HexrdConfig().working_dir)
         write_spots = indexing_config.get('_write_spots', False)
         self.ui.write_out_spots.setChecked(write_spots)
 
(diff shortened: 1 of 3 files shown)


hexrd/ui/indexing/fit_grains_options_dialog.py
```diff
@@ -271,3 +282,11 @@ def update_num_hkls(self):
         num_hkls = len(self.material.planeData.getHKLs())
         text = f'Number of hkls selected:  {num_hkls}'
         self.ui.num_hkls_selected.setText(text)
+
+    def set_working_dir(self):
+        caption = 'Select directory to write sports files to'
```
psavery, line 287: `sports` => `spots`

hexrd/ui/indexing/fit_grains_options_dialog.py
```diff
@@ -171,7 +175,10 @@ def update_config(self):
         self.tolerances_model.copy_to_config(config)
 
         indexing_config = HexrdConfig().indexing_config
+        indexing_config['analysis_name'] = self.working_dir
```
psavery, line 178: I haven't tried this, but I believe the `analysis_name` should be this instead: `str(Path(self.working_dir).stem)`.

As in, it should be the stem, not the full path.

hexrd/ui/indexing/fit_grains_options_dialog.py
```diff
@@ -206,6 +213,10 @@ def update_gui_from_config(self, config):
 
         indexing_config = HexrdConfig().indexing_config
         self.selected_material = indexing_config.get('_selected_material')
+        self.working_dir = indexing_config.get(
+            'working_dir', HexrdConfig().working_dir)
```
psavery, line 217: Because `indexing_config['working_dir']` should be the parent directory, and it appears that `self.working_dir` is supposed to be the full path, we should set `self.working_dir` similar to this instead:

```
working_dir = indexing_config.get('working_dir', str(Path(HexrdConfig().working_dir).parent))
analysis_name = indexing_config.get('analysis_name', str(Path(HexrdConfig().working_dir).stem))
self.working_dir = str(Path(working_dir) / analysis_name)
```

Maybe we should rename `self.working_dir` to something like `self.spots_path` to disambiguate the fact that it is not the same thing as `indexing_config['working_dir']`.

psavery (commented):
Let's fix the flake8 issues too

hexrd/ui/indexing/fit_grains_options_dialog.py
bnmajor: Oops - fixed, thanks.

hexrd/ui/indexing/fit_grains_options_dialog.py
bnmajor: Yes, that makes sense. Fixed.

hexrd/ui/indexing/fit_grains_options_dialog.py
bnmajor: Sounds good

bnmajor:
@psavery I've made the changes that you've suggested, so now when the user selects the directory to save the spots files to the full path is saved as `self.spots_path` (so that the user's selection is remembered), and `self.working_dir` is now `str(Path(self.spots_path).parent / self.working_dir.stem)`. So for example:

```
self.spots_path = '/home/brianna/hexrd/ge/spots'
self.working_dir = 'ge/spots'
indexing_config['analysis_name'] = 'spots'
indexing_config['working_dir'] = 'ge'
```
The only change that I made was not casting `Path().stem` to a string since `.stem` already returns a string (`.parent` does return  a `PosixPath` though)

joelvbernier (approved):

joelvbernier:
This is part of it, however, we will also want to generate the more detailed spots data table (output of `HEDMInstrument.pullspots()`  for user-selected grains.

psavery (approved):
LGTM. I tried it and it seems to work as it should.

Issue #935, opened 2021-05-19
reflection selection


saransh13:
The reflection selection menu in the main Materials panel doesn't behave as expected, with the text highlighting messed up. The cmd+A, shift etc keyboard options also don't seem to work. However, the reflection selector menu from inside the overlay manager seems to work as expected. Screenshot attached.

![Screen Shot 2021-05-19 at 12 00 12 PM](https://user-images.githubusercontent.com/15834451/118897781-d4008400-b8bf-11eb-8536-0b1ae9f39c44.png)

psavery:
Do you know if there was an exception raised in the materials panel or reflections table? I have noticed something like this before if an exception gets raised. If there is, we can fix that exception.

saransh13:
This was the exception.
```Traceback (most recent call last):
  File "/Users/saransh1/hexrdgui/hexrd/ui/materials_panel.py", line 231, in active_material_changed
    self.reflections_table.material = HexrdConfig().active_material
  File "/Users/saransh1/hexrdgui/hexrd/ui/reflections_table.py", line 57, in material
    self.update_table()
  File "/Users/saransh1/hexrdgui/hexrd/ui/reflections_table.py", line 173, in update_table
    self.update_selected_rows()
  File "/Users/saransh1/hexrdgui/hexrd/ui/reflections_table.py", line 71, in update_selected_rows
    self.selected_rows = self.map_selections_to_rows(self.selections)
  File "/Users/saransh1/hexrdgui/hexrd/ui/reflections_table.py", line 186, in map_selections_to_rows
    selected_hkls = [self.index_to_hkl_map[x] for x in selections]
  File "/Users/saransh1/hexrdgui/hexrd/ui/reflections_table.py", line 186, in <listcomp>
    selected_hkls = [self.index_to_hkl_map[x] for x in selections]
KeyError: 5```

psavery:
Yeah, I have encountered this too. I have to fix this.

psavery:
Bugs like this that put hexrdgui in an unusable state should be high priority.

It's not nice when this kind of bug happens, the GUI becomes unusable, and the only way to fix it is by closing the program.

This bug should be fixed by #956

PR #956, opened 2021-06-15
Add safety checks for dicts in reflections table

psavery:
For hexrd's PlaneData class, sometimes the exclusions list can be
longer than the hkl list that gets returned. I specifically noticed
this happening if tThMax gets reduced.

This was resulting in indexing errors in the reflections table, which
put the reflections table in a very unusable state.

To prevent this kind of thing from happening both now and in the future
(if other changes are made), add some safety checks for the dicts. This
will at least prevent the reflections table from becoming unusable.

The behavior also seems to be desirable, where even though the
exclusions are longer than the hkl list, the correct rows get selected
in the table.

Fixes: #935


bnmajor (approved):
LGTM

PR #957, opened 2021-06-15
Write grains.out file with spots files

psavery:
If the spots files were written, write out the grains.out file as well.


bnmajor (approved):
LGTM!

Issue #950, opened 2021-06-04
Apply lorentz polarization correction to images


saransh13:
This option will scale the pixels in the images based on the lorentz polarization factors. The computation of the polarization factor will be in the instrument class. The same interface as the "Apply Solid Angle Correction" can be used in the GUI. 

It might make sense to have a sub-option inside "Edit" menu option called "Corrections" where both these options could sit as well. Any future correction factors (such as sample view) can also be put in there.

psavery:
@saransh13 Sounds good.

I think I need to clean up some of the code in hexrdgui that does the corrections as well, to make it more standardized...

joelvbernier:
Agreed -- we can have a submenu for intensity corrections.

psavery:
@saransh13 Let me know when this is added to hexrd.

saransh13:
@psavery https://github.com/HEXRD/hexrd/pull/286 implements the functionality in hexrd. Note that the factors are not properties since it requires the user input of two numbers f_horizontal and f_vertical (they should add up to 1). These are the fraction of x-ray polarized in the horizontal and vertical directions.

psavery:
@saransh13 Sounds good to me!

PR #958, opened 2021-06-15
Standardize intensity corrections and add Lorentz polarization correction

psavery:
This standardizes intensity corrections by always performing them in `HexrdConfig`, rather
than duplicating the code in each of the views.

This also creates a new sub-menu under "Edit" called "Intensity Corrections", where pixel
solid angles were moved.

Additionally, the option to perform Lorentz polarization correction was added and placed
under "Intensity Corrections". This pops up a dialog and gets the user to select
horizontal and vertical fractions of polarization.

Fixes: #950


bnmajor (approved):
LGTM!

Issue #889, opened 2021-04-09
Look into writing GUI tests


psavery:
Here is a potentially helpful plugin for pytest: [pytest-qt](https://github.com/pytest-dev/pytest-qt)

And here's the example they give:

```python
def test_hello(qtbot):
    widget = HelloWidget()
    qtbot.addWidget(widget)

    # click in the Greet button and make sure it updates the appropriate label
    qtbot.mouseClick(widget.button_greet, QtCore.Qt.LeftButton)

    assert widget.greet_label.text() == "Hello!"
```

joelvbernier:
`qtbot` love it!

PR #961, opened 2021-06-18
Add basic testing infrastructure

psavery:
This includes a test that starts hexrdgui, loads an instrument config
file, and then loads some data.

This is a work in progress.

Fixes: #889


joelvbernier:
We should probably get this in place soon, but can defer until October...

psavery:
@joelvbernier Indeed. If this was merged, it would have actually caught the breakpoint in #1023.

psavery:
I think this can actually be ready for review and merging. Currently, all it does is load an instrument config file and load data, and confirm that these were successful. This can already catch some issues, though, such as the `breakpoint()` I accidentally left in #1023 (the tests failed with that change, and now they are passing again due to #1027).

cjh1 (approved):
LGTM

PR #962, opened 2021-06-18
Pin the hexrd version in conda

psavery:
This is to make sure the packaging uses the right version.

We'll have to remember to update this every time we are pushing a
new hexrdgui tag, or come with an easier way so we won't have to
remember...


psavery:
Looks like this fails because PRs use the hexrd pre-release channel, which doesn't have the hexrd releases on it (rather, it has the latest package on hexrd master).

joelvbernier (approved):

joelvbernier (approved):

PR #963, opened 2021-06-18
Always upload installer output from packaging

psavery:
This will help us figure out if the correct hexrd is being
installed...

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - importlib_resources
     - fabio
     - pyyaml
-    - hexrd=0.8.8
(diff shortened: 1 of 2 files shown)


psavery:
This was just for testing the CI. Closing...

psavery:
Just to leave a note here: there is unfortunately currently no way in CPack to show the install output when the install process succeeds. CPack doe snot make the output available in any way other than writing it to `InstallOutput.log` when an error occurs. Thus, we can only see the install output when an error occurs.

Maybe we should request CPack to make an option to always write out the `InstallOutput.log`, so that even if the install command succeeds, we can see the output and confirm that the correct versions of packages were installed.

PR #964, opened 2021-06-22
Fix hexrd version in packaging as well

psavery:
This also uses the exact constraint instead of the fuzzy constraint
for the version matching.

The hexrd version will be programmatically set in the future.


psavery:
@joelvbernier I confirmed that the packages produced by this PR are using the correct version of hexrd.

bnmajor (approved):

PR #965, opened 2021-06-23
Get packaging to always use the latest hexrd

psavery:
For release builds, this should use the latest hexrd release.

For pre-release builds, this should use the latest hexrd pre-release.

This constraint is necessary to get the conda packaging to always use
the latest hexrd. Otherwise, conda may choose older hexrd versions,
especially if older hexrd versions allow it to use a newer version of
a different package.


psavery:
Looks like this is working, but we'll wait to merge this until after hexrdgui 0.8.7 is released...

joelvbernier (approved):

Issue #952, opened 2021-06-09
FF-HEDM workflow enhancements


joelvbernier:
User feedback (thanks @minety, @georgejoe) has made it clear that we need a bit more feedback during the indexing and fitting processes.  Here is a running list (with some existing issues linked) encapsulating some of the highest priority needs:

- Indexing (`find-orientations`):
  - [x] Add ability to store/save the scored orientation list (like CLI); would also be nice to report stats on scores. (fixed by #966)
  - [x] Add ability to expose the quaternion grid search option (_i.e.,_ load a npy file of quaternions to test rather than using the seeded search) (fixed by #966)
  - [x] Report how many test orientations are generated in the seeded search process and issue a warning if that number exceeds 5e7 quaternions (fixed by #966)
  - [x] Report how many orientations are staged for clustering and issue a warning dialog if that number exceeds 1e6 (parameters are likely not ideal and this can overwhelm the memory in `dbscan`) (fixed by #966)
  - [x] Expose the ability to re-run the clustering step on scored orientations with new parameters, including ability to load a list of scored orientations (clobbering the results of `paintGrid`) (tracked in #968)
  - [ ] Add the ability to plot indexing results on the eta-ome maps (fixed by #1180)
- Fitting (`fit-grains`)
  - [x] Add ability to visualize spots to fit grains results dialog (tracked in #945)
  - [x] Add an option to the FitGrainsOptionsDialog to write out spots files (tracked in #944)
  - [x] Add spot tolerance viz to rotation series overlays (tracked in #386) to assist users in setting them

@psavery -- do we need to pull someone else in to help here?

joelvbernier:
@psavery -- added another functionality to the `find-orientations` enhancements above; should I convert that item to a separate issue?

psavery:
@joelvbernier No this is fine!

PR #966, opened 2021-06-23
FF HEDM Workflow Enhancements

psavery:
This adds several enhancements to the HEDM workflow that were mentioned in #952. Namely:

1. Added ability to save the scored orientations. This is currently in the bottom left corner of the eta omega maps viewer.
2. Allow loading a list of quaternions instead of performing a seed search.
2. Warn if more than 5e7 quaternions were generated in the seed search process, and ask if the user wants to proceed anyways. If they choose "no", the user will be taken back to the eta omega maps viewer so they can choose different settings, which will hopefully result in fewer quaternions.
3. Warn if more than 1e6 quaternions are staged for clustering, done in a similar way as 2.

Fixes several items in #952


psavery:
This PR now also allows loading a list of quaternions instead of performing a seed search. You can see how the eta omega maps viewer hides seed search settings when the quaternion grid search is being used:

grid_search

Here is with the seed search method enabled:

seed_search

joelvbernier (approved):

Issue #761, opened 2021-02-22
Save calibration picks in editable table.


joelvbernier:
This is a major request from the LLNL side.  We need to save all of the calibration pick points in an editable table.  We'll need slightly different layouts for the powder and Laue overlays.  We should save the picks in the raw pixels coordinates (_i.e._ push them through the current instrument's detector transforms to the image coords) so they can be reapplied in the polar view at updated angular positions (following updates to the detector transforms via the calibration process).

We'll also want the ability to export/import pick sets.

Eventually this will be part of a workflow...

joelvbernier:
This is the next highest-priority item in the LLNL task list @psavery -- adding some detail:
- [ ] we want to change the looping order for the powder diffraction picks (user request that I endorse) to
```
for ring in ring_set:
    for detector in detectors:
        _store_picks()
```
It is more intuitive to make all picks for a given ring over all detectors than what we currently have.
- [ ] we only need to store picks in the raw pixel coordinates so that the angles can be updated according to the current instrument parameters
- [ ] need I/O; HDF5 for similarity to other data saves?

psavery:
> * [x]  we only need to store picks in the raw pixel coordinates so that the angles can be updated according to the current instrument parameters (fixed by [Hand picked calibration improvements #1030](https://github.com/HEXRD/hexrdgui/pull/1030))

We ended up storing the picks both in angular coordinates and in warped Cartesian coordinates. When we import the picks, we convert from warped Cartesian coordinates into angular coordinates.

PR #967, opened 2021-06-24
Loop by rings first for powder calibration picker

psavery:
Previously, we were looping over detectors, then rings.

Now, for powder, we are looping over rings, then detectors.
Specifically, we are looping over the rings from min two theta to
max two theta, and then over the detectors from min eta to max eta.

Laue still loops over all spots in a detector first before moving
on to the next detector.

This fixes one of the items in #761


joelvbernier:
Oooh, good one.

joelvbernier (approved):

PR #969, opened 2021-06-25
Always use the associated dark subtraction file if it exists

bnmajor:


psavery (approved):
LGTM

PR #972, opened 2021-06-30
Add a "Copy" button to the messages widget

psavery:
This copies the text in the messages widget to the user's clipboard.

It is there for convenience, so that the user doesn't have to highlight
the text before copying.


bnmajor (approved):
Very cool! LGTM

Issue #886, opened 2021-04-06
Add ability to add charge state when selecting atoms from periodic table


saransh13:
In the Materials structure tab, user should have the ability to choose the charge state of an atom. The charge state changes the form factors, which can affect the quality of the fits when using whole powder pattern fitting techniques. 

The table of possible charge state for each element is tabulated in `hexrd.constants.chargestate` as a dictionary. The charge state will be or the form `0`, `n+` or `n-` for the case of neutral, cations and anions respectively. The form factor for an element, `X` can be queried using the `hexrd.constants.scatfac` dictionary and the key `X`, `"Xn-"` or `"Xn+"`

PR #973, opened 2021-07-01
Add ability to modify a material's charge

psavery:
Note that we have to update the unit cell and structure factor
manually [since the material is not currently doing that](https://github.com/HEXRD/hexrd/blob/300f43ffe35cdbd0a7827b80d0b9209f196771a0/hexrd/material.py#L864-L866).

Maybe we should just add the charge as an argument to
[`set_atomdata()`](https://github.com/HEXRD/hexrd/blob/300f43ffe35cdbd0a7827b80d0b9209f196771a0/hexrd/material.py#L942)?

Fixes: #886

Depends on: hexrd/hexrd#301

### DIFF
--- a/hexrd/ui/material_structure_editor.py
+++ b/hexrd/ui/material_structure_editor.py
@@ -253,13 +253,7 @@ class MaterialStructureEditor(QObject):
                 U_array[i] = scalar_to_tensor(U)
 
         mat = self.material
-        mat._set_atomdata(type_array, info_array, U_array)
-        mat.charge = charge_array
-
-        # Must force an update of the unit cell and structure factor
-        # for the change in the charge array.
-        mat._newUnitcell()
-        mat.update_structure_factor()
+        mat._set_atomdata(type_array, info_array, U_array, charge_array)
 
         self.material_modified.emit()
 
--- a/hexrd/ui/material_structure_editor.py
+++ b/hexrd/ui/material_structure_editor.py
@@ -23,11 +23,13 @@ DEFAULT_SITE = {
     'atoms': [
         {
             'symbol': 'O',
+            'charge': '0',
             'occupancy': 0.5,
             'U': 1e-6
         },
         {
             'symbol': 'Ce',
+            'charge': '0',
             'occupancy': 0.5,
             'U': 1e-6
         }
@@ -234,6 +236,7 @@ class MaterialStructureEditor(QObject):
         # Convert the sites back to the material data format
         info_array = []
         type_array = []
+        charge_array = []
         U_array = []
 
         for site in self.sites:
@@ -241,6 +244,7 @@ class MaterialStructureEditor(QObject):
                 info_array.append((*site['fractional_coords'],
                                    atom['occupancy']))
                 type_array.append(ptable[atom['symbol']])
+                charge_array.append(atom['charge'])
                 U_array.append(atom['U'])
 
         if any(isinstance(x, np.ndarray) for x in U_array):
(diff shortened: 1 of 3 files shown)


hexrd/ui/material_structure_editor.py
```diff
@@ -249,17 +250,25 @@ def update_material(self):
...
                 U_array.append(atom['U'])
 
         mat = self.material
         mat._set_atomdata(type_array, info_array, U_array)
+        mat.charge = charge_array
+
+        # Must force an update of the unit cell and structure factor
+        # for the change in the charge array.
+        mat._newUnitcell()
+        mat.update_structure_factor()
```
psavery, line 271: Here is where we are updating the unit cell and structure factor manually.

joelvbernier:
Where is this one at @saransh13 and @psavery ?

psavery:
> Where is this one at @saransh13 and @psavery ?

Ready for @saransh13 to review.

saransh13 (commented):
The following error is generated when I try to update the atomic species. For this example, I switched Nickel to Cobalt and got this error.

I have also pushed an update to `_set_atomdata`[https://github.com/HEXRD/hexrd/blob/1c25c6e70695010c8f1e55b5819a3d716ac44259/hexrd/material.py#L967] which takes the charge as an input. The corresponding changes to the guidelines should be here[https://github.com/HEXRD/hexrdgui/blob/5149e8eb7c059ff9876a41fa1b18caa80601e8f0/hexrd/ui/material_structure_editor.py#L262]. I tried to directly push the update to your fork, but I don't think I have the right permissions.

Screen Shot 2021-07-23 at 12 28 11 PM

psavery:
> I have also pushed an update to [`_set_atomdata`](https://github.com/HEXRD/hexrd/blob/1c25c6e70695010c8f1e55b5819a3d716ac44259/hexrd/material.py#L967) which takes the charge as an input. The corresponding changes to the `hexrdgui` should be [here](https://github.com/HEXRD/hexrdgui/blob/5149e8eb7c059ff9876a41fa1b18caa80601e8f0/hexrd/ui/material_structure_editor.py#L262). I tried to directly push the update to your fork, but I don't think I have the right permissions.

It looks like [`_set_atomdata`](https://github.com/HEXRD/hexrd/blob/bf0d30c57c7497f99b8d8402a5261f104fa17e29/hexrd/material.py#L967) isn't updated yet on hexrd master with those changes. Can you put up a PR for it?

psavery:
> The following error is generated when I try to update the atomic species. For this example, I switched Nickel to Cobalt and got this error.

I just fixed this. Thank you for finding it.

saransh13 (approved):
LGTM. The new PR for updating charge with other artomdata is up, so this could be updated to use that before merging. But otherwise this works really well.

hexrd/ui/material_structure_editor.py
psavery: This code is now fixed to pass the charge via the changes in hexrd/hexrd#301.

joelvbernier:
Should we merge HEXRD/hexrd/pull/301 and make the required changes here?  I can tag hexrd in that case.

psavery:
> Should we merge [HEXRD/hexrd/pull/301](https://github.com/HEXRD/hexrd/pull/301) and make the required changes here? I can tag hexrd in that case.

The required changes are actually already in place in this PR! We just need to merge hexrd/hexrd#301 first, and then we can merge this one.

PR #974, opened 2021-07-01
Add a thermal factor editor to allow tensor U

psavery:
This allows the thermal factor (U/B) to be either a tensor or a scalar.

As a scalar, the user can modify U in its spinbox as normal. If the
user wishes to convert it to a tensor, they must double-click the
spinbox, and then the thermal factor editor will appear, allowing
them to convert it into a tensor.

If U is a tensor, the spinbox text will just display "Tensor", and
the user must click on it in order to bring up the thermal factor
editor, in order to edit it.

This was tested via round-tripping (creating a tensor U, writing
it out to a materials file, resetting the application, and then
loading the materials file again, to ensure that the tensor U
was saved and loaded properly).

https://user-images.githubusercontent.com/9558430/124174052-7c823600-da71-11eb-920e-9e49100da719.mp4

Fixes: #513

### DIFF
--- a/hexrd/ui/material_properties_editor.py
+++ b/hexrd/ui/material_properties_editor.py
@@ -9,7 +9,7 @@ from hexrd.unitcell import _StiffnessDict
 from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.matrix_editor import MatrixEditor
 from hexrd.ui.ui_loader import UiLoader
-from hexrd.ui.utils import compose
+from hexrd.ui.utils import apply_symmetric_constraint, compose
 
 
 class MaterialPropertiesEditor:
@@ -132,12 +132,3 @@ class MaterialPropertiesEditor:
             self.ui.volume,
             self.ui.volume_per_atom,
         ]
-
-
-def apply_symmetric_constraint(x):
-    # Copy values from upper triangle to lower triangle.
-    # Only works for square matrices.
-    for i in range(x.shape[0]):
-        for j in range(i):
-            x[i, j] = x[j, i]
-    return x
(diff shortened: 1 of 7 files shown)


joelvbernier:
@saransh13 have you tested/reviewed this one?

saransh13 (commented):
The code works well for me. The only suggestion I have is that the conversion from scalar --> tensor and back to scalar is non-intuitive. The option to switch back is inside the box to fill in the values. Is it possible to add the tick box for tensor or scalar value in the main structure window (in place of convert to tensor) and then the box for the values displays the tensor or scalar box depending on what box is ticked?
Screen Shot 2021-07-23 at 12 16 04 PM

psavery:
> The code works well for me. The only suggestion I have is that the conversion from scalar --> tensor and back to scalar is non-intuitive. The option to switch back is inside the box to fill in the values. Is it possible to add the tick box for tensor or scalar value in the main structure window (in place of convert to tensor) and then the box for the values displays the tensor or scalar box depending on what box is ticked?
> Screen Shot 2021-07-23 at 12 16 04 PM

Sure, I can change that. 

Note: in the GUI, not all of the atoms have to have the same type of U. Some can be scalar, and some can be tensor. This makes viewing and editing scalars easier if there are tensors present. I'll make toggling the checkbox coerce all of the types to one or the other, but the user will still have the ability to change individual one types by double-clicking the boxes.

psavery:
@saransh13 Okay, I've updated the code, and I think it is more intuitive to use now. Can you give it a try and let me know what you think?

saransh13 (approved):
LGTM now.

PR #977, opened 2021-07-08
Add a column for hkl ids in the reflections table

psavery:
This also removes the vertical header (which displayed indices)
since that does not seem necessary with the hkl ids.

This also resizes the reflections table a little bit.

This also gets the columns to automatically resize based upon
the contents of the table.

image


bnmajor (approved):
LGTM!

PR #984, opened 2021-07-13
Move "Subtract Minimum" to an intensity correction

### LINKED ISSUE #979: The "Subtract minimum" setting is being saved and always applied to future image loading
The average aggregation in the GUI is giving different intensity results than the script does for the averaging. A specific case where this can be seen is in the MEC example where the "stripes" should be zero and are currently not.

psavery:
This removes the "Subtract Minimum" option from the transform dialog,
and moves it into the "Edit"->"Intensity Corrections" menu.

This means that the internal re-workings of "Subtract Minimum" have
changed as well. It used to be an operation in a `ProcessedImageSeries`,
but now it is only applied to the image the user is currently viewing,
just like the other intensity corrections do.

We may want to consider changing the intensity corrections to also use
operations in `ProcessedImageSeries` so the intensity corrections will
be applied to the whole image series if it gets passed around. Currently,
if the whole image series gets passed somewhere (like the HEDM workflow),
none of the intensity corrections will be applied.

These re-workings have also fixed #979, since "Subtract Minimum" is now
using the intensity correction framework, rather than its own framework.

Fixes: #979

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -456,6 +456,11 @@ class HexrdConfig(QObject, metaclass=QSingleton):
                 factor = panel.lorentz_polarization_factor(**kwargs)
                 images_dict[name] = img / factor
 
+        if HexrdConfig().intensity_subtract_minimum:
+            minimum = min([np.nanmin(x) for x in images_dict.values()])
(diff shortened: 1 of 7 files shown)


bnmajor (approved):
LGTM

Issue #986, opened 2021-07-15
Masking is slow


bnmajor:
- [x] Slow performance for rectangle/ellipse masking in polar view (fixed by #988)
- [x] Toggling mask visibility is slow (fixed by #988)

Issue #750, opened 2021-02-12
Polar view SNIP fixes


psavery:
- [x] Make 'Show SNIP' behavior match applying the SNIP (fixed by #988)
- [ ] Export SNIP background when exporting the polar image
- [ ] Export polar image without masks applied, and export the masks separately

Currently, the 'Show SNIP' behavior doesn't match the 'Apply SNIP?' behavior when 'Apply SNIP?' is unchecked. 'Apply SNIP?' will apply the FAST_SNIP_1D with zeros instead of NaNs. Also, the snip algorithm is ran before the user-defined masks are applied for `Apply SNIP?`, but the snip algorithm is ran after the user-defined masks are applied for `Show SNIP` without `Apply SNIP?`.

We should make the behaviors match.

joelvbernier:
Regarding the user-defined masks; exclude those from the SNIP calculations.  The user-defined masks are often used to remove artifacts introduced at the edges by the SNIP convolution.  Also, FAST_SNIP_1D cannot handle NaNs, which necessitates the fill with zeros.

psavery:
We might need to keep a deep copy around of the un-modified polar image so that the `Show SNIP` can access it. If you do this, feel free to put it in the [PolarView](https://github.com/HEXRD/hexrdgui/blob/36831d1fc8360ee26dae12b6cd8199b0356817f2/hexrd/ui/calibration/polarview.py#L28) class and set it in the `PolarPlot class somewhere around [here](https://github.com/HEXRD/hexrdgui/blob/36831d1fc8360ee26dae12b6cd8199b0356817f2/hexrd/ui/calibration/polar_plot.py#L44). Then, it can be accessed [here](https://github.com/HEXRD/hexrdgui/blob/36831d1fc8360ee26dae12b6cd8199b0356817f2/hexrd/ui/image_canvas.py#L790) for the `Show SNIP`.

I am also happy to work on this if you would like. Just let me know who should work on it.

PR #988, opened 2021-07-22
Masking performance improvements

psavery:
This adds several improvements to masking, primarily performance improvements.

Applying polar masks, and toggling their visibility, is now much faster, since the snipped image is cached and only the masking now changes:

https://user-images.githubusercontent.com/9558430/126709433-954befb3-97bb-48dd-8563-0e55ab7ac6df.mp4

Drawing masks via rectangles and ellipses is now animated via matplotlib's animation and blit features, and is much faster as well:

https://user-images.githubusercontent.com/9558430/126709481-9e19a4fd-8a03-4919-9995-f9e9500f038b.mp4

Fixes: #986
Also fixes the first item in #750

### DIFF
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -35,12 +35,18 @@ class InstrumentViewer:
     def angular_grid(self):
         return self.pv.angular_grid
 
+    @property
+    def img(self):
+        return self.pv.img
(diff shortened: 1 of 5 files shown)


bnmajor (approved):
LGTM - I can see some huge speedups in my testing especially for higher resolutions, thanks!

Issue #990, opened 2021-07-22
Look into polar view binary erosion to eliminate edge artifacts


psavery:
Look into polar view binary erosion from the outside of the image plate extent (nan region) to get rid of the edge artifacts.

PR #992, opened 2021-07-26
Add polar binary erosion to remove edge artifacts

psavery:
This adds an "Apply erosion?" checkbox to the polar tab of the
image mode widget, which can only be checked if "Apply SNIP?" is
checked.

If checked, a binary erosion is applied to the panel buffer mask, and
it uses 2x the snip width for the structure. This sets edge artifacts
to zero.

Here is an example:

No erosion
=========
no_erosion

With erosion
==========
with_erosion

Fixes: #990

### DIFF
--- a/hexrd/ui/calibration/polarview.py
+++ b/hexrd/ui/calibration/polarview.py
@@ -1,6 +1,8 @@
 import numpy as np
 
 from skimage.exposure import rescale_intensity
+from skimage.filters.edges import binary_erosion
+from skimage.morphology import disk
 
 from hexrd.transforms.xfcapi import detectorXYToGvec, mapAngle
 
@@ -8,7 +10,7 @@ from hexrd import constants as ct
 from hexrd.xrdutil import _project_on_detector_plane
 
 from hexrd.ui.hexrd_config import HexrdConfig
-from hexrd.ui.utils import SnipAlgorithmType, run_snip1d
+from hexrd.ui.utils import SnipAlgorithmType, run_snip1d, snip_width_pixels
 
 tvec_c = ct.zeros_3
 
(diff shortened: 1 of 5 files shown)


hexrd/ui/calibration/polarview.py
```diff
@@ -286,6 +288,12 @@ def apply_snip(self, img):
             self.snip1d_background = run_snip1d(img)
             # Perform the background subtraction
             img -= self.snip1d_background
+
+            if HexrdConfig().polar_apply_erosion:
+                mask = binary_erosion(
+                    ~self.raw_img.mask, structure=disk(2 * snip_width_pixels())
+                )
+                img[~mask] = 0
```
psavery, line 296: We might want to save the erosion mask out when we export the polar plot as well I guess, right @joelvbernier?

joelvbernier: Here's what my testing resulted in.  Starting at your line 292:
```
            if HexrdConfig().polar_apply_erosion:
                niter = HexrdConfig().polar_snip1d_numiter
                mask = binary_erosion(
                    ~self.raw_img.mask,
                    structure=rectangle(1, 2.1*niter*snip_width_pixels())
                )
                img[~mask] = 0
```

and just import `rectangle` in lieu of `disk`.

psavery: Looks like the arguments for `rectangle` need to be integers, so I will cast the second argument to an int like so:
```python
int(2.1*niter*snip_width_pixels())
```

psavery: This has been fixed.

psavery:
If we find out in the future that we need to allow the user to specify the rectangle `ncols`, we can find a place to allow them to specify it. Right now, the polar tab in the image mode widget is full, but we can be creative about moving things around.

joelvbernier (approved):

Issue #985, opened 2021-07-14
Add ability to specify 2𝜃 range for WPPF independent of polar view range


joelvbernier:
It is desireable to be able to specify (more specifically, to restrict) the 2𝜃 range that WPPF will perform the fit over instead of fixing it to the extent of the polarview image.  This will allow for the exclusion of problematic boundary effects when the extent of the polar view exceeds the detector extent(s) (see below).

Should be straightforward to do...

![image](https://user-images.githubusercontent.com/1154130/125701335-5f7464bd-77e8-4cf8-b8c9-8f85bf699e4e.png)

psavery:
@saransh13 Is there currently a way to restrict the WPPF two-theta range in hexrd? Maybe one place to implement it would be [in this property?](https://github.com/HEXRD/hexrd/blob/5e1385d163180a323d897d4e150ee8e3e4bfa4c2/hexrd/wppf/WPPF.py#L781-L786).

joelvbernier:
bumping up the priority on this one @saransh13 -- should be easy to implement, methinks.

saransh13:
So WPPF takes in a numpy array as input. If the input data is just cropped
while passing, WPPF should take care of the rest.

PR #997, opened 2021-07-28
Add ability to specify two theta limits for WPPF

psavery:
If the user checks the box to limit two theta, then the azimuthal
integral data that gets passed to WPPF will be limited to the
specified two theta range. WPPF will then operate on this limited
range.

Here's a (non-realistic) example that shows that it works:
example

Fixes: #985


joelvbernier:
Awesome!  LGTM.

bnmajor (approved):

Issue #894, opened 2021-04-20
Provide tables for elemental X-ray energy lookup 


joelvbernier:
Often the beam energy in an instrument results from a characteristic emission process.  For example, in the laser-driven shots we field on NIF and Omega, the X-rays are generated by ionizing an elemental foil, which [give characteristic He-&alpha; emission](https://xdb.lbl.gov/Section1/Sec_1-8.pdf).  Lab sources give [characteristic K-&alpha; emission](https://xdb.lbl.gov/Section1/Table_1-2.pdf) according to the anode material, and often the K-edge of a metal foil is used to tune/monitor monochromated synchrotron X-rays.

I am adding a database of the relevant tables to `hexrd/resources` in HDF5 format.

joelvbernier:
The table has the following format:
![image](https://user-images.githubusercontent.com/1154130/115450722-23746700-a1d1-11eb-8369-cba93c2f61e5.png)

joelvbernier:
@psavery -- what do you think, a menu item for optionally setting the instrument wavelength?  Or a button that fires up a widget in the instrument form view?

joelvbernier:
Depends on  https://github.com/HEXRD/hexrd/pull/256

joelvbernier:
@psavery -- the required hexrd dep was already merged for this one, so we would just need a new widget to select X-ray energy from [this HDF5 file](https://github.com/HEXRD/hexrd/blob/master/hexrd/resources/characteristic_xray_energies.h5) (including the possibility to average two entries).

psavery:
@joelvbernier Is this LLNL or HEDM, or either?

saransh13:
This should be LLNL.

PR #998, opened 2021-07-29
Add X-Ray Energy Selection via a Table

psavery:
This adds a button to the form view which allows the user to pick an
energy (or multiple energies to average) from a table.

The `characteristic_xray_energies.h5` from hexrd is being used to
generate this table.

The table is currently sorted so that atomic symbols come first,
and are sorted in alphabetical order, followed by atomic numbers.

https://user-images.githubusercontent.com/9558430/127437453-82adc761-239e-4c36-b7bc-e62635d6b73a.mp4

Fixes: #894

### DIFF
--- a/hexrd/ui/calibration_config_widget.py
+++ b/hexrd/ui/calibration_config_widget.py
@@ -7,9 +7,10 @@ from PySide2.QtWidgets import (
 import numpy as np
 
 from hexrd.ui import constants
-from hexrd.ui.hexrd_config import HexrdConfig
-from hexrd.ui.ui_loader import UiLoader
 from hexrd.ui.calibration.panel_buffer_dialog import PanelBufferDialog
+from hexrd.ui.hexrd_config import HexrdConfig
+from hexrd.ui.xray_energy_selection_dialog import XRayEnergySelectionDialog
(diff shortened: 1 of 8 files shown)


joelvbernier:
One comment:  the table values are all in eV, and we input energies in keV, so we'll need to scale the imported numbers by 1e-3 :-)

hexrd/ui/calibration_config_widget.py
```diff
@@ -92,6 +95,14 @@ def on_energy_wavelength_changed(self):
         finally:
             self.ui.cal_energy.blockSignals(block_signals)
 
+    def open_xray_energies_dialog(self):
+        dialog = XRayEnergySelectionDialog(self.ui)
+        if not dialog.exec_():
+            return
+
+        # The table has units in eV. Convert to keV.
+        self.ui.cal_energy.setValue(dialog.selected_energy / 1e3)
```
psavery, line 104: > One comment: the table values are all in eV, and we input energies in keV, so we'll need to scale the imported numbers by 1e-3 :-)

@joelvbernier that is actually done here!

joelvbernier (approved):

Issue #999, opened 2021-07-29
show SNIP button only works once


saransh13:
When I apply snip and then use the show SNIP button to display the snip, it works the first time. After closing that window, the show SNIP button does not do anything.

psavery:
This is actually a duplicate of #522, but that was closed without being fixed. I believe #1000 should fix this issue.

Issue #522, opened 2020-09-16
macOS only: snip1d popup only displays once


johnkit:
In the polar image mode, the "Show snip1d" button generates and displays a popup window. On macOS, if you close that popup, clicking the "Show snip1d" button again does not re-display the window. Internally I have confirmed that the figure is generated and the matplotlib figure.show() function is called.

psavery:
Possibly related: #204

johnkit:
Yes same thing. Thanks for pointing that out, Patrick.

Issue #204, opened 2019-10-28
"Show snip1d" only allows custom colormap on first button press


psavery:
If you press the button a second time, it always uses the default colormap, and the user cannot customize it.

PR #1000, opened 2021-07-29
Snip preview fixes

psavery:
1. Set window title for snip background preview

This was previously being set to "Figure 1".

2. Re-create SNIP preview plot each time

Re-using the figure and the axis seemed to be a reasonable thing to
do. However, it is causing issues for when the user re-shows it. 
On Linux, the user can't change the color map the second time they
show it. On Mac, it doesn't re-appear at all.

Instead of re-using, we will now re-create the figure from scratch each
time. It is not very resource intensive to do this, and figuring out 
why we can't re-use the figure and axis do not seem worth the effort.

Fixes: #204
Fixes: #522
Fixes: #999


bnmajor (approved):
LGTM!

PR #1006, opened 2021-08-05
Use conda-forge instead of anaconda channels

psavery:
This is an attempt to use the `conda-forge` channel instead of the
`defaults` and `anaconda` channels, for all packages.

This will hopefully fix the issue where we had to downgrade scikit-learn
because it would fail to install on Github Actions on Windows. It makes
more sense for us to install these packages from conda-forge since the
respective project developers actually maintain the packages on
conda-forge (whereas their packages on the anaconda/defaults channels
are maintained by anaconda).

We will first need to make sure the packaging and tests pass, and then
move on to test the packages manually.

If we run into crashing issues on Mac with openblas, as we have in the
past, we may be able to resolve it by setting the environment variable
`OPENBLAS_NUM_THREADS=1`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,7 +127,7 @@ jobs:
       working-directory: hexrdgui/packaging
       run: |
           conda activate hexrdgui-package
-          conda install anaconda scikit-learn==0.24.1
+          conda install --override-channels -c conda-forge anaconda-client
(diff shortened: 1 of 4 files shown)


psavery:
Debug output for Mac shows that `conda-pack` is coming from the `cjh1` channel, which is desired.

cjh1 (approved):
LGTM

Issue #1004, opened 2021-07-31
Ignore enter key for OmeMapsViewerDialog


psavery:
When the eta omega maps viewer appears, we need to ensure that if the user accidentally presses "Enter", it doesn't automatically execute "OK" or "Cancel". In fact, we need to make sure this gets fixed for all dialogs.

This should be taken care of automatically [here](https://github.com/HEXRD/hexrdgui/blob/7d79196c7e0c63dc5b83b50806122a08d6cfa3dc/hexrd/ui/ui_loader.py#L51), but it apparently is not. I confirmed that I can close the eta omega maps viewer immediately after it displays by pressing the "Enter" key...

Reported by @joelvbernier and @saransh13

psavery:
It looks like if `QDialog` receives the enter key press event, it successfully blocks the event (it normally forwards the enter key press event to it's default `QPushButton`, which is often in a `QDialogButtonBox`).

However, it looks like in some cases, such as in the `OmeMapsViewerDialog` here, that the `QPushButton` itself is receiving the enter key press event, thus bypassing the blocking that we have set up on the `QDialog`...

PR #1007, opened 2021-08-10
Block enter key press accept/reject for dialogs

psavery:
Previously, enter key presses would be blocked if the QDialog
received them. However, sometimes, the QPushButtons in the
QDialogButtonBox in the QDialog would receive the key press event
instead, thus bypassing our event filter. This would result in
the dialog being accepted/rejected accidentally, upon accidental
enter key press, and thus the issue we tried to prevent would not
always be fixed.

Now, the key event filter is installed on both the QDialogs and
on QPushButtons that follow this parent/child scheme:

QDialog -> QDialogButtonBox -> QPushButton

The event filter must be installed both on the QDialog and the
QPushButtons because it is currently unpredictable (to me,
at least) which one will receive the key press event, and if the
QDialog receives the key press event, it DOES NOT forward the
event to the QPushButton, but rather "clicks" the QPushButton.

This appears to fix all cases that I found.

Fixes: #1004

### DIFF
--- a/hexrd/ui/__init__.py
+++ b/hexrd/ui/__init__.py
@@ -1,24 +1,34 @@
 from PySide2.QtCore import QEvent, QObject, Qt
-from PySide2.QtWidgets import QDialog
+from PySide2.QtWidgets import QDialog, QPushButton
 
 
 class EnterKeyFilter(QObject):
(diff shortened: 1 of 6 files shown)


cjh1 (approved):
LGTM

PR #1011, opened 2021-08-16
Add additional validation for quaternions file

psavery:
This gives the user much less cryptic errors when they load a
quaternions file that has the wrong shape.


bnmajor (approved):
LGTM

Issue #1013, opened 2021-08-17
Mask manager 


saransh13:
Toggling the visible button in mask manager does not behave as expected. All the polar masks somehow only turn a single mask on/off. Remove mask button also has the same behavior. Video below.

https://user-images.githubusercontent.com/15834451/129782432-c0ea4bf8-aafe-4281-b9d3-5bb5d30e444d.mov

Issue #1012, opened 2021-08-17
Remove mask gives error


saransh13:
After loading an image and drawing masks in polar view, the remove mask option in Edit > Manage masks works fine. However, if a new image is loaded after removing all the old masks and adding new ones, the the remove mask button produces the following error

![Screen Shot 2021-08-10 at 11 34 45 AM](https://user-images.githubusercontent.com/15834451/129654445-bb223b27-9bef-4d47-99e8-f3e9cd76a9c5.png)

bnmajor:
@saransh13 Could you provide the exact steps that you took when you saw this error? So far I have been unable to reproduce this.

saransh13:
Ok trying to create a gif/video with this. I will upload shortly. Sometimes, it takes a few rounds of image reloads to reproduce this error.

bnmajor:
Thanks @saransh13! So there is an assumption being made for both toggled and removed masks that when the widget is clicked the `currentRow` is updated accordingly, and this value is what is used to determine which mask to hide/show/remove. This behavior is not guaranteed though, and I think that is what is causing this bug. I still have not been able to reproduce this locally but I think I have a solution and I will tag you in the PR when I have that up.

To note: At this point I believe that this and #1013 are caused by the same bad assumption.

PR #1015, opened 2021-08-18
Mask manager

bnmajor:
This branch fixes issues and bugs in the mask manager:

- Originally it was assumed that selecting a cell widget would also set the `currentRow` on the table, and that value is what was used to determine which mask had been selected. This is not guaranteed though and while it seemed to work on Linux machines it failed to work properly on Mac machines and the wrong mask was being affected. The name of the mask associated with the selected widget is now emitted with the signal so that the correct mask is always selected.
- Any existing masks are now removed when the configuration or detectors change. This was not the case before and this would create issues when trying to load a new configuration.
- Previously the user was able to use the context menu to request to export a mask that didn't exist and this would produce errors. The context menu is now only created when the user selects a valid row.

Fixes: #1012 
Fixes: #1013

### DIFF
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -100,14 +100,13 @@ class MaskManagerDialog(QObject):
             cb.setChecked(status)
             cb.setStyleSheet('margin-left:50%; margin-right:50%;')
             self.ui.masks_table.setCellWidget(i, 1, cb)
-            self.ui.masks_table.cellWidget(i, 1).toggled.connect(
+            cb.toggled.connect(
                 lambda checked, key=key: self.toggle_visibility(checked, key))
 
             # Add push button to remove mask
             pb = QPushButton('Remove Mask')
             self.ui.masks_table.setCellWidget(i, 2, pb)
-            self.ui.masks_table.cellWidget(i, 2).released.connect(
-                lambda i=i, key=key: self.remove_mask(i, key))
+            pb.clicked.connect(lambda i=i, key=key: self.remove_mask(i, key))
 
             # Connect manager to raw image mode tab settings
             # for threshold mask
@@ -120,7 +119,7 @@ class MaskManagerDialog(QObject):
     def setup_threshold_connections(self, checkbox, row, name):
         HexrdConfig().mode_threshold_mask_changed.connect(checkbox.setChecked)
         checkbox.toggled.connect(self.threshold_toggled)
-        self.ui.masks_table.cellWidget(row, 2).released.connect(
+        self.ui.masks_table.cellWidget(row, 2).clicked.connect(
             lambda row=row, name=name: self.remove_mask(row, name))
 
 
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -84,6 +84,7 @@ class MaskManagerDialog(QObject):
         self.ui.export_masks.clicked.connect(self.export_visible_masks)
         HexrdConfig().mode_threshold_mask_changed.connect(
             self.update_masks_list)
+        HexrdConfig().detectors_changed.connect(self.clear_masks)
 
     def setup_table(self, status=True):
         self.ui.masks_table.blockSignals(True)
@@ -233,3 +234,10 @@ class MaskManagerDialog(QObject):
             mtype, data = self.masks[mask]
             d[mask] = data
         self.export_masks(d)
+
+    def clear_masks(self):
+        HexrdConfig().polar_masks.clear()
+        HexrdConfig().raw_masks.clear()
+        HexrdConfig().visible_masks.clear()
+        self.masks.clear()
+        self.setup_table()
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -113,13 +113,16 @@ class MaskManagerDialog(QObject):
             # for threshold mask
             mtype, data = self.masks[key]
             if mtype == 'threshold':
-                self.setup_threshold_connections(cb, pb)
+                self.setup_threshold_connections(cb, i, key)
 
         self.ui.masks_table.blockSignals(False)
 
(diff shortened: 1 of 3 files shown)


bnmajor:
@saransh13 I think this should take care of the bugs that you were seeing but I was still unable to reproduce locally. Could you test this out and see if you're still having problems?

hexrd/ui/mask_manager_dialog.py
```diff
@@ -84,8 +84,10 @@ def setup_connections(self):
         self.ui.export_masks.clicked.connect(self.export_visible_masks)
         HexrdConfig().mode_threshold_mask_changed.connect(
             self.update_masks_list)
+        HexrdConfig().detectors_changed.connect(self.clear_masks)
 
     def setup_table(self, status=True):
+        self.ui.masks_table.blockSignals(True)
```
psavery, line 90: I think blocking signals should be done with a context manager. You can use `block_signals` from `hexrd.ui.utils` for this, like so:
```python
with block_signals(self.ui.masks_table):
    ...
```

Although exceptions are undesirable and hopefully not going to occur, they do happen sometimes, and I think it is best to keep hexrdgui in a usable state (if possible) when they do.

Without the context manager, if an exception were raised in the body of this function, the signals from the table would remain blocked as the user tried to continue using the program (since the code to unblock signals would never be reached).

Also, the context manager restores the previously blocked state once it is finished (rather than just setting `blockSignals(False)`), which can be important if a function higher in the call stack already blocked the signals and expects the signals to remain blocked (and, if the signals become unblocked, this can result in a hard-to-track bug). This may not be important in this specific case, but it makes the code easier to maintain in case we do something like that in the future.

hexrd/ui/resources/ui/mask_manager_dialog.ui
```diff
@@ -19,6 +19,15 @@
...
      </property>
+     <property name="editTriggers">
+      <set>QAbstractItemView::DoubleClicked</set>
+     </property>
+     <property name="alternatingRowColors">
+      <bool>true</bool>
+     </property>
+     <property name="sortingEnabled">
+      <bool>true</bool>
+     </property>
```
psavery, line 30: Can you also set the `selectionBehavior` to be `SelectRows`? I think that is more intuitive for this table especially because the context menu is essentially acting on a whole row, rather than a single item (which includes the particular column that was selected).

hexrd/ui/mask_manager_dialog.py
```diff
@@ -38,10 +38,10 @@ def create_masks_list(self):
         polar_data = HexrdConfig().polar_masks_line_data
         raw_data = HexrdConfig().raw_masks_line_data
 
-        for i, (key, val) in enumerate(polar_data.items()):
+        for _, (key, val) in enumerate(polar_data.items()):
             if not any(np.array_equal(m, val) for m in self.masks.values()):
                 self.masks[key] = ('polar', val)
-        for i, (key, val) in enumerate(raw_data.items()):
+        for _, (key, val) in enumerate(raw_data.items()):
```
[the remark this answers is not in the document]
bnmajor: Cleaned up

hexrd/ui/mask_manager_dialog.py
```diff
@@ -97,39 +99,39 @@ def setup_table(self, status=True):
             status = key in HexrdConfig().visible_masks
             cb.setChecked(status)
             cb.setStyleSheet('margin-left:50%; margin-right:50%;')
-            cb.toggled.connect(self.toggle_visibility)
             self.ui.masks_table.setCellWidget(i, 1, cb)
+            self.ui.masks_table.cellWidget(i, 1).toggled.connect(
+                lambda checked, key=key: self.toggle_visibility(checked, key))
```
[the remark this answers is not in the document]
bnmajor: That is the case and this should have been reverted after I added the lambda approach that you suggested. This change and the change from `clicked` to `released` were remnants from my testing the signals and both have been reverted, thanks!

hexrd/ui/mask_manager_dialog.py
bnmajor: Thank you, this makes sense. Returning to the previous state is a good point and not one I was thinking of, thanks for explaining everything! Added in.

hexrd/ui/resources/ui/mask_manager_dialog.ui
bnmajor: Yes, the behavior has been updated.

hexrd/ui/mask_manager_dialog.py
```diff
@@ -87,34 +87,34 @@ def setup_connections(self):
...
+                self.ui.masks_table.setCellWidget(i, 2, pb)
+                pb.clicked.connect(lambda i=i, key=key: self.remove_mask(i, key))
+
+                # Connect manager to raw image mode tab settings
+                # for threshold mask
+                mtype, _ = self.masks[key]
+                if mtype == 'threshold':
+                    self.setup_threshold_connections(cb, i, key)
+
+            self.ui.masks_table.blockSignals(False)
```
psavery, line 117: This line can now be removed, since the signals will be unblocked automatically.

hexrd/ui/mask_manager_dialog.py
```diff
@@ -84,52 +84,53 @@ def setup_connections(self):
...
+
+    def setup_threshold_connections(self, checkbox, row, name):
         HexrdConfig().mode_threshold_mask_changed.connect(checkbox.setChecked)
         checkbox.toggled.connect(self.threshold_toggled)
+        self.ui.masks_table.cellWidget(row, 2).released.connect(
```
psavery, line 122: Can you change this signal from `released` to `clicked` as well?

psavery (approved):
LGTM. Thank you for your work on this.

bnmajor:
Thanks @psavery! I will merge this in now because it makes some needed changes either way but I will not close the associated issues until @saransh13 can confirm that the bugs he was seeing have been fixed by this branch. So far I have been unable to reproduce them so I am hopeful but not positive.

Issue #1014, opened 2021-08-17
undescriptive error when width is not enabled in powder overlays during powder calibration


saransh13:
When powder calibration is started with a material with the widths turned off, then the following error is generated. This is not descriptive. The user should probably get a warning and the widths should be turned on to the default values.

<img width="417" alt="Screen Shot 2021-08-16 at 8 49 31 AM" src="https://user-images.githubusercontent.com/15834451/129787339-de044bdd-b892-490b-ad6c-9d3176a0e702.png">

PR #1016, opened 2021-08-24
For auto powder, set tth width if it is None

psavery:
Rather than the previous non-descriptive error
(`loop of ufunc does not support argument 0...`), if the visible
powder overlay does not have widths when starting the auto powder
calibration, automatically add a default width and warn the user.

Fixes: #1014


bnmajor (approved):
LGTM

Issue #982, opened 2021-07-13
Need a list editor for the Materials


joelvbernier:
I found the list editor for the materials in @donald-e-boyce's wxPython gui very useful for sorting/adding/copying/deleting materials:
![image](https://user-images.githubusercontent.com/1154130/125502009-63bb4f3e-ca8e-4cf2-b50d-94c16e4d5a68.png)
This could even be taken a step further with Ctrl and Shft click behavior for selecting multiple entires.  Very usefule for removing all the default materials you don't need.

psavery:
For this, I'm thinking it would be nice if we made a general "ListEditorWidget", that allows you to do the above things of moving items up and down in a list, deleting items, copying items, and potentially renaming items, for an arbitrary list of strings. For this particular use case, we probably want the strings to be unique (no duplicates).

When an action is performed, the widget can emit signals (like `item_copied.emit(name)`, `item_deleted.emit(name)`, `item_renamed.emit(old_name, new_name)` and `order_changed.emit()`). These would communicate to the materials panel what to do (copy a material, delete materials, rename a material, or change the order of the materials). I'm not sure whether emitting names or indices would be the best approach, but I know this will be treating the data like a list, whereas the materials internally are in a dict, and they must have unique names.

A dialog with this widget could appear when the user clicks the tool button (next to the materials combo box in the materials panel), which would replace the current tool button behavior of popping up the menu with "Add material", "Import material", and "Delete material". The dialog that appears can have an "Import material" button in it as well so that the user can still import materials.

We are currently storing the materials in a regular python dict in `HexrdConfig().materials` (a property that redirects to `HexrdConfig().config['materials']['materials']`). In several parts throughout hexrdgui, we rely on the materials having unique names, so I think we should keep the materials in some kind of dict, rather than changing it to a list. 

As for modifying the ordering of `HexrdConfig().materials`, dicts are ordered by insertion order as of `python>=3.6`, so we do have the option to continue using a regular python dict. To re-order it, though, we would have to create a new dict, as there are no functions to perform re-ordering, although I think creating a new dict shouldn't be expensive. Alternatively, we could use an [`OrderedDict`](https://docs.python.org/3/library/collections.html#collections.OrderedDict), which includes one function `move_to_end()` for modifying the order. We would need to call this multiple times, most likely, to achieve the desired order. I personally think it would probably be easiest to just create a new regular python dictionary each time with the desired order.

PR #1020, opened 2021-08-30
Add materials list editor

psavery:
This adds a "Materials List Editor" dialog that is used to edit the list
of materials.

The user can change the ordering of the materials, copy, delete, rename,
and import materials.

This "takes over" the functionality of the tool button in the materials
panel. That button no longer displays menu items (which included adding a
material, importing a material, and deleting a material). But now, when
the user clicks on the tool button, the materials list editor appears.

https://user-images.githubusercontent.com/9558430/131402667-2d428fdf-8c7f-4840-b81f-c0444b472d14.mp4

Fixes: #982


bnmajor (approved):
LGTM!

PR #1028, opened 2021-09-10
added dexela function key

joelvbernier:
The GUI was missing a key for the `Dexela_2923` distortion function, which was added to `hexrd` some time back...  The distortion package is a registry, so perhaps we should ask it for all available functions at runtime?

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -1644,6 +1644,8 @@ class HexrdConfig(QObject, metaclass=QSingleton):
             return 0
         elif func_name == 'GE_41RT':
             return 6
+        elif func_name == 'Dexela_2923':
+            return 8
 
         raise Exception('Unknown distortion function: ' + func_name)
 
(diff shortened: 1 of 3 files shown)


psavery (approved):
I added the distortion option for "Dexela_2923" to the GUI. 

Looks good to me!

joelvbernier:
@psavery -- can we populate this straight from `hexrd.distortion.maptypes()`?

psavery:
> @psavery -- can we populate this straight from `hexrd.distortion.maptypes()`?

Sure. It would also make things easier if we added a function like `hexrd.distortion.num_params()`, where you pass in only the `maptype` string (such as `GE_41RT` or `Dexela_2923`) and get the number of parameters for the map type.

joelvbernier:
Agree.  I was accessing that now using `len(DistortionClass.params)`, but it is a good candidate for a property.

psavery:
> Agree. I was accessing that now using `len(DistortionClass.params)`, but it is a good candidate for a property.

That's right. But a difficult thing with the way it's currently set up is that you have to instantiate the distortion class before you can call the `.params` property. And since the params are passed in as an argument, you have to already know the length of the params before calling `DistortionClass.params`.

PR #1030, opened 2021-09-14
Hand picked calibration improvements

psavery:
This adds a button to the line picker dialog that allows the user
to see a tree view of the picks they have chosen, and export these
picks.

The user is currently not able to edit the points, but that is a
feature that is planned for the near future.

example

Addresses several parts in  #761


psavery:
I currently don't know of any bugs with the contents of this PR. It could be tested more thoroughly, but we can also test it more in the future.

Some things left to do with the picks tree view that may be done in a separate PR:
1. Add importing the picks
2. Add ability to "re-pick by hand" a point via a context menu.
3. Add the picks tree view to the auto powder picking workflow, so that the user can also see and modify automatically picked points.

joelvbernier (approved):

Issue #1024, opened 2021-09-08
Need Save/Load for WPPF parameters


joelvbernier:
No method currently exists to save the WPPF parameters in the GUI; the underlying I/O methods exist on the LeBail and Rietveld classes (correct, @saransh13?).  I think we just need to expose those in the GUI on the WPPF widget.  I assume the format with be HDF5, but I defer to @saransh13.

saransh13:
Theres the `dump_hdf5` routine but no `load_hdf5`. I'll add that to WPPF.

PR #1033, opened 2021-09-20
Add ability to export/import WPPF parameters

psavery:
This adds buttons to the WPPF options dialog so that the user can
export/import the WPPF parameters as an HDF5 file.

The importing performs some validation to make sure all of the
keys in the imported file match the keys that are currently in the
GUI. The keys will not match if the user has different materials
selected, and/or a different peak shape method.

wppf

 Fixes: #1024


joelvbernier:
@saransh13 -- can you test/approve this?

joelvbernier (approved):

PR #1034, opened 2021-09-21
Import masks

### LINKED ISSUE #995: Support the option to load masks in mask manager

bnmajor:
This branch adds a new feature that allows users to import masks previously exported from the GUI. The export has been updated to save the raw masking coordinates as well as the mask name and its associated detector. This information is used on import to re-populate the mask manager and apply the masks to the correct detectors in the raw view.

import_masks

Fixes #995

### DIFF
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -1,3 +1,4 @@
+from hexrd.ui.create_raw_mask import convert_polar_to_raw
 import os
 import numpy as np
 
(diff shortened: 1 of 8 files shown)


hexrd/ui/mask_manager_dialog.py
```diff
@@ -242,3 +248,40 @@ def clear_masks(self):
...
+            raw_line_data = HexrdConfig().raw_masks_line_data
+            mask_data = masks_dict['masks']
+            for det, data in mask_data.items():
+                if det not in HexrdConfig().detector_names:
+                    msg = (
+                        f'Detectors must match.\n'
+                        f'Current detectors: {HexrdConfig().detector_names}.\n'
+                        f'Detectors found in masks: {list(mask_data.keys())}')
+                    QMessageBox.warning(self.ui, 'HEXRD', msg)
+                    return
```
psavery, line 278: Nice validation!

bnmajor:
This branch has quite a few updates now and has been tested with the attached file. I was able to create geometric masks in raw, as well as geometric, polygonal (hand drawn masks), and laue in the polar view. Toggling back and forth between views showed the same areas being masked in both views (confirmed by eye as well as comparing the hover values in both views). The attached file was created from these masks and they can be imported back in successfully. All masks should be visible in all views as well now (laue was previously not being displayed in the raw view).
[testing.zip](https://github.com/HEXRD/hexrdgui/files/7213753/testing.zip)

psavery (approved):
I tested this pretty thoroughly just now, and everything seemed to be working very well. Thank you for your work on this, Brianna!

We'll need to add the ability to export/import masks from a larger HDF5 state file. But we can do that in a separate PR. This seems to be a good point to merge this one.

PR #1038, opened 2021-09-23
Reusable mask import export

bnmajor:
This breaks up the import and export functions in the mask manager so that they can be reused elsewhere, like when exporting/importing the state.

The `export_masks` function can now accept a h5py group and will update that group if it is provided. If one is not provided the user is prompted to select or create a file instead. The `import_masks` function is now only called when import is selected in the mask manager dialog. The actual parsing of the h5py group is performed in `load_masks` which is now called from `import_masks` and could be connected to a signal when a state is loaded.

### DIFF
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -278,15 +278,16 @@ class MaskManagerDialog(QObject):
         masks_dict = {}
         with h5py.File(selected_file, 'r') as f:
             unwrap_h5_to_dict(f, masks_dict)
+        self.load_masks(masks_dict['masks'])
 
+    def load_masks(self, h5py_group):
         raw_line_data = HexrdConfig().raw_masks_line_data
-        mask_data = masks_dict['masks']
-        for det, data in mask_data.items():
+        for det, data in h5py_group.items():
             if det not in HexrdConfig().detector_names:
                 msg = (
                     f'Detectors must match.\n'
                     f'Current detectors: {HexrdConfig().detector_names}.\n'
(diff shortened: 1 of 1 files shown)


hexrd/ui/mask_manager_dialog.py
```diff
@@ -84,8 +84,8 @@ def setup_connections(self):
         self.ui.masks_table.cellDoubleClicked.connect(self.get_old_name)
         self.ui.masks_table.cellChanged.connect(self.update_mask_name)
         self.ui.masks_table.customContextMenuRequested.connect(
-            self.context_menu_event)
-        self.ui.export_masks.clicked.connect(self.export_visible_masks)
+            self.export_masks)
```
psavery, line 87: I think it makes more sense for `self.context_menu_event()` to call `export_masks()` rather than `self.export_masks()` to call the context menu event.

hexrd/ui/mask_manager_dialog.py
```diff
@@ -245,7 +251,7 @@ def export_visible_masks(self):
             for i, (det, mask) in enumerate(data):
                 parent = d.setdefault(det, {})
                 parent.setdefault(name, {})[str(i)] = mask
-        self.export_masks(d)
+        return d
```
psavery, line 254: I think it also makes sense for `export_visible_masks()` to export the masks, rather than return the masks to be exported.

joelvbernier:
I'll wait for your word @bnmajor

joelvbernier (approved):

joelvbernier:
Merge pending resolution of @psavery's comments.

PR #1044, opened 2021-09-24
Store all masks as raw

### LINKED ISSUE #1041: Polar masks are not all getting re-mapped consistently
Since masks drawn in the polar view get mapped out to (x, y) coordinates, the behavior should be such that they 'track' with the features in the polar view if the instrument parameters are modified.  This seems like it _might_ be the case for the laue masks, but definitely is not the case with the polygon and ellipse/rectangle type.  Observe:
image
and if I change the beam direction:
image
The masked regions should be persistent in the raw view, not the polar view.
image

bnmajor:
Fixes #1041


hexrd/ui/hexrd_config.py
```diff
@@ -1346,7 +1344,7 @@ def validate_overlays(self):
 
     @property
     def visible_polar_masks(self):
```
psavery, line 1346: Maybe we should change this name to `visible_masks`? But that can be done another time...

psavery (changes requested):
It looks good, and I gave it a try and it seems to work!

Just one warning to fix before I approve:
```python
/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/create_raw_mask.py:55: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  det_lines = [line for line in line_data if det in line]
```

hexrd/ui/hexrd_config.py
bnmajor: It was a quick and easy fix so I updated it

joelvbernier (approved):

bnmajor:
@psavery The comparison has been reverted so that the FutureWarning does not appear

psavery (approved):
LGTM

hexrd/ui/hexrd_config.py
```diff
@@ -1345,8 +1343,8 @@ def validate_overlays(self):
                 overlay['refinements'] = default_refinements
 
     @property
-    def visible_polar_masks(self):
-        masks = self.polar_masks.items()
+    def visible_masks(self):
```
joelvbernier, line 1346: this property has no setter, and there are various places trying to modify it.

PR #1045, opened 2021-09-24
Mask powder rings

bnmajor:
This branch is a quick first pass at supporting masking of powder rings. Testing on the tardis example looks good, but testing on the hydra example is showing some weird behavior on detector `ge2` in the raw view for reasons I have not yet figured out

polar_powder_masking
raw_powder_masking
ge_powder_polar
ge_powder_raw

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -618,6 +618,45 @@ class MainWindow(QObject):
         self.new_mask_added.emit(self.image_mode)
         HexrdConfig().polar_masks_changed.emit()
 
+    def action_edit_apply_powder_mask_to_polar(self):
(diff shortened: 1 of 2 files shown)


bnmajor:
The issue that I was seeing in the hydra case is unrelated to the changes in this branch. @joelvbernier will be creating a fix for this bug which has to do with trying to map the branch cut for eta since the points are generated in polar coordinates first.

joelvbernier (approved):

joelvbernier:
I'll fix this -- the issue is that the &eta; branch cut is on more or less the middle of ge2, so the points don't monotonically increase from one side to the other.  I'll have to fix this in the overlay generation.

PR #1051, opened 2021-09-27
Remove HexrdConfig().visible_polar_masks

psavery:
There is no longer any distinction between polar and raw masks, so
remove the visible polar masks property...

I gave this a quick test and it worked


bnmajor (approved):
LGTM!

PR #1052, opened 2021-09-29
Calibration features jvb

joelvbernier:
Reconciled some discrepancies in handling of overlays with offers in the polar view, and added machinery for auto-picking Laue spots.

### DIFF
--- a/hexrd/ui/calibration/pick_based_calibration.py
+++ b/hexrd/ui/calibration/pick_based_calibration.py
@@ -158,7 +158,7 @@ class LaueCalibrator(object):
     def _autopick_points(self, raw_img_dict, tth_tol=5., eta_tol=5.,
                          npdiv=2, do_smoothing=True, smoothing_sigma=2,
                          use_blob_detection=True, blob_threshold=0.25,
-                         fit_peaks=True):
+                         fit_peaks=True, min_peak_int=1., fit_tth_tol=0.1):
         """
 
 
@@ -338,8 +338,8 @@ class LaueCalibrator(object):
                               % (det_key, iRefl, coms[0], coms[1],
                                  patch_size[0], patch_size[1]))
                         '''
-                        row_cen = 0.08*patch_size[0]
-                        col_cen = 0.08*patch_size[1]
+                        row_cen = fit_tth_tol * patch_size[0]
+                        col_cen = fit_tth_tol * patch_size[1]
                         if np.any(
                             [coms[0] < row_cen,
                              coms[0] >= patch_size[0] - row_cen,
@@ -347,7 +347,7 @@ class LaueCalibrator(object):
                              coms[1] >= patch_size[1] - col_cen]
                         ):
                             continue
-                        if (fit_par[0] < 1.):
+                        if (fit_par[0] < min_peak_int):
                             continue
 
                         # intensities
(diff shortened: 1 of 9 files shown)


hexrd/ui/overlays/powder_diffraction.py
```diff
@@ -124,8 +124,16 @@ def generate_ring_points(self, tths, etas, panel, display_mode):
...
-            xys_full = panel.angles_to_cart(ang_crds, tvec_c=self.tvec)
+            # Convert nominal powder angle coords to cartesian
+            # !!! Tricky business; here we must consider _both_ the SAMPLE
+            #     CS origin and anything specified for the XRD COM for the
+            #     overlay.  This is so they get properly mapped back to the
+            #     the proper cartesian coords.
+            xys_full = panel.angles_to_cart(
+                ang_crds,
+                tvec_s=self.instrument.tvec,
+                tvec_c=self.tvec
```
psavery, line 135: This makes me think we need to fix `tvec` in other calls to `panel.angles_to_cart()` in hexrdgui. It looks to me that in several places, we have been using `instr.tvec` for `tvec_c`, whereas this looks to me like we should be using the powder options `tvec` and set `tvec_s` to the instrument `tvec.

hexrd/ui/overlays/powder_diffraction.py
```diff
@@ -138,10 +146,14 @@ def generate_ring_points(self, tths, etas, panel, display_mode):
...
 
             if display_mode == ViewType.polar:
-                # !!! apply offset correction
-                ang_crds, _ = panel.cart_to_angles(xys,
-                                                   tvec_c=self.instrument.tvec)
-
+                # Apply offset correction
+                # !!! In polar view, the nominal angles refer to the SAMPLE
+                #     CS origin, so we omit the addition of any offset to the
+                #     diffraction COM in the sameple frame!
```
psavery, line 152: Tricky business!

hexrd/ui/calibration/pick_based_calibration.py
```diff
@@ -45,12 +53,13 @@ def enrich_pick_data(picks, instr, materials):
                     )
                     pick_data[data_key][det_key] = xy_picks
             elif pick_type == 'powder':
-                # need translation vector
+                # !!! need translation vector from overlay
```
psavery, line 56: That `pick_data['options']['tvec']` IS coming from the overlay

hexrd/ui/calibration/pick_based_calibration.py
```diff
@@ -93,7 +102,8 @@ def instr(self):
 
     @property
     def plane_data(self):
-        self._plane_data.wavelength = self._instr.beam_energy
+        self._plane_data.wavelength = self.energy_cutoffs[-1]
+        self._plane_data.exclusions = None
```
psavery, line 106: Do we need to make a deep copy of the plane data so the hexrdgui one doesn't get modified? (which will make the GUI out of sync with the underlying plane data)

psavery:
The Laue auto picking seems to be working, although for the tardis example I have, it appears to only pick 3 points?
image

Those points look accurate, however. Not sure what's wrong, I can confirm the plane data has the right hkls and the min and max energy cutoffs are the same. Can you check @joelvbernier?

psavery:
Oh, it seems if I play with the options just a little bit, I get all of the picks:

image

I think they look good too. I guess this PR is ready for review...

joelvbernier:
LGTM -- I can work on parameter optimization later...  Thanks for getting this done!

psavery (approved):

PR #1061, opened 2021-10-14
Move import panels

bnmajor:
Move the import tools for the HEDM and LLNL workflows into dialogs instead. There is now an `Import` menu under `File` with options to open the LLNL, HEDM, or image stack import tools. The option to select the workflow has been removed.
import_menu

@joelvbernier I am sure the naming for these menu options will need to be tweaked to something more appropriate but this is ready for you to take a look at. As of now:
- The HEDM tool does not close until you press `okay`, allowing you to keep tweaking options/reloading if neccessary. Pressing okay will load the images if they have not been read in already. Cancel will not read the images in but *will* remember any settings/selections.
- The image stack dialog is no longer part of the HEDM tool, but is now its own menu option. If an image stack is selected and all values are properly set the HEDM tool will be opened afterwards to allow the user to complete and final processing they may need.
- Pressing "okay" on the LLNL tool will close the dialog and load all of the cropped and masked images.

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -99,8 +99,8 @@ class MainWindow(QObject):
         self.add_materials_panel()
 
         self.simple_image_series_dialog = SimpleImageSeriesDialog(self.ui)
-        self.llnl_import_tool_dialog = LLNLImportToolDialog(self.color_map_editor,
-                                                  self.ui)
+        self.llnl_import_tool_dialog = LLNLImportToolDialog(
+                                        self.color_map_editor, self.ui)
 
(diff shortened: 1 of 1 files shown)


bnmajor:
@psavery Thanks! I've updated the dialog titles to `Image Stack Import`, `LLNL Import Tool`, and `Simple Image Series Import` to better reflect what each dialog is meant to be used for. I also updated the guard around removing any existing template boundaries in order to prevent the error that was produced if the LLNL dialog was opened and then immediately closed. In the `Simple` dialog there is now logic to only support re-reading an imageseries if one or more of the selections have changed. This prevents unnecessary reloading without ignoring changes that have been made before the dialog is accepted.

PR #1070, opened 2021-10-19
Reverse frames

### LINKED ISSUE #1062: Need option in imageseries processing for reversing frame order
Currently, `OmegaImageSeries` only supports _positive_ (ccw) &Delta;&omega; while we eventually will include support for instruments that rotate cw (@donald-e-boyce...), there is a need to support rotation series measured as such now.

The `ProcessedImageSeries` class supports a `'frame_list'` kwarg that allows us to reverse the order of the frames, _e.g._
```
pims = imageseries.process.ProcessedImageSeries(ims, opt_list, frame_list=np.arange(len(ims))[::-1])
```
Since we use `ProcessedImageSeries` already, this should be an easy option to support -- when everything lines up in the GUI, but a user can't index any grains, the reversed (cw) omega is the cause 99% of the time!

bnmajor:
Fixes #1062 

Will need to rebased on master after #1061 is merged in

@joelvbernier This adds the option to `Simple Imageseries Import` dialog since that seems to be where it is needed. If we need the option elsewhere as well let me know.
reverse_ims

### DIFF
--- a/hexrd/ui/image_stack_dialog.py
+++ b/hexrd/ui/image_stack_dialog.py
@@ -53,6 +53,7 @@ class ImageStackDialog(QObject):
         self.clear_images.connect(
             self.simple_image_series_dialog.clear_from_stack_dialog)
         self.ui.add_omega.toggled.connect(self.add_omega_toggled)
+        self.ui.reverse_frames.toggled(self.reverse_frames)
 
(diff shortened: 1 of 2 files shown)


joelvbernier:
I think we could use it in the stack import as well!

PR #1077, opened 2021-10-21
added proper extent to polarview class and use it in the imshow

joelvbernier:
I noticed that there was a discrepancy between the peak positions in plotted polar image and the row-wise (read: azimuthal) sum that depended on the 2&theta; pixel size.  Further examination showed that the cause was the extents; in the existing class the axis extents are static, fixed to the interactor values.  This is not correct, since the max values may differ from the user-requested value by up to a pixel due to truncation.  The true angular grid vectors in the polarview generator are effectively calculated as
```
nbins = int((max_val - min_val)/bin_size)
bin_edges = bin_size*np.arange(nbins + 1) - min_val
```
so that if the range is not evenly divisible but the bin_size, the _true_ max with be less than the requested.

I added an `extent` property on the `PolarView` class, and use it in the image render.  Tested and looks good!

### DIFF
--- a/hexrd/ui/resources/ui/image_mode_widget.ui
+++ b/hexrd/ui/resources/ui/image_mode_widget.ui
@@ -1,827 +1,827 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ui version="4.0">
-  <class>image_mode_widget</class>
-  <widget class="QWidget" name="image_mode_widget">
-    <property name="geometry">
-      <rect>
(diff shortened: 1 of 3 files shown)


joelvbernier:
I want testing Brianna's branch, and on my windows, somehow the changed files got DOS-ified.  There should be only whitespace differences outside of `polarview.py` and `polar_plot.py`.  Sorry for the mess!

PR #1089, opened 2021-10-29
Change excluded grain columns to [9, 15)

psavery:
In the fit grains options dialog, the previously excluded grain columns
would be fixed values (null values, essentially) if the user ran through
the indexing workflow to get the grains.

However, if the user loads in the grains via a `grains.out` file, these
may be non-null values, and it can be useful to see them.

Thus, show the parameters in case they are non-null.


joelvbernier (approved):

joelvbernier:
LGTM!

PR #1104, opened 2021-12-21
Bump Python to 3.9

cjh1:
Motivated by issues with the version of Mac OS X SDK used to build
python 3.8.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -44,7 +44,7 @@ jobs:
       uses: conda-incubator/setup-miniconda@v2
       with:
         auto-update-conda: true
-        python-version: 3.8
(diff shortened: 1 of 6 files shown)


psavery:
This looks good to me. I guess we just need to test it on the different operating systems?

psavery:
Will do!

cjh1:
I can confirm the generated zip and msi work on Windows.

psavery:
I tested the dmg file by running through the HEDM workflow on my M1 Macbook Pro and did not encounter any noticeable issues. I also double-checked the sdk for the python inside and it was built with 11.0, so that looks good.

It looks good to me in my testing!

psavery (approved):
LGTM

PR #1106, opened 2021-12-23
Calibration fixes

joelvbernier:
More updates to the calibration routines.  We will eventually need to replace these with classes from hexrd...

### DIFF
--- a/hexrd/ui/calibration/pick_based_calibration.py
+++ b/hexrd/ui/calibration/pick_based_calibration.py
@@ -613,6 +613,9 @@ class PowderCalibrator(object):
                        raw cartesian coordinates
                     4. build residual on the measured and recalculated (x, y)
                 """
+                # the data structure is:
+                #     [x, y, tth, eta, h, k, l, dsp0]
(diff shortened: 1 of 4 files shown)


hexrd/ui/calibration/pick_based_calibration.py
```diff
@@ -605,24 +604,45 @@ def _evaluate(self, reduced_params, data_dict, output='residual'):
...
+                # to (tth, eta)
+                meas_xy = pdata[:, :2]
+                updates_angles, _ = panel.cart_to_angles(
+                    meas_xy,
+                    tvec_s=self.instr.tvec,
+                    apply_distortion=True
+                )
+
+                # derive ideal tth positions from additional ring point info
+                hkls = pdata[:, 3:6]
```
joelvbernier, line 626: This is incorrect; should be `pdata[:, 4:7]`

joelvbernier: this was fixed in the commit below

PR #1109, opened 2022-01-04
Fix TARDIS IP template for LLNL workflow

bnmajor:
Do not set polygon patches to "closed" if there are multiple polygons being drawn within one patch. This prevents a bug with the `TARDIS` instrument where a line was being drawn between the inner and outer polygon in an attempt to "close" the shape.

### DIFF
--- a/hexrd/ui/interactive_template.py
+++ b/hexrd/ui/interactive_template.py
@@ -41,6 +41,10 @@ class InteractiveTemplate:
         verts = self.panels['default'].cartToPixel(data)
         verts[:, [0, 1]] = verts[:, [1, 0]]
         self.shape = patches.Polygon(verts, fill=False, lw=1, color='cyan')
+        if np.any(np.isnan(verts)):
+            # This template contains more than one polygon and the last point
+            # should not be connected to the first. See Tardis IP for example.
+            self.shape.set_closed(False)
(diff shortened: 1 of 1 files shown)


hexrd/ui/interactive_template.py
```diff
@@ -41,6 +41,10 @@ def create_shape(self, module, file_name, det, instr):
         verts = self.panels['default'].cartToPixel(data)
         verts[:, [0, 1]] = verts[:, [1, 0]]
         self.shape = patches.Polygon(verts, fill=False, lw=1, color='cyan')
+        if np.any(np.isnan(verts)):
```
psavery, line 44: FYI, there's a util function for this that's supposed to be the fastest way to check for nans in an array: `hexrd.ui.utils.has_nan`.

But this is probably fine if `verts` isn't going to be too big.

psavery (approved):
This fixes the issue for me! LGTM

hexrd/ui/interactive_template.py
bnmajor: Lets keep things consistent and simple since there is no guarantee we won't need a larger/more complicated boundary for a future instrument. Updated to use `hexrd.ui.utils.has_nan`. Thanks!

joelvbernier:
Thanks!

PR #1113, opened 2022-01-10
Move dockable widgets to right of main canvas

bnmajor:
Updated layout with the dock widgets all on the right side to start:
new_layout

### DIFF
--- a/hexrd/ui/resources/ui/main_window.ui
+++ b/hexrd/ui/resources/ui/main_window.ui
@@ -41,7 +41,7 @@
      <x>0</x>
      <y>0</y>
      <width>1600</width>
-     <height>26</height>
+     <height>22</height>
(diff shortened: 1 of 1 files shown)


psavery:
I think we want to be careful to not reduce the default size of the canvas. It looks smaller here to me.

bnmajor:
@psavery It is smaller with this new layout to allow for the minimum size requirements of all of the widgets on the right and left. We could shrink the minimum sizes of the widgets? That would require re-doing the layouts a bit or allowing a horizontal scroll bar though. I'm happy to make any tweaks we need, this first pass is just a direct replicate of the new layout that Joel requested:
Screen Shot 2022-01-06 at 12 41 56 PM

joelvbernier (approved):

PR #1115, opened 2022-01-10
Restyle button for the tree view on MacOS

### LINKED ISSUE #1114: MacOS QPushButton alignment in QTreeView
For some reason on the mac the styling is different than on other platforms and the `QPushButton` added in the tree view for the panel buffer is cut off:
Screen Shot 2022-01-06 at 12 41 56 PM

When comparing the mac to linux there seems to be some sort of margin set on the button or padding set on the tree row/around the item delegate (although we don't see this issue with the checkbox):
linux_margin
mac_margin

After looking around for users with similar issues and attempting to adjust the style every way I could think of there were three solutions that I came up with, all of which have trade-offs:
1. Setting the padding to zero pixels in the button style sheet. This does adjust the text but not the button border for some reason:
  
    mac_adjusted_padding

2. Setting the padding to zero and setting the border in the style sheet as well. This redraws the border in the correct position but loses the mac styling (setting the style sheet on an item overrides whatever base os styling is initially applied):
  
    adjusted_border_mac'

3. There is an option to set a universal style that is platform independent. This fixes the button styling issue, but applies the Linux styling across all platforms, which means Windows and Mac lose their nicer, os-consistent styling (This screenshot is on the mac):
  
    fusion_styling

bnmajor:
Fixes #1114

### DIFF
--- a/hexrd/ui/tree_views/value_column_delegate.py
+++ b/hexrd/ui/tree_views/value_column_delegate.py
@@ -28,6 +28,10 @@ class ValueColumnDelegate(QStyledItemDelegate):
         key = item.data(BaseTreeItemModel.KEY_COL)
         if key == constants.BUFFER_KEY:
             edit_btn = QPushButton(BUTTON_LABEL, parent)
+            edit_btn.setStyleSheet(
+                'padding: 0px; border: 1px solid lightgray;'
+                'border-radius: 5px; background-color: white;'
+            )
 
             def _clicked():
                 def _set_enable(enabled):


hexrd/ui/tree_views/value_column_delegate.py
```diff
@@ -28,6 +28,10 @@ def createEditor(self, parent, option, index):
         key = item.data(BaseTreeItemModel.KEY_COL)
         if key == constants.BUFFER_KEY:
             edit_btn = QPushButton(BUTTON_LABEL, parent)
+            edit_btn.setStyleSheet(
+                'padding: 0px; border: 1px solid black;' +
```
psavery, line 32: FYI, you don't need the plus sign here, but no big deal.

psavery (approved):
LGTM

PR #1128, opened 2022-01-13
Default update_needed to True if its not set

cjh1:
It may not be set when loading an old state file.


cjh1:
@joelvbernier This will resolve your issue.

psavery (approved):

joelvbernier:
Thanks @cjh1 and @psavery

Issue #510, opened 2020-09-11
Add warning dialog on quit


joelvbernier:
Just accidentally killed the GUI when my CMD + TAB was a CMD + Q on mac, which makes me appreciate the utility of a warning dialog to ask the user if they are SURE they want to quit 🤣

PR #1129, opened 2022-01-14
Prompt user to confirm exit on application quit

### LINKED ISSUE #1116: Warning dialog on GUI quit
I think a worthwhile feature would be a warning dialog asking the user if they are sure they want to quit upon catching one of the various ways of sending the terminate signal to the GUI.  I have on more than one occasion killed the GUI accidentally (by an errant ctrl+q or otherwise...).

bnmajor:
Fixes: #510
Fixes #1116

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -73,6 +73,7 @@ class MainWindow(QObject):
 
         loader = UiLoader()
         self.ui = loader.load_file('main_window.ui', parent)
+        self.confirm_application_close = True
 
         self.thread_pool = QThreadPool(self)
         self.progress_dialog = ProgressDialog(self.ui)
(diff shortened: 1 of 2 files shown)


hexrd/ui/main_window.py
```diff
@@ -747,6 +747,14 @@ def update_config_gui(self):
 
     def eventFilter(self, target, event):
         if type(target) == QMainWindow and event.type() == QEvent.Close:
+            msg = 'Are you sure you want to quit?'
+            buttons = (QMessageBox.No | QMessageBox.Yes)
+            default = QMessageBox.No
+            response = QMessageBox.question(
+                self.ui, 'HEXRD', msg, buttons, default)
```
psavery, line 754: It looks to me that the `buttons` and `default` arguments are the same as the defaults for `QMessageBox.question()`. Maybe we can remove them?

psavery (commented):
It works for me in my testing! I guess we just need to figure out how to make it work with the CI.

hexrd/ui/main_window.py
bnmajor: Yes, thank you. Removed.

joelvbernier:
@bnmajor and @psavery -- looks like we have some CI issues to address?

psavery (approved):
Looks excellent to me!

PR #1141, opened 2022-01-26
Refactor overlays to be classes instead of dicts

psavery:
The overlay dicts have become more and more complicated as time has
gone on, and new synchronizations between their values and other
parts of the program (such as the image series) have made them even
more complicated to maintain.

As such, it has become a good time to convert them to be classes
instead of dicts, where the classes can internally keep up with any
synchronizations that are needed.

This will also hopefully fix a bug we were encountering where an
overlay was being modified to have a refinements list from another
overlay. This shouldn't be possible with objects.

There is a new format for serializing the overlays, but backward
compatible readers are also now used so that the transition is
seamless.

Since the overlays are used in many, many parts of the code, we
really need to do some thorough testing to find and fix any bugs
this refactoring may cause.


psavery:
I've tested several different workflows with this PR, and so far, things seem to be working fine.

I think it will be ready for merging after our next release (don't want to merge it before the next release in case it introduces new bugs).

psavery:
This is ready as long as the last hexrdgui release is in good shape.

joelvbernier (approved):

PR #1143, opened 2022-01-28
Catch IndexError on mouse move

psavery:
This error would happen occasionally in the cartesian and polar views
when the mouse is moved close to an edge of the canvas. It appears that
it was happening because the x_data for the event was slightly over the
max, which led to an out-of-bounds index being computed. This would
then result in an IndexError.

This error is relatively harmless, but it may worry users, so catch it
to prevent it from appearing.

I believe this error has been happening more commonly for me after #1138 was merged.


bnmajor (approved):
I was not able to replicate the error myself but this looks good!

PR #1144, opened 2022-01-28
Convert to dockable

bnmajor:


bnmajor:
PEP8 issues are all because of the `:=` operator.

hexrd/ui/image_stack_dialog.py
```diff
@@ -367,11 +368,13 @@ def build_data(self):
                 'max_file_frames': self.state['max_file_frames'],
                 'max_total_frames': self.state['max_total_frames']
             },
-            'reverse_frames': self.state['reverse_frames']
+            'reverse_frames': self.state.get('reverse_frames', False)
         }
-        if not self.state['omega_from_file']:
+        if not self.state['omega_from_file'] and len(self.state["wedges"]):
```
psavery, line 373:
```suggestion
        if not self.state['omega_from_file'] and self.state['wedges']:
```

If `self.state['wedges']` will be a list, dict, string, or `None`, then this is the [pep8 recommended](https://www.python.org/dev/peps/pep-0008/) way to check it.

psavery (commented):
I've tested the `SimpleImageSeries` dialog and it's looking good!

A few things I'm wondering about:

1. It looks like you can place the other dock widgets (the color map editor, for instance) in any position around the render window. However, for these import widgets, you can only place them to the left of the render window. Do you know why that is?
2. In "View"->"Dock Widgets", the "Simple Image Series" and "LLNL Import Tool" are there, but the "Image Stack" is missing.
3. Not for this PR, but it might be nice if we eventually save the user's dock widget settings to the QSettings, so that wherever they prefer their dock widgets to be, they will be there if they close and re-open the application.

bnmajor:
@psavery
> 1. It looks like you can place the other dock widgets (the color map editor, for instance) in any position around the render window. However, for these import widgets, you can only place them to the left of the render window. Do you know why that is?

I am honestly not sure about this. All three dialogs have the allowed dock areas set to "all". I am able to dock the Simple Image Series on the right or left but not top or bottom and LLNL and Image stack can only be docked to left. I can dig into this some more and try to figure out what is going on.

> 2. In "View"->"Dock Widgets", the "Simple Image Series" and "LLNL Import Tool" are there, but the "Image Stack" is missing.

This has been fixed.

> 3. Not for this PR, but it might be nice if we eventually save the user's dock widget settings to the QSettings, so that wherever they prefer their dock widgets to be, they will be there if they close and re-open the application.

Sounds good. I will create an issue for that so that I come back to it. Maybe I can also add the only-dock-on-left bug to that issue as well so that we can get this in for now? Then in the new PR I can address docking everywhere and remembering that choice.

hexrd/ui/llnl_import_tool_dialog.py
```diff
@@ -76,9 +76,8 @@ def setup_connections(self):
             self.update_template_style)
         self.ui.line_color.clicked.connect(self.pick_line_color)
         self.ui.line_size.valueChanged.connect(self.update_template_style)
-        self.ui.cancel.clicked.connect(self.ui.reject)
-        self.ui.rejected.connect(self.reset_panel)
-        self.ui.rejected.connect(self.cancel_workflow.emit)
+        self.ui.cancel.clicked.connect(self.close_widget)
+        self.ui.cancel.clicked.connect(self.cancel_workflow.emit)
```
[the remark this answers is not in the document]
bnmajor: Yes, this has been changed.

hexrd/ui/image_stack_dialog.py
```diff
@@ -392,49 +395,47 @@ def check_steps(self):
...
+                    f'total number of frames: {steps} total steps, '
+                    f'{total_frames} total frames.')
             else:
-                return False
+                msg = f'The omega wedges are incomplete.'
+            QMessageBox.warning(None, 'HEXRD', msg)
+            error = True
+        if not error:
+            self.build_data()
+            self.close_widget()
```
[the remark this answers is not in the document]
bnmajor: Fixed

hexrd/ui/image_stack_dialog.py
bnmajor: Thank you, fixed.

hexrd/ui/llnl_import_tool_dialog.py
```diff
@@ -473,18 +476,25 @@ def completed(self):
...
 
     def show(self):
         self.ui.show()
+
+    def close_widget(self, accepted=False):
+        self.reset_panel()
+        if not accepted:
+            self.ui.cancel.clicked.connect(self.cancel_workflow.emit)
+        if self.ui.isFloating():
+            self.ui.close()
```
psavery, line 500: I think `self.cancel_workflow.emit()` needs to be called, rather than a new connection being made, right?

And actually, I didn't realize you'd have to add an extra argument and extra logic to implement this correctly. Maybe we can instead change this function back to the way it was before, and add a function like this:
```python
def on_canceled(self):
    self.close_widget()
    self.cancel_workflow.emit()
```

The `self.ui.cancel.clicked` signal can then be connected to this instead.

hexrd/ui/llnl_import_tool_dialog.py
```diff
@@ -249,7 +248,7 @@ def load_images(self):
                 if not success:
                     return
 
-            ImageLoadManager().read_data(files, parent=self.ui)
+            ImageLoadManager().read_data(files, parent=self.ui.parent())
```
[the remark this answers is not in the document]
psavery: Can we add that comment to before the other `ImageLoadManager().read_data()` call as well?

hexrd/ui/llnl_import_tool_dialog.py
bnmajor: That connect should not have been there, that was a sloppy copy and paste, sorry. Cleaned up and updated to use the `on_canceled` method.

hexrd/ui/llnl_import_tool_dialog.py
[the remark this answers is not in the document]
bnmajor: Added the comment to the second `read_data` as well.

hexrd/ui/llnl_import_tool_dialog.py
psavery: Looks good now

psavery:
> @psavery
> 
> > 1. It looks like you can place the other dock widgets (the color map editor, for instance) in any position around the render window. However, for these import widgets, you can only place them to the left of the render window. Do you know why that is?
> 
> I am honestly not sure about this. All three dialogs have the allowed dock areas set to "all". I am able to dock the Simple Image Series on the right or left but not top or bottom and LLNL and Image stack can only be docked to left. I can dig into this some more and try to figure out what is going on.

It looks like if we move the other dock widgets out, that we can move those in. So I think things are probably good.

psavery (approved):
LGTM

PR #1146, opened 2022-01-31
Update enable states immediately in select grains

psavery:
This fixes an issue where the "OK" button would be disabled sometimes
when the SelectGrainsDialog was first initialized, even though there
was a valid selection.


bnmajor (approved):
LGTM

Issue #1153, opened 2022-02-05
Issues with overlay refactor


joelvbernier:
- [x] Laue overlays are not getting the eta ranges from the current polar view interactors; fixed to (-180º, 180º)
- [x] Slider behavior in the Crystal Parameters panel of the overlay widget is weird under change of range
- [x] The colormap in the hand point picker seems to be set to the raw; it should mirror the canvas scaling (probably unrelated...)
- [x] changing lattice parameters throws the following exception: 
    ```
    Traceback (most recent call last):
  File "/Users/bernier2/Documents/GitHub/hexrdgui/hexrd/ui/materials_panel.py", line 179, in material_edited
    self.update_refinement_options()
  File "/Users/bernier2/Documents/GitHub/hexrdgui/hexrd/ui/materials_panel.py", line 268, in update_refinement_options
    self._overlay_manager.update_refinement_options()
  File "/Users/bernier2/Documents/GitHub/hexrdgui/hexrd/ui/overlay_manager.py", line 222, in update_refinement_options
    self.overlay_editor.update_refinement_options()
  File "/Users/bernier2/Documents/GitHub/hexrdgui/hexrd/ui/overlay_editor.py", line 78, in update_refinement_options
    self.powder_overlay_editor.update_refinement_options()
AttributeError: 'PowderOverlayEditor' object has no attribute 'update_refinement_options'
```

psavery:
It looks like the slider behavior is also weird in 0.8.15 under a change of range. We should make the behavior similar to that of the sliders in the instrument calibration slider view.

joelvbernier:
@psavery added a checkbox for the slider behavior above.  Also added a new exception.

psavery:
The last item is fixed by #1169

PR #1158, opened 2022-02-11
Ensure overlays have the correct eta period

psavery:
This keeps the overlays in sync with the polar res eta period on the
HexrdConfig object so that it will match the user selections.

Previously, the overlays were not receiving the eta periods at all...

Fixes the first item in #1153


joelvbernier (approved):

joelvbernier:
LGTM

PR #1159, opened 2022-02-15
Do not reset the minimum width for the slider

### LINKED ISSUE #1152: Moving frame slider for omega image series causes window to grow in size
If you load an omega image series in unaggregated mode, where there is an "Omega range" displayed along with the frame slider, and your window is not maximized, then when you slide along, the main window grows larger.

It looks to me that the main window growing larger is related to the minimum size of the slider increasing [here](https://github.com/HEXRD/hexrdgui/blob/ace51145e34f15809d0c27e4de3f3142d0dd7f21/hexrd/ui/image_series_toolbar.py#L59). We should look into this.

bnmajor:
Fixes #1152

Bug:
slider_bug
Now fixed:
slider_fixed


psavery (approved):
LGTM

PR #1165, opened 2022-02-18
Wait to apply masks until state is loaded

bnmajor:
This prevents a key mismatch bug that can occur where the detector names have been updated but the imageseries has not yet been loaded, so the masks cannot be applied correctly.


psavery (approved):
LGTM

PR #1185, opened 2022-04-09
increased ranges for spinboxes

joelvbernier:
looking at EIger data with 32-bit range, was pegging the max val.


psavery (commented):
LGTM

joelvbernier:
👍🏼

PR #1187, opened 2022-04-11
Update ImageStackDialog detectors on change

### LINKED ISSUE #1181: Stack importer detector keys not being updated upon loading a new state file
I was working in a session with detectors `Cspad_1, Cspad_2, ...`.  I loaded a different session with a single detector `GE2`.  When I open the stack importer, the old detector keys remain, and even if I import the desired imageseries with the "simple" importer, these key values will not change, leading to an exception when trying to load.  Seems like a missing signal indicating the instrument has changed.
Screen Shot 2022-04-04 at 8 30 17 PM

bnmajor:
Fixes #1181

### DIFF
--- a/hexrd/ui/image_stack_dialog.py
+++ b/hexrd/ui/image_stack_dialog.py
@@ -114,13 +114,16 @@ class ImageStackDialog(QObject):
                 'add_omega_data': True,
                 'reverse_frames': False,
             }
-            for det in self.detectors:
-                self.state[det] = {
-                    'directory': '',
-                    'files': '',
-                    'search': '',
-                    'file_count': 0,
-                }
+        self.setup_detectors_state()
+
+    def setup_detectors_state(self):
+        for det in self.detectors:
+            self.state[det] = {
+                'directory': '',
+                'files': '',
+                'search': '',
+                'file_count': 0,
+            }
 
     def set_wedges(self):
         if self.ui.omega_wedges.rowCount() == 0:
@@ -446,13 +449,7 @@ class ImageStackDialog(QObject):
         self.detector = self.detectors[0]
         # update the state
         self.state['dets'] = self.detectors
-        for det in self.detectors:
-            self.state[det] = {
-                'directory': '',
-                'files': '',
-                'search': '',
-                'file_count': 0,
-            }
+        self.setup_detectors_state()
         # update the GUI
         self.ui.detectors.clear()
         self.ui.detectors.addItems(self.detectors)
--- a/hexrd/ui/image_stack_dialog.py
+++ b/hexrd/ui/image_stack_dialog.py
@@ -58,6 +58,7 @@ class ImageStackDialog(QObject):
         self.ui.reverse_frames.toggled.connect(self.reverse_frames)
         self.ui.button_box.accepted.connect(self.check_data)
         self.ui.button_box.rejected.connect(self.close_widget)
+        HexrdConfig().detectors_changed.connect(self.detectors_changed)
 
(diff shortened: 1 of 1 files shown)


hexrd/ui/image_stack_dialog.py
```diff
@@ -437,3 +439,20 @@ def show(self):
...
+        self.detector = self.detectors[0]
+        # update the state
+        self.state['dets'] = self.detectors
+        for det in self.detectors:
+            self.state[det] = {
+                'directory': '',
+                'files': '',
+                'search': '',
+                'file_count': 0,
+            }
```
psavery, line 454: I see this is copied from `setup_state()`. Maybe we should put it in a function or property to reduce duplicated code? This would be especially helpful in case we decided to change the default detector settings in the future.

hexrd/ui/image_stack_dialog.py
```diff
@@ -437,3 +439,20 @@ def show(self):
...
         if self.ui.isFloating():
             self.ui.close()
+
+    def detectors_changed(self):
+        self.detectors = HexrdConfig().detector_names
+        self.detector = self.detectors[0]
+        # update the state
+        self.state['dets'] = self.detectors
+        for det in self.detectors:
+            self.state[det] = {
```
psavery, line 449: On a note separate to this PR: I see that we are setting the detector keys directly on `self.state` along with other keys [here](https://github.com/HEXRD/hexrdgui/blob/70fe5531afc768f3388a659ab5cd4df409e6db36/hexrd/ui/image_stack_dialog.py#L102-L116). If any detector names (which are user chosen) were to ever clash with these other names on `self.state`, it would cause a bug that would be hard to track down. So I'd recommend nesting the detector keys inside something else, like `self.state['det_settings'][det_name]` to avoid the possible clash. But I know the clash is unlikely and is therefore not urgent and doesn't need to be fixed on this PR.

hexrd/ui/image_stack_dialog.py
```diff
@@ -208,10 +209,11 @@ def total_frames(self):
...
     def change_detector(self, det):
-        self.detector = det
-        self.setup_gui()
-        if self.state['apply_to_all']:
-            self.state[det]['search'] = self.ui.search_text.text()
+        if det:
+            self.detector = det
+            self.setup_gui()
+            if self.state['apply_to_all']:
+                self.state[det]['search'] = self.ui.search_text.text()
```
psavery, line 216: I personally would just do:
```python
if not det:
    return
```

to perform an early return rather than having to add extra indentation to the subsequent lines of code. But it's a style thing and doesn't need to be changed for this PR.

psavery (commented):
I gave this a try and it seems to work. Could you fix the code duplication comment and I will approve it?

hexrd/ui/image_stack_dialog.py
bnmajor: Yes, sounds good. Moved into its own function `setup_detectors_state`

psavery (approved):
LGTM

hexrd/ui/image_stack_dialog.py
bnmajor: That's a fair point. Not for this PR like you mentioned, but this would be a good change. Along with that it would probably be smart to then move the detector settings (and others) into properties to make things simpler and cleaner.

hexrd/ui/image_stack_dialog.py
bnmajor: I just went ahead and changed the check to your suggestion to keep things cleaner.

PR #1189, opened 2022-04-11
Ensure HexrdConfig().has_images returns a bool

psavery:
It appears that in some cases, it was returning a dict, and when we used it as
the argument for QWidget.setEnabled(), Qt would complain that the function only
accepts booleans, but was provided with a dict.

Ensure that it returns a bool so that this error is avoided.


joelvbernier (approved):

joelvbernier:
LGTM

Issue #1190, opened 2022-04-14
Angular periods must be made consistent in the indexing results dialog


joelvbernier:
@psavery -- quick fix: I noticed that the indexing results plotter is not grabbing the periods for η and ω.  It should get these (well, the  latter, most critically) from the ranges on the actual maps (which get them from the imageseries...).  The period for ω should be adjusted to 
```
eta_ome.omegaEdges[0] + np.radians([0, 360])
```
here is an example with ω∊[0, 2π]; clearly the angles calculation is being done from [-π, π].
![image](https://user-images.githubusercontent.com/1154130/163485313-652bb761-788a-4682-bf72-971286eb7887.png)

psavery:
@joelvbernier That's right! I was leaving the `ome_period` to be the default argument [here](https://github.com/HEXRD/hexrd/blob/6129cfeea8a8ff5fcc6bd655e9e54a21f65dd616/hexrd/instrument.py#L1603-L1607).

I just fixed it, and I get better results. You can see an image in #1193.

However, I don't see an argument anywhere for the eta period. It doesn't look like there is one on 
`HEDMInstrument.simulate_rotation_series()`. Where is this specifically? And should I set it to this:

```python
eta_ome.etaEdges[0] + np.radians([0, 360])
```

psavery:
Actually, I see now how to use the eta period, and I have added its usage to #1193. However, it seems unnecessary in the present state because it appears that the eta omega maps themselves are not generated with the user-specified eta period (but always come out with an eta period of -pi to pi). Perhaps this is something that needs to be fixed as well?

PR #1193, opened 2022-04-15
Specify ome_period in indexing results spots

psavery:
It was previously being set to the default, which was [-π, π]. Now,
it is being set based upon the first omega edge value in the eta
omega maps. This appears to generate results correctly.

example

Fixes: #1190


psavery:
I have added the usage of the eta period to this PR. However, it seems unnecessary in the present state because it appears that the eta omega maps themselves are not generated with the user-specified eta period (but always come out with an eta period of -pi to pi). Perhaps this is something that needs to be fixed as well?

joelvbernier:
Since the maps are designed to always span 360° in η, I think we can leave it...  until it becomes an issue.

joelvbernier:
LGTM.

joelvbernier (approved):

Issue #1197, opened 2022-04-20
Do not load all materials on start-up


psavery:
We should change the behavior to only load in one default material, maybe CeO2. This will significantly speed up start-up, as the computations when we first generate the list of materials is expensive.

Then, we can perhaps add something like a button called "Import from Defaults" in the material list editor:

![image](https://user-images.githubusercontent.com/9558430/164130945-00cdeff7-4324-45ca-8a56-263c185e2c04.png)

That will allow the user to select one of the default materials in the HDF5 file and load it into the GUI.

This way, hexrdgui will only start with one material by default, but the user will still retain the abillity to import all of the previous materials through that button.

joelvbernier:
Endorse an Import from defaults option in the add menu.

PR #1203, opened 2022-04-27
Only load the "CeO2" material by default

psavery:
This modifies hexrdgui to only load the "CeO2" material by default,
which improves start-up time because some computations that are ran
when the materials are created take up some time.

The other default materials are available in the material list editor
via a "Import from Defaults" button. If the user clicks this button,
they may select other default materials to import.

If the program exits cleanly, the names of the default materials that
were loaded get saved, so that when hexrdgui is restarted, these default
materials will also be loaded again. Additionally, if any overlays were
saved that are using default material names, these default materials will
also be loaded on start-up.

Fixes: #1197

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -243,7 +243,7 @@ class HexrdConfig(QObject, metaclass=QSingleton):
         self.load_calibration_flags_order()
 
         # Load the default materials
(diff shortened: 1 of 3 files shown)


joelvbernier (approved):

joelvbernier:
looks good!

PR #1207, opened 2022-04-28
Add safety check for drawing overlays in raw view

psavery:
This prevents an exception when everything hasn't been synced up yet.


bnmajor (approved):
LGTM!

PR #1210, opened 2022-05-03
Move hand picked fibers into standalone widget

psavery:
This widget will be re-used in the future for Laue spots on normal views.

Currently, it does not support that, but moving the logic into a
standalone widget is a necessary first step.


joelvbernier (approved):

joelvbernier:
@psavery -- one strange behavior: the selector on the eta-omega widget doesn't affect any changes unless I switch focus to the main window, then re-focus on the eta-omega widget.  Thoughts?

joelvbernier (commented):
@psavery -- one strange behavior: the selector on the eta-omega widget doesn't affect any changes unless I switch focus to the main window, then re-focus on the eta-omega widget.  Thoughts?

psavery:
> @psavery -- one strange behavior: the selector on the eta-omega widget doesn't affect any changes unless I switch focus to the main window, then re-focus on the eta-omega widget. Thoughts?

Is this on Mac? It sounds like that SDK issue we were running into on Mac a while back.

joelvbernier:
On Linux!

PR #1212, opened 2022-05-07
changed labels, defaults, units

joelvbernier:
the tolerance for fit tth chaged to a relative percent error; used to be in degrees.  Updated labels and defaults in the ui.


psavery (commented):
LGTM

Issue #1216, opened 2022-05-13
GUI startup error


joelvbernier:
Both Darren and I started seeing this error after the merge of PR #1208.  I am getting it after loading a session with a single material (CeO2 in this case) and quitting the GUI.  Upone trying to reopen, we get the following exception:
```
Traceback (most recent call last):
  File "/home/jbernier/miniconda3/envs/hexrdgui-dev/bin/hexrdgui", line 33, in <module>
    sys.exit(load_entry_point('hexrdgui', 'gui_scripts', 'hexrdgui')())
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/main.py", line 31, in main
    window = MainWindow()
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/main_window.py", line 79, in __init__
    self.progress_dialog = ProgressDialog(self.ui)
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/progress_dialog.py", line 13, in __init__
    self.messages_widget = MessagesWidget(self.ui)
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/messages_widget.py", line 41, in __init__
    self.capture_output()
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/messages_widget.py", line 67, in capture_output
    self.update_logging_streams()
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/messages_widget.py", line 85, in update_logging_streams
    HexrdConfig().logging_stdout_stream = sys.stdout
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/singletons.py", line 23, in __call__
    cls._instances[cls] = super().__call__(*args, **kwargs)
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/hexrd_config.py", line 226, in __init__
    self.load_settings()
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/hexrd_config.py", line 400, in load_settings
    self.load_from_state(state)
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/hexrd_config.py", line 356, in load_from_state
    setattr(self, name, value)
  File "/home/jbernier/Documents/GitHub/hexrdgui/hexrd/ui/hexrd_config.py", line 411, in _imported_default_materials
    self.load_default_material(x)
[...]

PR #1218, opened 2022-05-13
Ensure the _imported_default_materials value is list

cjh1:
A list with a single value will come back from QSettings as a str we need to check for this a convert to a list.


cjh1:
Fixes: #1216

cjh1:
@joelvbernier This should resolve your issue, let me know if it works for you.

Issue #1142, opened 2022-01-27
Hand picker tool re-appears at the end for tardis example


psavery:
After hand picking both the powder and laue overlays for our tardis example, after the calibration finishes, the hand picking tool re-appears at the end for some reason even though we aren't currently hand picking.

![image](https://user-images.githubusercontent.com/9558430/151397198-977de8bf-4704-42bb-a4f9-a8a559608e07.png)

 It doesn't do anything (because hand picking is over), and if you close the dialog, you get an error like the following:

```python
Traceback (most recent call last):
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/line_picker_dialog.py", line 262, in reject
    self.clear()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/line_picker_dialog.py", line 127, in clear
    self.zoom_canvas.cleanup()
AttributeError: 'NoneType' object has no attribute 'cleanup'
```

It's a relatively benign bug, but it'd be nice to clean up.

Issue #1107, opened 2021-12-23
Fixes to existing calibration methods


joelvbernier:
Hey @psavery -- I noticed that a few things were 'off' in the calibration, so I started a fix branch under PR #1106.

Testing so far looks good, but there are still a few things that need to be addressed:

- [x] The lattice parameter(s) are not getting updated on the material in the "fast powder" method even if they are flagged for refinement; it echos back the changed values, but that must be on a copy of the material's `PlaneData`, and so there is no update of lattice parameters once the optimization is finished.
- [x] The mixed Laue/Powder calibration is still buggy.  Seems to behave badly if points are deleted (could not reproduce - may be fixed by refactoring, such as in #1223. Re-open if still an issue)
- [x] We need to be able to store the picks so that subsequent calibrations can be re-run without haveing to re-pick (fixed by #1223)

joelvbernier:
@psavery -- let's try to connect on this before next week's meeting.

psavery:
@joelvbernier Sounds good

psavery:
> * [ ]  The mixed Laue/Powder calibration is still buggy.  Seems to behave badly if points are deleted.

So, I actually just ran through the simulated tardis example (it has one powder and one laue overlay), and I compared
the difference (in the picks data, which is input to the calibration methods) between skipping points (not picking at all) 
and picking then deleting. It looks like they are the same, so it looks to me that the deleting behavior is working as
intended (if you delete a point, it is as if you skipped it).

However, maybe you are doing something different from what I did. If you can reproduce the issue, maybe you can
change [this line](https://github.com/HEXRD/hexrdgui/blob/ace51145e34f15809d0c27e4de3f3142d0dd7f21/hexrd/ui/calibration/calibration_runner.py#L342) to be `self.dump_results()`, and then run it again? If you do that, it will dump basically all of the
input to the calibration methods, and we can take a closer look to see if the input to the calibration methods is correct.
This will help us determine whether the problem is in the input to the calibration methods, or in the calibration methods
themselves.

> * [ ]  We need to be able to store the picks so that subsequent calibrations can be re-run without haveing to re-pick.

You mean storing them internally, right?

We can actually already do via files. Basically, when you run through the calibration, you can at any point click 
"Picks Table" to view the picks table, and then you can save/load your picks. But storing them internally would
be faster.

PR #1223, opened 2022-05-20
Store calibration picks in overlays

psavery:
This refactors the calibration runner to store the picks in the overlays. This also adds a few new features that are now possible from the refactor. New features are summarized as follows:

1. Overlay picks are now saved/loaded in state files.
2. The user may view the overlay picks from the main menu via "View"->"Overlay Picks". Here, they may also be edited, exported, and imported.
3. In the "Laue and Powder" calibration, there are now 4 options for picking each overlay: Current (just use what the overlay has currently), Auto, Hand, and Load (load the picks for this overlay from a file).

An example video of some of the functionality can be seen below.

https://user-images.githubusercontent.com/9558430/169612119-53bc0234-a2f8-42bd-8ef6-8a3dd9d7cbf5.mp4

One current limitation is that the overlay picks may only be viewed from the main window while in the polar view.

Fixes: #1107
Fixes: #1142

### DIFF
--- a/hexrd/ui/calibration/calibration_runner.py
+++ b/hexrd/ui/calibration/calibration_runner.py
@@ -234,7 +234,6 @@ class CalibrationRunner(QObject):
         # Set the new picks on the overlay
         updated_picks = tree_format_to_picks(results)
         overlay.calibration_picks = updated_picks[0]['picks']
-
         self.use_current_pick_points()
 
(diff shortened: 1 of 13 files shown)


joelvbernier:
Great!  I think we should cut a new release before merging into master.  I'll take a crack at testing today.

joelvbernier (approved):

Issue #1227, opened 2022-05-26
Windows state files are incompatible with Mac/Linux


psavery:
Since pathlib was used for some paths inside the HDF5 file, windows uses `\` for the paths, and Mac/Linux use `/` for the paths.

Fix the state files to always use `/` for the paths so that state files may be transferred between all operating systems.

PR #1228, opened 2022-05-26
Remove use of Pathlib in state files

psavery:
Pathlib was previously being used to refer to paths within the HDF5 state file. This meant that on Windows, paths would use `\` instead of `/`.

Fix up the state file code so that `/` is always used, and add automatic fixing of old state files that have the Windows issue. We now check for the Windows issue upon reading any state file, and if the issue is found, it is fixed automatically.

An example state file with the issue is attached.

[windows_broken.h5.gz](https://github.com/HEXRD/hexrdgui/files/8778869/windows_broken.h5.gz)

Fixes: #1227

### DIFF
--- a/hexrd/ui/state_compatibility.py
+++ b/hexrd/ui/state_compatibility.py
@@ -1,5 +1,7 @@
 import h5py
 
+from hexrd.ui.hexrd_config import HexrdConfig
+
 
 def update_if_needed(file_path):
     # Find and fix any issues with the state file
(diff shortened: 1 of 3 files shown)


hexrd/ui/state_compatibility.py
```diff
@@ -0,0 +1,43 @@
...
+    # file instead of `/`
+    with h5py.File(file_path, 'r') as rf:
+        return any(x.startswith('config\\') for x in rf.keys())
+
+
+def fix_issue_1227(file_path):
+    if not has_issue_1227(file_path):
+        return
+
+    print('State file found to contain issue #1227. Fixing it up...')
```
cjh1, line 20: We should probably try to switch all our `print`'s  to log statement sometime, I know we have discussed this.

cjh1 (approved):

Issue #1229, opened 2022-05-26
Split the lorentz polarization correction to two options


saransh13:
Split the Lorentz polarization correction in intensity correction menu to two options: (1) Lorentz correction and (2) polarization correction. This option will let users integrate liquid diffraction signals where only polarization correction needs to be applied without the Lorentz correction.

The function to compute this in `hexrd.instrument` will need to be modified, along with the relevant changes to `hexrdgui` functions.

PR #1230, opened 2022-05-26
LP split

saransh13:
Split the Lorentz and polarization factor intensity corrections to two different options. Tested with some real liquid diffraction profiles and seems to work with old state files.

Fixes: #1229

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -1048,13 +1048,6 @@ class MainWindow(QObject):
         HexrdConfig().apply_polarization_correction = b
 
     def apply_lorentz_correction_toggled(self, b):
-        if not b:
-            # Just turn it off and return
-            HexrdConfig().apply_lorentz_correction = b
-            return
-
-        # The dialog should have modified HexrdConfig's Lorentz options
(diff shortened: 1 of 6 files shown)


hexrd/ui/main_window.py
```diff
@@ -1025,23 +1029,33 @@ def enforce_view_mode(self, raw_only):
...
+    def apply_lorentz_correction_toggled(self, b):
+        if not b:
+            # Just turn it off and return
+            HexrdConfig().apply_lorentz_correction = b
+            return
+
         # The dialog should have modified HexrdConfig's Lorentz options
         # already. Just apply it now.
-        HexrdConfig().apply_lorentz_polarization_correction = b
+        HexrdConfig().apply_lorentz_correction = b
```
psavery, line 1058: I think this whole function can just be `HexrdConfig().apply_lorentz_correction = b`

psavery (approved):

hexrd/ui/main_window.py
saransh13: Aah yes, I'll push these changes.

PR #1231, opened 2022-05-27
Skip missing keys in the config status setter

psavery:
There can sometimes be a mismatch, so just skip missing keys.


bnmajor (approved):
LGTM!

Issue #1235, opened 2022-05-31
Powder overlay issue


saransh13:
Powder overlay not showing up on one of the TARDIS panels.

![Screen Shot 2022-05-31 at 12 33 53 PM](https://user-images.githubusercontent.com/15834451/171270325-ec7ff32b-5979-4e11-a06b-bde82388c84e.png)

psavery:
@saransh13 So I think you must have run into a situation where the eta period in the overlay are not being synchronized with the eta period in the hexrdgui settings. Do you know the exact steps you took to cause this?

Then again, it might be worth considering for us to refactor this in a way so that the overlay always uses the eta period in the hexrdgui settings, so that there will be no synchronization issue.

psavery:
@saransh13 Even though I was unable to reproduce your issue, I believe #1236 should fix it. Could you confirm with me that fixes the issue for you?

saransh13:
Yes, it does fix the issue! PR LGTM

PR #1236, opened 2022-05-31
Always use eta period directly from HexrdConfig()

psavery:
This avoids synchronization issues that we have been running into, such as
in #1235. Now, the eta period for the overlays can never be out of sync with
the eta period for the view.

Fixes: #1235


psavery:
Approved by @saransh13 in #1235.

PR #1243, opened 2022-06-08
Ensure raw masks are used for masked images dict

psavery:
If we are in the polar view, then the `HexrdConfig().masks` will be
polar masks instead of raw masks. Ensure that they are raw masks instead.


psavery:
@saransh13 approved this verbally

PR #1244, opened 2022-06-09
Migrate PowderCalibrator to the one in hexrd

psavery:
We were previously using a custom PowderCalibrator class for the
"Laue and Powder" calibration in hexrdgui. However, this class is
not maintained and no longer works properly.

Migrate over to using the PowderCalibrator from hexrd. This one
is more maintained and updated.

The migration required modifying the data input, since the
PowderCalibrator from hexrd expects the input data to be in a different
format.

There may be some options we should be using and are not using yet.

### DIFF
--- a/hexrd/ui/calibration/calibration_runner.py
+++ b/hexrd/ui/calibration/calibration_runner.py
@@ -373,13 +373,15 @@ class CalibrationRunner(QObject):
         flags = HexrdConfig().get_statuses_instrument_format()
         instr.calibration_flags = flags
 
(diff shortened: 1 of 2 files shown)


psavery:
@joelvbernier I believe this should be working properly now.

There are some very slight differences in the results when compared to the "Auto Powder" method, but I confirmed these slight differences are only caused by a small loss of precision when converting between polar and Cartesian in the "Laue and Powder" workflow.

joelvbernier (approved):

PR #1245, opened 2022-06-13
Bump Python to 3.10

cjh1:


cjh1:
I have tested the packages on all three platforms and things seem to work well with 3.10

psavery (approved):
LGTM

PR #1246, opened 2022-06-14
Masks to panel buffer

### LINKED ISSUE #1226: Add ability to set detector panel buffer from masks.
We could use a 3rd option for configuring the panel buffer on the detector: set from masks.  I have lost track of the logic of how the various masking options are handled @bnmajor -- but we could make it explicit regarding setting the `panel_buffer` attribute on each detector with the masks defined on that detector.

bnmajor:
Fixes #1226

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -729,8 +729,8 @@ class HexrdConfig(QObject, metaclass=QSingleton):
     def masked_images_dict(self):
         """Get an images dict where masks have been applied"""
         images_dict = self.images_dict
-        for name, img in images_dict.items():
-            for det, mask in self.raw_masks_dict.items():
+        for det, mask in self.raw_masks_dict.items():
+            for name, img in images_dict.items():
                 if det == name:
                     img[~mask] = 0
 
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -699,8 +699,8 @@ class HexrdConfig(QObject, metaclass=QSingleton):
         return self.intensity_corrected_images_dict
 
     @property
-    def masked_images_dict(self):
-        """Get an images dict where masks have been applied"""
+    def masks_dict(self):
+        """Get a masks dict"""
         # We must ensure that we are using raw masks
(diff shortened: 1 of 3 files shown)


hexrd/ui/hexrd_config.py
```diff
@@ -699,8 +699,8 @@ def images_dict(self):
         return self.intensity_corrected_images_dict
 
     @property
-    def masked_images_dict(self):
-        """Get an images dict where masks have been applied"""
+    def masks_dict(self):
```
psavery, line 702: Maybe we should rename this to `raw_masks_dict` to clarify that it only returns raw masks?

hexrd/ui/hexrd_config.py
```diff
@@ -710,15 +710,29 @@ def masked_images_dict(self):
...
+            masks_dict[name] = final_mask
+
+        return masks_dict
+
+    @property
+    def masked_images_dict(self):
+        """Get an images dict where masks have been applied"""
+        images_dict = self.images_dict
+        for name, img in images_dict.items():
+            for det, mask in self.masks_dict.items():
```
psavery, line 733: Can we set `masks_dict = self.masks_dict` outside of the loops so it doesn't rebuild the masks dict for every iteration of the images dict?

hexrd/ui/mask_manager_dialog.py
```diff
@@ -306,3 +308,14 @@ def rebuild_masks(self):
             HexrdConfig().polar_masks_changed.emit()
 
         self.update_masks_list('raw')
+
+    def masks_panel_to_buffer(self):
```
psavery, line 312: Shouldn't this be named `masks_to_panel_buffer`?

psavery:
I tried it locally and it seems to be working.

psavery (approved):
Works for me!

hexrd/ui/hexrd_config.py
bnmajor: Yes! Fixed.

hexrd/ui/mask_manager_dialog.py
bnmajor: Oops, yes it should have been, thanks! Fixed the naming

PR #1248, opened 2022-06-15
Add option to clear panel buffer

### LINKED ISSUE #1247: Add the ability to remove the panel buffer
Taken from #1226

bnmajor:
Fixes #1247

### DIFF
--- a/hexrd/ui/calibration/panel_buffer_dialog.py
+++ b/hexrd/ui/calibration/panel_buffer_dialog.py
@@ -35,6 +35,7 @@ class PanelBufferDialog(QObject):
     def setup_connections(self):
         self.ui.select_file_button.pressed.connect(self.select_file)
         self.ui.config_mode.currentIndexChanged.connect(self.update_mode_tab)
+        self.ui.clear_panel_buffer.clicked.connect(self.clear_panel_buffer)
         self.ui.accepted.connect(self.on_accepted)
         self.ui.rejected.connect(self.on_rejected)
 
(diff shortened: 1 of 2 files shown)


hexrd/ui/calibration/panel_buffer_dialog.py
```diff
@@ -160,3 +161,18 @@ def mode(self, v):
     def update_mode_tab(self):
         mode_tab = getattr(self.ui, self.mode + '_tab')
         self.ui.tab_widget.setCurrentWidget(mode_tab)
+
+    def clear_panel_buffer(self):
+        # Clear the config options on the internal config
+        config = HexrdConfig().config
+        detector_config = config['instrument']['detectors'][self.detector]
```
psavery, line 168:
```suggestion
        detector_config = HexrdConfig().detector(self.detector)
```

hexrd/ui/calibration/panel_buffer_dialog.py
```diff
@@ -160,3 +161,18 @@ def mode(self, v):
...
+
+        buffer_default = {'status': 0}
+        buffer = detector_config.setdefault('buffer', buffer_default)
+
+        # Must match the detector size
+        detector_shape = (detector_config['pixels']['rows']['value'],
+                          detector_config['pixels']['columns']['value'])
+
+        empty_mask = np.ones(detector_shape, dtype=bool)
+        buffer['value'] = empty_mask
```
psavery, line 178: I believe that we can just set `buffer['value'] = [0., 0.]` to nullify it, or even `buffer['value'] = None` (but we would need to make sure we are handling the `None` case properly in the GUI - I think we are not, even though it is valid in hexrd).

hexrd/ui/calibration/panel_buffer_dialog.py
bnmajor: Perfect, fixed.

hexrd/ui/calibration/panel_buffer_dialog.py
bnmajor: Much cleaner, thank you. Fixed to use the `buffer['value'] = [0., 0.]` option.

hexrd/ui/calibration/panel_buffer_dialog.py
```diff
@@ -160,3 +161,12 @@ def mode(self, v):
...
         self.ui.tab_widget.setCurrentWidget(mode_tab)
+
+    def clear_panel_buffer(self):
+        # Clear the config options on the internal config
+        detector_config = HexrdConfig().detector(self.detector)
+
+        buffer_default = {'status': 0}
+        buffer = detector_config.setdefault('buffer', buffer_default)
+
+        buffer['value'] = buffer['value'] = [0., 0.]
```
psavery, line 172: Uh oh, looks like `buffer['value'] =` is there twice?

psavery (approved):
Once my last comment is addressed, then LGTM

hexrd/ui/calibration/panel_buffer_dialog.py
bnmajor: Fixed!

Issue #1198, opened 2022-04-20
Add option to display direct beam mark.


joelvbernier:
The `instrument.PlanarDetector` class has a `beam_position` attribute.  It returns the cartesian (x, y) coordinates of the intersection of the direct beam with its (extended) plane (NaNs if no intersection exists).  It would be helpful to add a toggle to plot a glyph on the various views that denotes the direct beam position.  Some details:

- probably should clip the beam position to the detector extents for the purpose of plotting
- will want a marker style editor to set preferences
- will need to convert to angles for polar view
- will need to loop over all detectors to find the ones that have an intersection.  Probably should cull duplicates for the cartesion and polar views in the case of detectors with overlapping coverage.

PR #1249, opened 2022-06-17
Add option to display the beam marker

psavery:
This adds an option in the main window to display the beam marker, that
defaults to `True`.

If `True`, the beam marker will be drawn in all three views.

The user may edit the style of the beam marker by toggling it on. This
makes the style editor for the beam marker appear.

Here is the style editor (it appears when you toggle the beam marker on):
image

And here are some example images of the beam marker in action:

# Cartesian (GE)
image

# Raw (pxrdip)
pxrdip_raw

# Polar (pxrdip)
pxrdip_polar

Fixes: #1198


psavery:
My latest commit makes it so it is only drawn if it intersects with the detector panel. I think this PR is ready for review.

joelvbernier (approved):

Issue #769, opened 2021-02-25
Add callback to remove duplicate atoms in material structure panel


joelvbernier:
@saransh13 material class has the ability to use the space group to detect and remove duplicate atoms in the unit cell -- this is important to get the density and structure factors correct.  It is also easy to accidentally add a duplicate site by hand, especially if you are unknowingly using an unsymmetrized structure.

So... we need a "Remove duplicate sites" button to call this functionality, which with update the site list for the material.  AFAIK it will need a tolerance (single scalar in Å) as a threshold to determine duplicate positions (@saransh13?).

psavery:
Sounds good.

saransh13:
PR https://github.com/HEXRD/hexrd/pull/196 added this functionality in heard. I discovered a small bug in there which will be fixed in an upcoming PR.

joelvbernier:
@cjh1 -- function exists; need a button in the materials --> structure tab.  @saransh13 can provide the fucntion API.

PR #1250, opened 2022-06-17
Add button to remove duplicate atoms

psavery:
This is in the material structure editor.

Fixes: #769

### DIFF
--- a/hexrd/ui/material_structure_editor.py
+++ b/hexrd/ui/material_structure_editor.py
@@ -67,6 +67,8 @@ class MaterialStructureEditor(QObject):
         self.ui.table.selectionModel().selectionChanged.connect(
             self.selection_changed)
 
+        self.ui.remove_duplicate_atoms.clicked.connect(
+            self.remove_duplicate_atoms)
         self.ui.apply.pressed.connect(self.update_material)
         self.ui.reset.pressed.connect(self.update_gui)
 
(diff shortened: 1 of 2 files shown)


hexrd/ui/material_structure_editor.py
```diff
@@ -304,7 +316,7 @@ def coords_equal(v1, v2, tol=1.e-5):
 
             U_array = new_U_array
 
-        for i, atom in enumerate(info_array):
+        for i, atom in enumerate(pos_array):
```
saransh13, line 319: I'm confused what Line 319-350 is for? Is it checking if two sites have the same atoms?

hexrd/ui/material_structure_editor.py
```diff
@@ -291,8 +303,8 @@ def coords_equal(v1, v2, tol=1.e-5):
                 all(abs(x - y) < tol for x, y in zip(v1, v2))
             )
 
-        info_array = mat._atominfo
-        type_array = mat._atomtype
+        pos_array = mat.atom_pos
+        type_array = mat.atom_type
         charge_array = mat.charge
```
saransh13, line 308: The charge also needs to be synced. That's part of PR https://github.com/HEXRD/hexrd/pull/450

saransh13 (approved):
A couple of comments but PR works looks good.

hexrd/ui/material_structure_editor.py
psavery: It is actually checking if two atoms have a matching site. This is for generating the GUI based off of what is in the `Material` class. It is because the GUI was designed to have the user select a site and then view/edit atoms in that particular site. So it compares the coordinates to group atoms into different sites.

hexrd/ui/material_structure_editor.py
psavery: Sounds good, thank you.

Issue #1232, opened 2022-05-27
xy file writer


saransh13:
Include ability to export polar view line outs (azimuthal integration) as simple `.xy` text files. Can use `np.savetxt` for this with the extension `.xy`

psavery:
I think I can do this at the same time as #1188

[Here](https://github.com/HEXRD/hexrdgui/blob/35f0fab8f32226e9a0bfec2496374b6a4e7fd4a1/hexrd/ui/calibration/polar_plot.py#L93-L137) is the current function in use. The difference with the `.xy` and `.csv` writers, though, will be that they will only write out the lineout (the other formats write out a bunch of other data, too, including the masks, snip background, extent, coordinates, and intensities).

saransh13:
Ok, maybe the writer can pop up an option to select which data to write?
That could give flexibility in include everything or just the tth-intensity
values.

Just the tth-intensity in the .xy format is useful since that is an input
to a number of other analysis programs. Let me know if you think this makes
sense.

Issue #1188, opened 2022-04-11
Add option to save polar view lineout to csv


joelvbernier:
Several users have expressed the desire for an option to export the lineout in the polar view to csv.  We could probably add the option to save either the lineout _or_ the 2-d polarview (as displayed) in the option.  Anything to add @darrencpagan?

joelvbernier:
I think we will want 2 text-based options:

1. Save the lineout to txt
  - [ ] spec header
  - [ ] add GUI export menu item option
2. Save polar image to Maud esg format
  - [ ] formalize @joelvbernier's function
  - [ ] add GUI exoprt menu option

psavery:
Do we have a function for exporting to Maud ESG format?

joelvbernier:
I do, and I should add that!

PR #1251, opened 2022-06-17
Add ability to save lineout as xy or csv

psavery:
If the image mode is "Polar", and the user selects
"File" -> "Export" -> "Current View", and they
name their file with a ".xy" or ".csv" extension, the azimuthal
integral plot is written in XY or CSV format, respectively. This
follows the previous pattern where the user could write out to NPZ
format by specifying a ".npz" extension.

However, there may certainly be some room to make this more visible
or user friendly. I'm open to suggestions. Maybe right-clicking the
azimuthal integral plot in the polar view should bring up a context
menu with the option to export it, and then they choose which file
format (XY or CSV)? They could likewise do this for exporting other
views.

This PR is also a good first step.

Fixes: #1188
Fixes: #1232

### DIFF
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -99,6 +99,22 @@ class InstrumentViewer:
         # Re-format the data so that it is in 2 columns
         azimuthal_integration = np.array(azimuthal_integration).T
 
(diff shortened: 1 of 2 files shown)


bnmajor (approved):
LGTM!

PR #1257, opened 2022-07-19
Fix non-responding widgets for multiple indexing runs

psavery:
The previous design pattern created a list of `QSignalBlocker` objects,
and it relied on Python to perform reference counting and delete this list
at the end of the function, so that the widgets would become unblocked.

For some reason, if you run indexing multiple times, it appears that this
list is sometimes not deleted at the end of the function.

Instead of relying on Python to perform reference counting correctly,
let's use a context manager so that we can be sure the widgets get
unblocked.

This fixes an issue where widgets were not emitting signals in the eta
omega maps viewer when running the HEDM workflow multiple times.

We should also go back and fix any other cases of the design pattern, since
it appears to be error-prone.

### DIFF
--- a/hexrd/ui/calibration_crystal_editor.py
+++ b/hexrd/ui/calibration_crystal_editor.py
@@ -2,7 +2,7 @@ import copy
 import numpy as np
 from numpy.linalg import LinAlgError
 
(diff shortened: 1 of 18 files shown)


cjh1 (approved):
LGTM

PR #1259, opened 2022-07-22
Add ability to delete grains in some dialogs

psavery:
This adds the ability to delete grains in the fit grains options
dialog (before fit grains is ran), and in the fit grains results
dialog (after fit grains is ran).

The user may delete grains by right-clicking them in the table and
clicking "Delete grain(s)". Note that it is currently set up to not
allow the user to delete all grains.

### DIFF
--- a/hexrd/ui/indexing/fit_grains_options_dialog.py
+++ b/hexrd/ui/indexing/fit_grains_options_dialog.py
@@ -18,6 +18,7 @@ from pathlib import Path
 class FitGrainsOptionsDialog(QObject):
     accepted = Signal()
     rejected = Signal()
+    grains_table_modified = Signal()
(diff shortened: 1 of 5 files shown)


psavery:
NOTE: we need to make sure hexrd's fit-grains will work for discontinuous grain ids (for instance, there are 3 grains, and the user deletes grain id 1, so there are grain ids 0 and 2 left).

joelvbernier:
@psavery -- here is a proposed interim solution; upon deletion, can we just renumber the remaining grains in the table sequentially in the GUI form?

joelvbernier (approved):

psavery:
@joelvbernier The fix seems to be working great!

Issue #1260, opened 2022-08-08
Reconcile EtaOmeMaps API change


joelvbernier:
# NOTE: this is only necessary after https://github.com/HEXRD/hexrd/pull/454 is merged!

@psavery -- since the `EtaOmeMap.iHKLList` now contains the intrinsic IDs (that is, as generated) this 
https://github.com/HEXRD/hexrdgui/blob/a3e45bc324d14560cf574cde5bdb8f28dffd803d/hexrd/ui/indexing/create_config.py#L28
should be modified to be
```
omaps['active_hkls'] = material.planeData.getHKLID(material.planeData.getHKLID.getHKLs().T)
```
Although the `getHKLID` method should just be modified to act this way if no arg is given...

@psavery -- we will at least also need to check these areas as well:
```
calibration/calibrationutil.py:129:        iHKLLists = []
calibration/calibrationutil.py:146:                iHKLLists.append(hklsCur)
calibration/calibrationutil.py:150:        return iHKLLists, mergedRanges
indexing/indexing_results_dialog.py:95:        hkl_indices = self.ome_maps.iHKLList
indexing/ome_maps_viewer_dialog.py:419:        hkl_indices = self.data.iHKLList
```

PR #1264, opened 2022-08-09
Update new active hkl indexing for eta omega maps

psavery:
We now use master hkl indices for the active hkl indexing in eta omega maps.
Update hexrdgui to reflect that.

Fixes: #1260
Depends on: hexrd/hexrd#454


cjh1 (approved):
LGTM

Issue #1270, opened 2022-08-24
Allow for specification of beam vector by components.


joelvbernier:
To enhance the UX and ameliorate the complexity of the spherical angles, it is desired to have the option to set the instrument beam vector via vector components.  Some details:

1. Enable the new input option in the **Form View** of the instrument configuration widget (3 float spin boxes)
2. Additionally include a checkbox to set the `HEDMInstrument.source_distance` property from the input vector magnitude; the `HEDMInstrument.beam_vector` is strictly unit magnitude, so the input needs normalization at any rate.

The associated callback for normalizing the input would be:
```
from hexrd.transforms import xfcapi

# if we need the magnitude
# !!! input_vec is array_like (3, )
if set_source_dist:
    instr.source_distance = xfcapi.rowNorm(np.atleast_2d(input_vec))

# set beam vector
instr.beam_vector = xfcapi.unitRowVector(np.asarray(input_vec).flatten())

```

Depends on resolution of https://github.com/HEXRD/hexrd/issues/463

PR #1271, opened 2022-08-24
Add beam vector specification by components

psavery:
Fixes: #1270


psavery:
I believe this is now ready, but it depends on hexrd/hexrd#446.

Here is an example video.

To summarize: a Cartesian input mode was added to the beam vector, and a source distance can now be edited as well. When the source distance is disabled, it is set to infinity.

If the source distance is disabled, editing the Cartesian input values results in some text being displayed that show the normalized values (which is what hexrd will receive). If the source distance is enabled, though, the Cartesian input values are used to compute the magnitude. Editing the magnitude also likewise scales the Cartesian input values.

https://user-images.githubusercontent.com/9558430/186761082-1dd7dd99-f9b6-44ad-87b9-9afad927e331.mp4

joelvbernier:
Very nice @psavery !

joelvbernier:
@psavery -- two things: in addition to the "convention" feature, I guess we need to take stock of the full list of PRs that need to be merged before this will work...  setting source distance obviously requires https://github.com/HEXRD/hexrd/pull/446, and I think that is it?  And I think I can merge that without breaking anything else :-) 

The `PowderCalibrator` gets a new kwarg, but it has a default of `None` -- we should probably think about adding a `PowderCalibrator` CI test for the cases with and without distortion at this point given that it is a critical feature.  If I set up two problems and the comparison, can you help map it into GitHub Action-friendly tests?

psavery:
Hi Joel! I added the convention in the latest commit (see attached image).

Yes, I think HEXRD/hexrd#446 is the only PR this depends on.

And yes, I can quickly add a GitHub Actions test if you give me the example!

Screenshot from 2022-08-25 22-20-38

hexrd/ui/resources/calibration/default_instrument_config.yml
```diff
@@ -1,6 +1,7 @@
 beam:
   energy: 71.6764
   vector: {azimuth: 90.0, polar_angle: 90.0}
+  source_distance: .inf
```
joelvbernier, line 4: Strictly shouldn't need this, since the instrument class has the default.  But it doesn't hurt.  You verified that `.inf` is the yaml spec for IEEE infinity?

psavery: Yes I did. It is working.

joelvbernier (approved):

PR #1275, opened 2022-08-30
Add pinhole distortion options to powder overlay editor

psavery:
This adds pinhole distortion options to the powder overlay editor to be used in calibration. These pinhole settings are being added in hexrd in hexrd/hexrd#446.

https://user-images.githubusercontent.com/9558430/187340753-43735eef-e81e-4b5a-ac08-d2e585d739e7.mp4

Depends on: hexrd/hexrd#446
Depends on: #1271
Fixes: #1273
Fixes: #1279

PR #1281, opened 2022-09-05
Panel buffer options

### LINKED ISSUE #1226: Add ability to set detector panel buffer from masks.

bnmajor:
Adds some of the features requested in #1226:

- Can clobber any existing panel buffer
- Can do a logical OR to merge with existing
- Can do a logical AND with existing

This will close #1226 (re-opened)

### DIFF
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -1,4 +1,4 @@
-import copy
+import math
 from hexrd.ui.create_polar_mask import rebuild_polar_masks
 from hexrd.ui.create_raw_mask import rebuild_raw_masks
 import os
@@ -19,6 +19,8 @@ from hexrd.ui.utils import block_signals, unique_name
 from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.ui_loader import UiLoader
 from hexrd.ui.constants import ViewType
+
+import matplotlib.pyplot as plt
 
 
 class MaskManagerDialog(QObject):
@@ -80,6 +82,7 @@ class MaskManagerDialog(QObject):
         self.ui.export_masks.clicked.connect(self.write_all_masks)
         self.ui.import_masks.clicked.connect(self.import_masks)
         self.ui.panel_buffer.clicked.connect(self.masks_to_panel_buffer)
+        self.ui.view_masks.clicked.connect(self.show_masks)
 
         HexrdConfig().mode_threshold_mask_changed.connect(
             self.update_masks_list)
@@ -369,3 +372,17 @@ class MaskManagerDialog(QObject):
             buffer['value'] = mask
         msg = 'Masks set as panel buffers.'
         QMessageBox.information(self.parent, 'HEXRD', msg)
+
+    def show_masks(self):
+        num_dets = len(HexrdConfig().detector_names)
+        cols = 2 if num_dets > 1 else 1
+        rows = math.ceil(num_dets / cols)
+
+        fig = plt.figure()
+        fig.canvas.manager.set_window_title('User Created Masks')
+        for i, det in enumerate(HexrdConfig().detector_names):
+            axis = fig.add_subplot(rows, cols, i + 1)
+            axis.set_title(det)
+            axis.imshow(HexrdConfig().raw_masks_dict[det])
+        fig.canvas.draw_idle()
+        fig.show()
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -313,37 +313,48 @@ class MaskManagerDialog(QObject):
         self.update_masks_list('raw')
 
     def masks_to_panel_buffer(self):
+        show_dialog = False
+        selection = 'Replace buffer'
+        for det in HexrdConfig().detectors.values():
+            buff_val = det.get('buffer', {}).get('value', None)
+            if isinstance(buff_val, np.ndarray) and buff_val.ndim == 2:
+                show_dialog = True
+                break
+
+        if show_dialog:
+            dialog = QDialog(self.ui)
+            layout = QVBoxLayout()
+            dialog.setLayout(layout)
+
+            options = QComboBox(dialog)
+            options.addItem('Replace buffer')
+            options.addItem('Logical AND with buffer')
+            options.addItem('Logical OR with buffer')
+            layout.addWidget(options)
+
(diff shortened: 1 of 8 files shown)


bnmajor:
Need to:
- [ ] Only provide options dialog if 2-D panel buffer already exists
- [ ] For threshold mask use the current frame

joelvbernier (approved):

hexrd/ui/mask_manager_dialog.py
```diff
@@ -310,12 +313,53 @@ def rebuild_masks(self):
...
+
+            buttons = QDialogButtonBox.Ok | QDialogButtonBox.Cancel
+            button_box = QDialogButtonBox(buttons, dialog)
+            button_box.accepted.connect(dialog.accept)
+            button_box.rejected.connect(dialog.reject)
+            layout.addWidget(button_box)
+
+            if not dialog.exec_():
+                # canceled
+                return
```
psavery, line 343: FYI: I think you can pretty much do the same thing here with just [`QInputDialog.getItem()`](https://doc.qt.io/qt-5/qinputdialog.html#getItem). But you don't need to change this if it works.

psavery:
Okay, here is my test of this:

For this input image that is thresholded (where blue is thresholded):

thresholded

And this original panel buffer:

original_panel_buffer

When I apply the threshold masks to the panel buffer, here is what I get for the different scenarios:

Replacement
===========
result_of_replacement

Logical Or
========
result_of_or

Logical And
=========
result_of_and

It looks to me like the "And" and "Or" are reversed. Maybe this is because the panel buffer has 1 set to valid pixels and 0 set to invalid pixels. What do you think, @joelvbernier ?

joelvbernier:
Yes @bnmajor and @psavery -- the `panel_buffer` behaves in the oppsite notice of a mask; valid (unmasked) pixels are `True` and invalid (masked) pixels are `False`.  So if the masks are `True` where masked -- like a numpy masked array -- we need to apply the '`~`' operator when setting the detector `panel_buffer` property with a mask.  @bnmajor can you verify the convention of the mask arrays?  We get that sorted and we can merge this up!

joelvbernier:
Actually, this look ok -- OR is adding the extra strip along with the existing buffer, and AND is only adding the pixels that are `True` in both, hence the salt and pepper look.  LGTM!

psavery (approved):
@joelvbernier confirmed they are correct, so LGTM once all your final edits are pushed @bnmajor

bnmajor:
@joelvbernier This branch has been updated to fix the issues that you were seeing as well as some I noticed. If you and/or @psavery wouldn't mind giving this one more test I think it should be all set now.
- Threshold mask are now applied in Polar View
- Masks were not being toggled on/off in raw view if threshold is visible. This is fixed now (we are no longer updating a copy of the imageseries but rather applying/removing the threshold mask like any other mask).
- Deleting threshold mask when not visible no longer creates error

psavery:
A few things I've noticed that I believe need to be addressed:

- [ ] Toggling the "Set Threshold Mask" checkbox on and off repeatedly does not seem to behave properly. Sometimes, the checkbox is unchecked but the masks are applied, and sometimes, the checkbox is checked but the masks are not applied. The "Visible" checkbox in the mask manager also does not appear to stay in sync with the "Set Threshold Mask" checkbox.
- [ ] The threshold masks appear different in the polar view than in the raw view. This is because the threshold mask is being created after several steps of the polar view processing (image warping, rescaling, and snipping). The pixel values at this point are pretty different than what they were originally, so the thresholding ends up being pretty different as well. I'm wondering if we should perhaps instead create the polar threshold masks by taking the original raw threshold masks and warping them (like is done for the polar images themselves), then applying these. What do you think, @joelvbernier?
- [ ] In the mask manager, if you rename the threshold mask, editing the visibility from the mask manager no longer works properly. Should we disable renaming it, if we are going to treat "threshold" as a key word for the name of the mask? If we do this, we should also not allow other masks to be named "threshold".
- [ ] In the mask manager, if you press the "Remove Mask" button for the threshold mask, it should probably also uncheck "Set Threshold Mask" as well.

joelvbernier:
@psavery -- since the threshold mask is generated in raw image coordinates (pixel coordinates --> cartesian), it should behave the same as the others, which are stored in native cartesian (or pixel) coordinates (right @bnmajor?).  Snipping doesn't affect the masks as it does not alter coordinates, just image intensities.  In theory, masks in raw coordinates get transformed to polar using either the straight `cart_to_angles` on a detector-by-detector basis _or_ using the `warp_image` method on the `PolarView` class.

psavery:
@bnmajor Joel and I talked about converting the threshold masks to polar. I think what we are going to do is to take the raw threshold masks and convert them to the polar image via warping. In #1275, I've added a `PolarView.warp_image()` function that we can use to do this, and I can help set that up after #1275 is merged.

psavery:
LGTM

joelvbernier (approved):

joelvbernier:
The "equal to" option is not giving the correct results.  in the tardis example, equal to 0 is blanking the whole image.

joelvbernier:
Also need to increase the limits of the QScientificSpinbox to ±2**32

joelvbernier:
Also mask merging with panel buffer does not seem to be working -- I did logical OR with the existing and my panel buffers are now uniformly `True`, which is definitely not the case

bnmajor:
> The "equal to" option is not giving the correct results. in the tardis example, equal to 0 is blanking the whole image.

Fixed in [58a4142](https://github.com/HEXRD/hexrdgui/pull/1281/commits/58a4142826acb1ca5f618201948e4b9b782c1283)

> Also need to increase the limits of the QScientificSpinbox to ±2**32

If this is a request for all `QScientificSpinbox`es this should be a separate PR.

> Also mask merging with panel buffer does not seem to be working -- I did logical OR with the existing and my panel buffers are now uniformly True, which is definitely not the case

This was because the mask that was generated was incorrect, not because of the `logical_or` operation. Fixed in [58a4142](https://github.com/HEXRD/hexrdgui/pull/1281/commits/58a4142826acb1ca5f618201948e4b9b782c1283)

@joelvbernier This branch also includes the `View Masks` option now.

masking

PR #1291, opened 2022-09-24
force nsteps to be integer

joelvbernier:
The `nsteps` was coming back as a float from the stack importer in some cases, causing an exception.


psavery:
I'll defer this one to @bnmajor

bnmajor (approved):
LGTM

PR #1303, opened 2022-10-21
Llnl template bugs

### LINKED ISSUE #1300: TARDIS image plate outline disappears when rotating
When I am loading TARDIS image plates and I try to rotate the image, the cyan outline disappears as soon I try to move it. This happens both with the keyboard up/down keys and dragging with the mouse.

### LINKED ISSUE #1302: Overlay badness in LLNL Import Tool
@bnmajor I just noticed this one.  For some reason, if you apply "Transform Image" in the LLNL import tool, some strange things happen
1. The overlay is redrawn differently (see extra line in the boundary; are the nan coordinates that separate the outer contour from the hole removed?)
2. If you attempt to use the mouse to do any rotations, the overlay disappears.
3. Rotation by keys works as expected.

This is in the >=0.9.0 release all platforms.  If you can get to this one it would be much appreciated.

llnl_import_error

bnmajor:
Partially Fixes #1302, but the gist of #1300 is still unresolved.

### DIFF
--- a/hexrd/ui/interactive_template.py
+++ b/hexrd/ui/interactive_template.py
@@ -290,6 +290,9 @@ class InteractiveTemplate:
         contains, info = self.shape.contains(event)
         if not contains:
             return
+        # FIXME: Need to come back to this to understand why we
+        # need to set the press value twice
+        self.press = self.shape.xy, event.xdata, event.ydata
         self.center = self.get_midpoint()
         self.shape.set_transform(self.ax.axes.transData)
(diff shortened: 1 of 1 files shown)


joelvbernier:
Looks like it resolves the polygon part, but the disappearing outline on the mouse rotate _after_ "Add Transform" remains...  Some consolation, is that even though the outline disappears, it seems to still be doing the right thing.  Arrow keys work fine, making it even more bizarre!

PR #1304, opened 2022-10-21
modified key input deltas for crop

joelvbernier:
The previous delta for arrow key input for crop boundary rotation was 0.01 radians (~0.573°); changed to 0.00175 radians (~0.1°).

Likewise the previous delta for translation was 1 pixel; reduced to 0.5.


bnmajor (approved):
LGTM

PR #1309, opened 2022-11-01
added update to total angle from key input

joelvbernier:
The `total_angle` variable wasn't being updated internally even though the detector overly was in the GUI.  Noticed after testing a mixed keyboard and mouse input and getting strange results on finalizing cropping and reloading the instrument in the LLNL Import Tool.  This seems to have fixed the issue (thanks for pointing me in the right direction @bnmajor!).


bnmajor (approved):
Awesome, thanks!

Issue #1310, opened 2022-11-05
Calibration hand picks are *not* being updated with changing instrument parameters


joelvbernier:
I just noticed in 0.9.5 that the pick point (2θ, η) are **_not_** being updated with changing instrument parameters.  The picks are supposed to be stored via their _transformed_ pixel (or equivalently cartesian) coordinates in the raw image via each detector's `angles_to_cart` method, so that if the instrument parameters are changed (_e.g._, following the update after a calibration) the apparent (2θ, η) are updated as well via each detector's `cart_to_angles` method.  The native storage is supposed to be in raw pixel coordinates and the displayed (2θ, η) is calculated from the current instrument parameters.

Instead, the current behavior has the picked (2θ, η) remaining static; if I select "current" picks after a previously executed calibration where the instrument parameters (and hence, the polar view) are updated, the values are unchanged upon review.  The effect is that the picks no longer appear to align with the underlying features.  So either the values are not being refreshed in the picks viewer, or the are not being updated from intrinsic pixel coordinates, or both.

PR #1313, opened 2022-12-20
Store calibration picks in Cartesian coordinates

psavery:
Previously, the calibration picks were being stored in angular coordinates. This was a problem, however, because if the instrument/detector parameters were modified, the picks would not be updated to reflect that change.

This commit changes the calibration picks to be stored in Cartesian coordinates instead (similar to how they are stored in the HDF5 file). And to alleviate the transition, a `calibration_picks_polar` property was added that automatically converts between the polar/cartesian coordinates.

Here are some images illustrating the issue and solution.

# Here are some auto picked points:

original_overlay_picks

# On master, if we modify the tilt of a detector, the picks no longer line up:

original_overlay_picks_tilted

# With this fix, the picks now line up correctly:

fixed_overlay_picks_tilted

Fixes: #1310

### DIFF
--- a/hexrd/ui/calibration/auto/powder_runner.py
+++ b/hexrd/ui/calibration/auto/powder_runner.py
@@ -319,4 +319,4 @@ def save_picks_to_overlay(overlay, data_dict):
 
             picks[det_key].append(to_angles(hkl_picks, panel).tolist())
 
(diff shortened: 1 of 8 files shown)


saransh13 (approved):
LGTM.

PR #1316, opened 2022-12-30
Add extra sample points for masks

psavery:
For masks with few points (such as the rectangle mask, which only had
four corners for the points, and hand drawn masks with few points),
the masks would not be converted between the polar/raw views very well.

For example, for the rectangle mask, the four corners would be converted
between the views correctly. But the edges in the converted form would be
straight, when they would often need to be curved.

We need to add extra sample points for the masks so that conversions from
straight edges to curved edges are performed correctly. This commit does
that.

Here is an example that illustrates this:

# Rectangular mask created in polar view
masked_polar

# Current master branch conversion to raw view
masked_raw

You can see that the rectangle covers up most of the powder line in the polar view. But in the raw view,
it does not. The point coordinates were converted to the raw view correctly, but because there are only
four points, the edges are straight when they should be curved.

Adding extra sample points in between fixes this.

# This branch conversion to raw view
masked_raw_correct

### DIFF
--- a/hexrd/ui/create_polar_mask.py
+++ b/hexrd/ui/create_polar_mask.py
@@ -5,10 +5,15 @@ from skimage.draw import polygon
 from hexrd.ui.create_hedm_instrument import create_hedm_instrument
 from hexrd.ui.calibration.polarview import PolarView
 from hexrd.ui.hexrd_config import HexrdConfig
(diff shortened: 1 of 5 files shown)


bnmajor (approved):
LGTM, thanks for this fix!

Issue #1315, opened 2022-12-29
Auto powder calibration


saransh13:
Auto calibration is picking points on the polar image in the ellipse, rectangular, and hand drawn masked region defined by the user. The example is attached below.

![Screen Shot 2022-12-29 at 1 44 39 PM](https://user-images.githubusercontent.com/15834451/210014362-449df6f0-2ea7-4fb2-83e3-e7ff9db94034.png)

joelvbernier:
Those are only masked in the view for the lineout, fyi.Sent from my iPhoneOn Dec 29, 2022, at 13:51, saransh13 ***@***.***> wrote:﻿
Auto calibration is picking points on the polar image in the ellipse, rectangular, and hand drawn masked region defined by the user. The example is attached below.

psavery:
@saransh13 So, it looks like the auto powder calibrator is indeed receiving the masked images. However, the images are just plain numpy arrays, and the masked sections are just pixels with a value of 0. They are not treated as special otherwise. This means that regions of the masked section that are close to its borders are still getting points picked in it. See here for an example:

# No masks
![no_masks](https://user-images.githubusercontent.com/9558430/210094201-68264013-a5b4-45f0-9193-2e5097a59797.png)

# Hand drawn mask
![mask_no_panel_buffer](https://user-images.githubusercontent.com/9558430/210094228-3cf31c48-d0dd-4e51-bc8c-c171be677329.png)

You can see that some points were still picked in the masked region, but only close to its borders. 

However, if I go to the mask manager ("Edit" -> "Masks" -> "Manage Masks"), and I click "Masks to Panel Buffer", then the masked regions get applied to the panel buffer. The auto powder calibrator DOES pay attention to the panel buffer regions, and it won't pick points in them. See below, that points were not picked in the masked region anymore after they were applied to the panel buffer:

# Masks applied to panel buffer
![panel_buffer](https://user-images.githubusercontent.com/9558430/210094632-e6e7375f-a9a4-45bf-8aeb-0bb2ec13a29d.png)

So now we should discuss whether this behavior should be changed. Do we want to do something to the masked regions before the auto powder calibrator receives them, so that points are not picked in them? For example, maybe making the masked region `nan` instead of 0 would work. Or using a numpy masked array could work (although we may need to change something on the HEXRD side to make this work as well).

Or should we keep this behavior, where the calibrator only avoids picking points in the panel buffer regions?

PR #1317, opened 2023-01-02
Skip masked regions during auto picking

psavery:
Previously, masked regious could still have picks in it, since the masked regions were just treated as pixels with a value of zero.

To prevent masked regions from being auto picked, we are now applying the masked regions to the panel buffer during auto picking. The auto picker avoids panel buffer regions, so this fixes that issue.

Fixes: #1315


saransh13 (approved):
LGTM!

PR #1322, opened 2023-01-23
Maud

bnmajor:
~~WIP~~

First pass at exporting the required file format for Maud. The [new function](https://github.com/HEXRD/hexrdgui/compare/master...bnmajor:maud?expand=1#diff-da700bf8138e505bdb7fb2d673146b9567d8505912dad974e22c3f757d976754R156-R176) to write out the file has been adapted from of the original script by Joel:
```
import h5py
​
import numpy as np
​
header0 = """
_pd_block_id noTitle|#%d
​
_diffrn_detector Image Plate
_diffrn_detector_type Image Plate
_pd_meas_step_count_time ?
_diffrn_measurement_method ?
_diffrn_measurement_distance_unit mm
_pd_instr_dist_spec/detc %f
_diffrn_radiation_wavelength ?
_diffrn_source_target ?
_diffrn_source_power ?
_diffrn_source_current ?
_pd_meas_angle_omega 0.0
_pd_meas_angle_chi 0.0
_pd_meas_angle_phi 0.0
_riet_par_spec_displac_x 0
_riet_par_spec_displac_y 0
_riet_par_spec_displac_z 0
_riet_meas_datafile_calibrated false
"""
​
header = "_pd_block_id noTitle|#%d"
​
block_hdr = """
%s
​
_pd_meas_angle_eta %f
​
loop_
_pd_proc_2theta_corrected
_pd_meas_intensity_total
"""
​
with open('dummy.esg', 'w') as fid:
    for i, eta in enumerate(np.average(eta_vec.reshape(36, 50), axis=1).flatten()):
        if i == 0:
            hstr = header0 % (i, np.linalg.norm(panel.tvec))
        else:
            hstr = header % i
        print(block_hdr % (hstr, eta), file=fid)
        tth_eta = np.vstack([tth_vec, np.tile(eta, len(tth_vec))]).T
        det_xy = panel.angles_to_cart(np.radians(tth_eta))
        radii = np.sqrt(np.sum(det_xy*det_xy, axis=1))
        for rho, inten in zip(radii, summed[i]):
            print('{0:.6f}  {1:.6f}'.format(rho, inten), file=fid)
```
- [Headers](https://github.com/HEXRD/hexrdgui/compare/master...bnmajor:maud?expand=1#diff-77b66f4b11bb957c24e405592be9ef58e55a664ded06b89d94af94cd3776c2c4R1-R33) have been updated and tested and appear to be correct now
- `eta_vec` created ~~[here](https://github.com/HEXRD/hexrdgui/compare/master...bnmajor:maud?expand=1#diff-f9308985cbba0fcd56144d0011f8be6cdc78d2fde975680c1ea9f13b7da95c64R123-R125)~~ [here](https://github.com/HEXRD/hexrdgui/pull/1322/files#diff-da700bf8138e505bdb7fb2d673146b9567d8505912dad974e22c3f757d976754R160)
- `tvec` pulled from [instrument](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/calibration/polarview.py#L60-L61)
- `tth_eta` ~~built from `tth_vec` computed [here](https://github.com/HEXRD/hexrdgui/compare/master...bnmajor:maud?expand=1#diff-f9308985cbba0fcd56144d0011f8be6cdc78d2fde975680c1ea9f13b7da95c64R119-R121)~~ pulled from [azimuthal integration](https://github.com/HEXRD/hexrdgui/pull/1322/files#diff-da700bf8138e505bdb7fb2d673146b9567d8505912dad974e22c3f757d976754R168)
- `summed` is pulled from the [raw re-scaled image](https://github.com/HEXRD/hexrdgui/compare/master...bnmajor:maud?expand=1#diff-da700bf8138e505bdb7fb2d673146b9567d8505912dad974e22c3f757d976754R173)

The file is written out and can be loaded by Maud but is not loaded properly (no plot is generated). No errors or feedback given.

### DIFF
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -158,6 +158,7 @@ class InstrumentViewer:
 
         with open(filename, 'w') as fid:
             eta_vec = np.degrees(self.angular_grid[0])
+            intensities = self.raw_rescaled_img
             for i, eta in enumerate(np.average(eta_vec, axis=1).flatten()):
                 if i == 0:
                     hstr = header0 % (i, np.linalg.norm(self.pv.tvec_s), eta)
@@ -166,7 +167,6 @@ class InstrumentViewer:
                 fid.write(hstr)
                 fid.write(block_hdr)
                 tth = HexrdConfig().last_unscaled_azimuthal_integral_data[0]
-                intensities = self.raw_rescaled_img
                 for rho, inten in zip(tth, intensities[i]):
                     vals = f' {rho}  {inten}\n'
                     fid.write(vals)
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -156,17 +156,17 @@ class InstrumentViewer:
     def write_maud(self, filename='polar_to_maud.esg'):
         filename = Path(filename)
 
-        with open(filename, 'wb') as fid:
+        with open(filename, 'w') as fid:
             eta_vec = np.degrees(self.angular_grid[0])
             for i, eta in enumerate(np.average(eta_vec, axis=1).flatten()):
                 if i == 0:
                     hstr = header0 % (i, np.linalg.norm(self.pv.tvec_s), eta)
                 else:
                     hstr = header % (i, eta)
-                fid.write(hstr.encode('ascii'))
-                fid.write(block_hdr.encode('ascii'))
+                fid.write(hstr)
+                fid.write(block_hdr)
                 tth = HexrdConfig().last_unscaled_azimuthal_integral_data[0]
                 intensities = self.raw_rescaled_img
                 for rho, inten in zip(tth, intensities[i]):
                     vals = f' {rho}  {inten}\n'
-                    fid.write(vals.encode('ascii'))
+                    fid.write(vals)
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -9,6 +9,7 @@ from hexrd.ui.constants import ViewType
 from hexrd.ui.create_hedm_instrument import create_hedm_instrument
 from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.overlays import update_overlay_data
+from hexrd.ui.calibration.utils.maud_headers import header0, header, block_hdr
 
 
 def polar_viewer():
@@ -151,3 +152,21 @@ class InstrumentViewer:
             with h5py.File(filename, 'w') as f:
                 for key, value in data.items():
                     f.create_dataset(key, data=value)
+
+    def write_maud(self, filename='polar_to_maud.esg'):
+        filename = Path(filename)
+
+        with open(filename, 'wb') as fid:
+            eta_vec = np.degrees(self.angular_grid[0])
+            for i, eta in enumerate(np.average(eta_vec, axis=1).flatten()):
+                if i == 0:
+                    hstr = header0 % (i, np.linalg.norm(self.pv.tvec_s), eta)
+                else:
+                    hstr = header % (i, eta)
+                fid.write(hstr.encode('ascii'))
+                fid.write(block_hdr.encode('ascii'))
+                tth = HexrdConfig().last_unscaled_azimuthal_integral_data[0]
+                intensities = self.raw_rescaled_img
+                for rho, inten in zip(tth, intensities[i]):
+                    vals = f' {rho}  {inten}\n'
+                    fid.write(vals.encode('ascii'))
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -158,15 +158,25 @@ class InstrumentViewer:
 
         with open(filename, 'w') as fid:
             eta_vec = np.degrees(self.angular_grid[0])
-            intensities = self.raw_rescaled_img
+            intensities = self.img
+            first_block = True
             for i, eta in enumerate(np.average(eta_vec, axis=1).flatten()):
-                if i == 0:
+                if np.all(np.isnan(intensities[i])):
+                    # Skip this block
+                    continue
+
+                if first_block:
                     hstr = header0 % (i, np.linalg.norm(self.pv.tvec_s), eta)
+                    first_block = False
                 else:
                     hstr = header % (i, eta)
+
                 fid.write(hstr)
                 fid.write(block_hdr)
                 tth = HexrdConfig().last_unscaled_azimuthal_integral_data[0]
                 for rho, inten in zip(tth, intensities[i]):
+                    if np.isnan(inten):
+                        continue
+
                     vals = f' {rho}  {inten}\n'
                     fid.write(vals)
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -822,7 +822,7 @@ class MainWindow(QObject):
         self.ui.action_run_wppf.setEnabled(is_polar and has_images)
         self.ui.action_edit_apply_laue_mask_to_polar.setEnabled(is_polar)
         self.ui.action_edit_apply_powder_mask_to_polar.setEnabled(is_polar)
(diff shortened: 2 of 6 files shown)


bnmajor:
@saransh13 Maybe something jumps out as incorrect to you here? I was testing with the TARDIS config and images at the moment.

Also, at the moment there is simply a `To Maud` option under the export menu when in polar view. It would be good to know if this is sufficient, as well as if there is a better way to name/refer to the export option.

bnmajor:
@saransh13 Tested with TARDIS produces the following in Maud now:
maud

saransh13 (changes requested):
Other than the one comment, LGTM!

saransh13:
It looks like Mac is just using the name of the last file that the program interacted with. I'm ok with merging if this is not the behavior on other platforms.
Screen Shot 2023-01-23 at 2 51 08 PM

hexrd/ui/resources/ui/main_window.ui
```diff
@@ -496,6 +497,14 @@
...
    </property>
   </action>
+  <action name="action_export_to_maud">
+   <property name="enabled">
+    <bool>false</bool>
+   </property>
+   <property name="text">
+    <string>To Maud</string>
+   </property>
+  </action>
```
psavery, line 507: Can we add a tooltip like the following?

```
Export to Maud ESG format. Only available in the polar view.
```

hexrd/ui/main_window.py
```diff
@@ -512,6 +514,16 @@ def on_action_export_current_plot_triggered(self):
             HexrdConfig().working_dir = os.path.dirname(selected_file)
             return self.ui.image_tab_widget.export_current_plot(selected_file)
 
+    def on_action_export_to_maud_triggered(self):
+        filters = 'ESG files (*.esg)'
+
+        selected_file, selected_filter = QFileDialog.getSaveFileName(
+            self.ui, 'Export to Maud', HexrdConfig().working_dir, filters)
```
psavery, line 521:
```suggestion
        default_path = os.path.join(HexrdConfig().working_dir, "maud.esg")
        selected_file, selected_filter = QFileDialog.getSaveFileName(
            self.ui, 'Export to Maud', default_path, filters)
```

@saransh13 If you want a different default name rather than the last filename used (as mentioned in your other comment), I believe we can modify that here, like shown above.

hexrd/ui/calibration/polar_plot.py
```diff
@@ -151,3 +152,21 @@ def write_image(self, filename='polar_image.npz'):
...
+            eta_vec = np.degrees(self.angular_grid[0])
+            for i, eta in enumerate(np.average(eta_vec, axis=1).flatten()):
+                if i == 0:
+                    hstr = header0 % (i, np.linalg.norm(self.pv.tvec_s), eta)
+                else:
+                    hstr = header % (i, eta)
+                fid.write(hstr.encode('ascii'))
+                fid.write(block_hdr.encode('ascii'))
+                tth = HexrdConfig().last_unscaled_azimuthal_integral_data[0]
+                intensities = self.raw_rescaled_img
```
psavery, line 169: Since `self.raw_rescaled_img` is a computed property, can you move this line outside of the loop?

hexrd/ui/calibration/polar_plot.py
```diff
@@ -151,3 +152,21 @@ def write_image(self, filename='polar_image.npz'):
...
+    def write_maud(self, filename='polar_to_maud.esg'):
+        filename = Path(filename)
+
+        with open(filename, 'wb') as fid:
+            eta_vec = np.degrees(self.angular_grid[0])
+            for i, eta in enumerate(np.average(eta_vec, axis=1).flatten()):
+                if i == 0:
+                    hstr = header0 % (i, np.linalg.norm(self.pv.tvec_s), eta)
+                else:
+                    hstr = header % (i, eta)
```
psavery, line 165: This is a very nice design!

hexrd/ui/resources/ui/main_window.ui
bnmajor: Added in, thanks!

hexrd/ui/main_window.py
bnmajor: @saransh13 I've updated the code to use this approach now

hexrd/ui/calibration/polar_plot.py
bnmajor: Yes, thank you!

hexrd/ui/calibration/polar_plot.py
bnmajor: This was @joelvbernier's design :slightly_smiling_face: He had a great start on getting the export correct!

saransh13 (dismissed):

saransh13 (changes requested):
The background is not subtracted, and there are no masks (custom) applied to the file that is written out. One option is a pop up box which opens after user selects `Export to Maud`. This pop up box can have check boxes to turn on background subtraction (`snip_background`) and apply masking to the output file.

Screen Shot 2023-01-24 at 1 06 07 PM

saransh13:
I had another discussion today with an experienced MAUD user, and she suggested removing all the rows which have `nan` intensities (instead of replacing it with `0`). Instead of filling it with zeros, MAUD ignores it. Image of MAUD file in its current form, and a "fixed" version of the same file with zeros intensity rows removed and the associated images upon load below:

Screen Shot 2023-01-24 at 3 49 33 PM

Screen Shot 2023-01-24 at 3 54 00 PM

hexrd/ui/calibration/polar_plot.py
```diff
@@ -151,3 +152,21 @@ def write_image(self, filename='polar_image.npz'):
             with h5py.File(filename, 'w') as f:
                 for key, value in data.items():
                     f.create_dataset(key, data=value)
+
+    def write_maud(self, filename='polar_to_maud.esg'):
+        filename = Path(filename)
+
+        with open(filename, 'w') as fid:
+            eta_vec = np.degrees(self.angular_grid[0])
+            intensities = self.raw_rescaled_img
```
psavery, line 161: The reason SNIP and masks are not exported (as mentioned [here](https://github.com/HEXRD/hexrdgui/pull/1322#pullrequestreview-1268301956)) is because the `self.raw_rescaled_img` is being used here. This can be changed to `self.img` instead to include the background subtraction and masking.

I personally think it would be better to not have extra checkboxes pop up for whether to include the background subtraction/masking, and just export the polar image as-is. This is because the user can already turn off SNIP background subtraction in the polar view settings and turn off masking in the mask manager before exporting to Maud if they don't want to include them, and if they go this route, they have visual confirmation in the polar view of what they are actually exporting before it is done.

Still, if it is a common operation of users to visually see the SNIP background subtraction and masking in the polar view, but want to exclude them when exporting to Maud, we should add a pop-up with checkboxes.

bnmajor:
@psavery Thanks so much for the commit!

@saransh13 I think Patrick makes a good point and it makes sense that the user exports what they see (which is now the case), but if it would be preferable to have the dialog still let me know and I will add that in! Also now that we are using `self.img` instead of `self.raw_rescaled_image` (which did not have NaNs) we simply skip the rows with NaNs as intensity values, having the same effect that you explained above. A quick example of before:
tardis_maud_old
And after the updates (with background subtraction and masking):
tardis_maud_new

psavery:
@saransh13 Right now, it looks like the Maud image has an inverted y-axis compared to the polar view. Is this expected?

Here's the polar view:

image

Here's the Maud image that came out of it:

image


You can see that the rectangle I masked in the polar view is on the top detector in hexrdgui, but on the bottom detector in Maud. I think this might just be that the y-axis is inverted.

hexrd/ui/calibration/polar_plot.py
saransh13: Hmm yes, I think this is a better idea.

saransh13:
The azimuthal angle increases top to bottom in hexrd, and bottom to top in MAUD. I think the image is rendered correctly.

psavery (approved):
@saransh13 verified the image looks correct, so LGTM!

saransh13 (approved):

PR #1323, opened 2023-01-27
Fix convert_panel_buffer_to_2d_array()

psavery:
The 2-value panel buffer represents the `x, y` directions in mm,
not in pixels.

So we need to convert these values to pixels first, and then we also needed
to fix an issue where the logic wouldn't work if the borders were 0. This
would result in the whole array being `False`. That is now fixed.

### DIFF
--- a/hexrd/ui/utils/__init__.py
+++ b/hexrd/ui/utils/__init__.py
@@ -482,12 +482,20 @@ def convert_panel_buffer_to_2d_array(panel):
         # Just make a panel buffer with all True values
         panel.panel_buffer = np.ones(panel.shape, dtype=np.bool)
     elif panel.panel_buffer.shape == (2,):
-        # The two integers are specifying the borders in x and y
-        borders = panel.panel_buffer.astype(int)
+        # The two floats are specifying the borders in mm for x and y.
+        # Convert to pixel borders. Swap x and y so we have i, j in pixels.
+        borders = np.round([
+            panel.panel_buffer[1] / panel.pixel_size_row,
+            panel.panel_buffer[0] / panel.pixel_size_col,
+        ]).astype(int)
 
         # Convert to array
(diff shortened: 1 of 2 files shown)


hexrd/ui/utils/__init__.py
```diff
@@ -482,12 +482,20 @@ def convert_panel_buffer_to_2d_array(panel):
...
         panel.panel_buffer = np.ones(panel.shape, dtype=np.bool)
     elif panel.panel_buffer.shape == (2,):
-        # The two integers are specifying the borders in x and y
-        borders = panel.panel_buffer.astype(int)
+        # The two floats are specifying the borders in mm for x and y.
+        # Convert to pixel borders. Swap x and y so we have i, j in pixels.
+        borders = np.round([
+            panel.panel_buffer[1] / panel.pixel_size_row,
+            panel.panel_buffer[0] / panel.pixel_size_col,
+        ]).astype(int)
```
saransh13, line 490: I need help understanding the logic of switching the order, but I trust your judgment.

saransh13 (approved):
Works for me too. We can merge these changes!

hexrd/ui/utils/__init__.py
psavery: So, for instance, in Cartesian coordinates, the `x` is from left to right. But if we convert that to pixels, the columns (`j`) is what is from left to right instead.

So when we go from `xy` => `ij`, we need to swap `x` and `y`. But I'm happy to verify this too.

psavery: Yes I just verified that this works as expected. For an example, I set the `x` panel buffer to roughly about 44% of the width of the panel. You can see that after using our conversion function to a 2D panel buffer, a large chunk from both the left and right side is `False`:

example

PR #1332, opened 2023-02-04
Only rerender stereo border when option changes

psavery:
If the user changes the option to show the stereo border, don't rerender the whole image. Only rerender the stereo border artist.


saransh13 (approved):
LGTM!

Issue #1331, opened 2023-02-02
error messages in Messages window


donald-e-boyce:
This error comes up freuqently.

```
AttributeError: 'NoneType' object has no attribute 'degrees'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/ui/indexing/run.py", line 150, in ome_maps_selected
    config = create_indexing_config()
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/ui/indexing/create_config.py", line 52, in create_indexing_config
    config.set('material:tth_width', np.degrees(material.planeData.tThWidth))
TypeError: loop of ufunc does not support argument 0 of type NoneType which has no callable degrees method
Traceback (most recent call last):
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/ui/image_canvas.py", line 481, in update_overlays
    self.draw_overlay(overlay)
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/ui/image_canvas.py", line 284, in draw_overlay
    for axis, data in self.overlay_axes_data(overlay):
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/ui/image_canvas.py", line 230, in overlay_axes_data
    if not overlay.data:
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/ui/overlays/overlay.py", line 281, in data
    self._data |= self.generate_overlay()
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/ui/overlays/rotation_series_overlay.py", line 193, in generate_overlay
    sim_data = instr.simulate_rotation_series(
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/instrument.py", line 1622, in simulate_rotation_series
    results[det_key] = panel.simulate_rotation_series(
  File "/Users/deboyce/miniconda3/envs/hexrdgui/lib/python3.10/site-packages/hexrd/instrument.py", line 3380, in simulate_rotation_series
    full_hkls = xrdutil._fetch_hkls_from_planedata(plane_data)
[...]

PR #1340, opened 2023-02-23
Make a deep copy of materials in indexing config

psavery:
HEXRD sometimes modifies these materials. This is why we were originally
setting tth width on the config, so that our material would not be modified.
    
However, HEXRD also sometimes modifies other properties. So make a deep copy
of the materials so that they do not get unintentionally modified in HEXRDGUI.

This also fixes the error that would occur when `np.degrees()` was being called on the `tth_width` when it was `None`.

Fixes: #1331

### DIFF
--- a/hexrd/ui/indexing/create_config.py
+++ b/hexrd/ui/indexing/create_config.py
@@ -1,8 +1,6 @@
 import copy
 import os
 
-import numpy as np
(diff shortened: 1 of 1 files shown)


bnmajor (approved):
Nice!

Issue #1335, opened 2023-02-10
Edit > Transform detector produces error


saransh13:
`Traceback (most recent call last):
  File "/Users/soderlind3/hexrdgui/hexrd/ui/transform_dialog.py", line 76, in apply_transforms
    ilm.begin_processing(postprocess=True)
  File "/Users/soderlind3/hexrdgui/hexrd/ui/image_load_manager.py", line 143, in begin_processing
    thread_pool = QThreadPool(self.parent)
TypeError: 'PySide2.QtCore.QThreadPool.__init__' called with wrong argument types:
  PySide2.QtCore.QThreadPool.__init__(builtin_function_or_method)
Supported signatures:
  PySide2.QtCore.QThreadPool.__init__(typing.Optional[PySide2.QtCore.QObject] = None)`

psavery:
@bnmajor Since the `ImageLoadManager` inherits from `QObject`, it has a `parent()` function which returns the `QObject`'s parent.

However, I see we are setting `self.parent = ...` in `read_data()` as well. The issue above must have occurred before that was set.

In any case, to avoid the name conflict, perhaps we should name `self.parent` to something else? Then in the `QThreadPool` function, we can probably pass in `self` as the parent instead of `self.parent` I think?

bnmajor:
@saransh13 Are you able to provide a little more context as to when you saw this bug? Just loading a config and images and trying to apply a transform (to single images or all) I am not seeing this.

bnmajor:
@psavery I've opened a PR that at least addresses the issues that you raised above, thank you!

@saransh13 Please feel free to try #1341 when you have a chance as this may be the issue (as Patrick pointed out), but I was unable to reproduce the error so I am not positive that this is the solution

psavery:
Hi @bnmajor. It appears that this issue occurs if you load in a state file and then try to run `Edit -> Transform Detector`. I believe this is because the `ImageLoadManager().read_data()` function does not run when you load in a state file, and therefore the `ui_parent` is not set. I'll send you an example state file so you can try it out.

After merging #1341, the error displayed is now as follows:
```python
Traceback (most recent call last):
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/transform_dialog.py", line 76, in apply_transforms
    ilm.begin_processing(postprocess=True)
[...]

PR #1341, opened 2023-02-23
Do not override the QObject parent

bnmajor:
Instead, save the ui parent as its own attribute, ui_parent.

Addresses part of #1335

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -122,13 +122,13 @@ class ImageLoadManager(QObject, metaclass=QSingleton):
                     files[idx].append('/'.join([path, f]))
         return files
 
-    def read_data(self, files, data=None, parent=None):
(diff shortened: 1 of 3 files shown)


psavery (approved):
LGTM

Now, if that error crops up again, it should say something like "no attribute named `ui_parent`" or something like that.

bnmajor:
@psavery Thanks! I mentioned it in the original issue as well, but I wasn't able to reproduce this error to definitively confirm this fixes it. Should we just get this in anyway (like you said, it needed to be corrected either way) and wait to close the issue until Saransh confirms it's good?

psavery:
@bnmajor That sounds good to me!

psavery:
This PR is good because it fixes the variable shadowing issue. I will merge it.

However, I just found out how to reproduce the issue in #1335, and this does not unfortunately fix it. I will post more about it there.

PR #1343, opened 2023-02-23
Parse the arguments within the main() method

psavery:
This makes it so that the argument parsing does not happen except in the `hexrdgui` entrypoint.

This is helpful for things like `pytest`, where we don't want HEXRDGUI to parse the `pytest` arguments.

It could also allow us to develop other scripts while skipping the arg parsing.


bnmajor (approved):
LGTM

PR #1349, opened 2023-03-01
Make sure the transform dialog uses the read_data path

### LINKED ISSUE #1335: Edit > Transform detector produces error

bnmajor:
This makes sure that no matter how the data is loaded (manually or via state file) all required attributes are correctly set in the ImageLoadManager.

Fixes #1335

### DIFF
--- a/hexrd/ui/image_load_manager.py
+++ b/hexrd/ui/image_load_manager.py
@@ -122,18 +122,18 @@ class ImageLoadManager(QObject, metaclass=QSingleton):
                     files[idx].append('/'.join([path, f]))
         return files
 
-    def read_data(self, files, data=None, ui_parent=None):
+    def read_data(self, files=None, data=None, ui_parent=None, **kwargs):
         # When this is pressed read in a complete set of data for all detectors.
         # Run the imageseries processing in a background thread and display a
         # loading dialog
-        self.parent_dir = HexrdConfig().images_dir
-        self.set_state()
-        self.ui_parent = ui_parent
-        self.files = files
+        if files:
+            self.parent_dir = HexrdConfig().images_dir
+            self.files = files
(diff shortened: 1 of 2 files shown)


hexrd/ui/image_load_manager.py
```diff
@@ -122,18 +122,18 @@ def match_selected_files(self, fnames, search):
...
+            self.parent_dir = HexrdConfig().images_dir
+            self.files = files
+            self.empty_frames = data['empty_frames'] if data else 0
         self.data = {} if data is None else data
-        self.empty_frames = data['empty_frames'] if data else 0
-
-        self.begin_processing()
+        self.ui_parent = ui_parent
+        self.set_state(kwargs.get('state', None))
+        self.begin_processing(kwargs.get('postprocess', False))
```
psavery, line 136: Rather than having kwargs (where we might pass in an an argument with a wrong keyword without any errors), can we have `state=None` and `postprocess=False` in the arguments?

psavery (approved):
It works for me when I test it. LGTM!

PR #1350, opened 2023-03-06
Add ability to load state file during startup

psavery:
It is an optional command line argument.


bnmajor (approved):
This is awesome, very useful option!

PR #1354, opened 2023-03-09
Add Rygg pinhole distortion

psavery:
This modifies the GUI pinhole distortion methods to use the Rygg one introduced in hexrd/hexrd#491.

Need to fix:
- [x] If there are multiple overlays, we apply the Rygg pinhole tth distortion to one of them, and then we apply this distortion to the whole polar view, the transformation of the other overlays appears to have some issue. Specifically, most of them disappear. We don't have this issue with the sample layer tth distortion.
- [x] On polar view x-axis, only show `sam` for sample correction. Show `pin` for pinhole correction.


psavery:
@saransh13 This one is ready for review now too!

saransh13 (approved):
Tested all the features for TARDIS, and everything works as expected. LGTM!

Issue #1353, opened 2023-03-08
error when fitting the Pilatus 


donald-e-boyce:
After loading the Pilatus composite example in the hexrd-examples repo, I tried to run a simple fit for the three translations of the 0_0 detector. The config was loaded from the hexrd file, and the images were loaded.  I got the following error during the point-picking dialog:  

```
Traceback (most recent call last):
  File "/Users/deboyce/Source/hexrdgui/branches/hexrdgui-dev/hexrd/ui/async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/Users/deboyce/Source/hexrdgui/branches/hexrdgui-dev/hexrd/ui/calibration/auto/powder_runner.py", line 112, in run_extract_powder_lines
    self.save_picks_to_overlay()
  File "/Users/deboyce/Source/hexrdgui/branches/hexrdgui-dev/hexrd/ui/calibration/auto/powder_runner.py", line 273, in save_picks_to_overlay
    save_picks_to_overlay(self.active_overlay, self.data_dict)
  File "/Users/deboyce/Source/hexrdgui/branches/hexrdgui-dev/hexrd/ui/calibration/auto/powder_runner.py", line 326, in save_picks_to_overlay
    overlay.calibration_picks_polar = picks
  File "/Users/deboyce/Source/hexrdgui/branches/hexrdgui-dev/hexrd/ui/overlays/powder_overlay.py", line 183, in calibration_picks_polar
    panel = instr.detectors[det_key]
KeyError: 'Pilatus2M_0_0'
```
Actually, the same thing happened in the monolithic case, but I couldn't reproduce it.  

Here is a pic of the GUI state:
<img width="1606" alt="image" src="https://user-images.githubusercontent.com/1141241/223780980-1b24ecdb-ae8b-4ddd-9f66-9a32335698fb.png">

psavery:
I see! I was able to reproduce this issue. I will look into it.

psavery:
So I actually found out this only occurs if you perform "Fast Powder" auto-picking while in the *Cartesian* view mode. It is now fixed in #1355!

PR #1355, opened 2023-03-10
Skip converting fast powder picks to polar

psavery:
We are now storing the calibration picks as Cartesian coordinates -
not polar coordinates. As such, we do not need to convert to polar
coordinates anymore, but can set the Cartesian picks directly.

This fixes an issue where, if you were looking at the Cartesian view,
the auto picks would fail to convert to polar coordinates (due to the
fact that the Cartesian view has a fake instrument it uses).

This PR also fixes an issue where an error would occur if no Laue spots
could be generated due to a lack of HKLs.

Fixes: #1353


bnmajor (approved):
LGTM!

PR #1356, opened 2023-03-11
Add example "Help" button to simple image series dialog

psavery:
This adds a convenience function for linking `helpRequested` in a `QDialogButtonBox` to opening a URL on a web browser, and it adds an example with the simple image series dialog.

This also removes the instructions for installation from the README and replaces it with a link to the new install instructions in the hexrdgui documentation.


bnmajor:
@psavery This is awesome! Just one question - the example icon on the docs page is the question mark but in my testing I am seeing (what I believe) is a life preserver? Just curious about the differences there. Either way this is gonna be great to have!

help

bnmajor (approved):

psavery:
I did a little research this morning on the icons. What I found out is:

1. The icons in that image above are the native Qt icons. I believe they will be what we see on most of the operating systems (since those icons come packaged with Qt).
2. Qt on Linux can sometimes use themes from the desktop settings. Mine is apparently using a `Yaru` theme, but there are other themes as well (such as `gnome`).
3. We do have the option to package our own themes too (although I don't know all of the details about how to do this). If we like `Yaru` enough, maybe we can figure out how to package it with Qt.
I wish the Qt docs were clearer on how to check and change icon themes.

Issue #1359, opened 2023-03-20
IMAGE PLATE 3 for TARDIS import


saransh13:
During the loading of the image plate 3 for the TARDIS import tool, the boundaries of the image are correctly defined, but part of the image plate does not have usable intensity (the part where the intensity suddenly drops at the edges). This part of the image plate is bleeding out during the stereographic projection as well as the polar view. 

The boundaries should remain unchanged so that the user is able to line up the boundaries with the image, but when the user hit `complete and reload detectors` then, only the middle part of the image should be selected. The image below shows the part of the image plate 3 that needs to be cropped off. Equal amounts should be cropped from left and right (equal to the cropping on the right side since it's larger)

![Picture1](https://user-images.githubusercontent.com/15834451/226449673-7f155d09-68e4-4181-a7b9-79c0927a20cb.png)

saransh13:
@bnmajor In the IGOR code, the image is cropped to 238x1330 pixels centered at the hole in the middle (see image below) We should stick to the same dimensions

![Screen Shot 2023-03-30 at 3 37 06 PM](https://user-images.githubusercontent.com/15834451/228978948-fe1bb605-6c09-4cdf-ba84-39cbc96d391d.png)

PR #1360, opened 2023-03-24
Cylindrical detector updates

bnmajor:
This is a first pass at the cropping required for TARDIS image plate 3, as described in #1359. Relies on HEXRD/hexrd#507

### DIFF
--- a/hexrd/ui/interactive_template.py
+++ b/hexrd/ui/interactive_template.py
@@ -334,9 +334,7 @@ class InteractiveTemplate:
         if event.key == 'left' or event.key == 'up':
             angle *= -1.
         elif event.key != 'right' and event.key != 'down':
-            angle *= 1.            
-        # DEBUG
-        # print(f'In "on_key_rotate": {angle}')
+            angle *= 1.
         self.total_rotation += angle
         self.rotate_template(self.shape.xy, angle)
         self.redraw()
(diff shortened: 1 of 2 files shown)


bnmajor:
@saransh13 It would be great to get your feedback on these PRs. A couple of things:
1. I am not sure what steps I am not taking that you are, but even without any changes my output does not look like yours. This makes it hard for me to know how close to correct or incorrect my PRs may be. (Note: color map changed to make it more obvious how different my results look).
tardis_polar
tardis_stereo
2. I am arbitrarily cropping this image plate based on eyeballing your suggested cropping regions. Consequently, this may not be accurate. Some questions I have:
    1. Is this some arbitrary amount that needs cropped? Will it vary? Or is this something precise we can account for?
    2. If this is something more precise it may be simple to automatically crop, otherwise we should consider a special use case for TARDIS detector 3 where the user is provided a second tool to sub-select the piece that they want to keep.

The output produced with my changes:
tardis_polar_new
tardis_stereo_new

saransh13:
The only feedback is the detector bounding lines don't become dashed lines once I hit `ok` in the `import tool`
Screen Shot 2023-03-27 at 9 35 15 AM


The `polar` and `stereographic` view look great with the cropped images!
Screen Shot 2023-03-27 at 9 36 15 AM
Screen Shot 2023-03-27 at 9 36 40 AM

psavery:
These commits are in #1395. Closing.

PR #1361, opened 2023-03-25
Simulated overlay

### LINKED ISSUE #1319: Simulated transparent pattern in overlay
Add a feature to add a transparent simulated powder diffraction pattern in the azimuthally averaged lineout. Something that looks like the image below.

The powder simulator is already implemented in the `Material` class. It needs to be integrated and an option added in the overlay button. This overlay will also need to have a peak width parameter which changes how broad the shaded peaks are.

Putting it here so we can keep track of this, but this issue is only a medium priority.

Screen Shot 2023-01-09 at 4 21 49 PM

bnmajor:
~WIP~

Still needs:
- [x] Naming testing
- [x] Do not disable editor inputs (on first entry)
- [x] Make state persistent between sessions
- [x] Save selections to state file
- [x] Make sure plot is re-scaled to fit all data as overlays are added/changed
- [x] Suggest default `fwhm`/`scale` values?
- [x] Re-apply overlays when toggling away from/back to polar view
- [x] Ability to save azimuthal plot
- [x] Refactor/clean-up the changes made in the plot update
- [x] Add legend

~Most of these are pretty simple fixes and should be in soon. Just making this branch available in the meantime as a preview.~

azimuthal_plot

This branch should be ready now. Fixes #1319

### DIFF
--- a/hexrd/ui/azimuthal_overlay_manager.py
+++ b/hexrd/ui/azimuthal_overlay_manager.py
@@ -7,6 +7,7 @@ from hexrd.ui import utils
 
 from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.azimuthal_overlay_editor import AzimuthalOverlayEditor
+from hexrd.ui.azimuthal_overlay_style_picker import AzimuthalOverlayStylePicker
 from hexrd.ui.ui_loader import UiLoader
 from hexrd.ui.utils import block_signals
 
@@ -244,4 +245,5 @@ class AzimuthalOverlayManager:
         HexrdConfig().azimuthal_overlay_modified.emit()
 
     def edit_style(self):
-        pass
+        self._style_picker = AzimuthalOverlayStylePicker(self.active_overlay, self.ui)
+        self._style_picker.exec_()
--- a/hexrd/ui/azimuthal_overlay_manager.py
+++ b/hexrd/ui/azimuthal_overlay_manager.py
@@ -1,0 +1,231 @@
+from PySide2.QtCore import Qt, QItemSelectionModel
+from PySide2.QtWidgets import (
+    QCheckBox, QComboBox, QHBoxLayout, QHeaderView, QSizePolicy,
+    QTableWidgetItem, QWidget
+)
+from hexrd.ui import utils
+
+from hexrd.ui.hexrd_config import HexrdConfig
+from hexrd.ui.ui_loader import UiLoader
+from hexrd.ui.utils import block_signals
+
+
+COLUMNS = {
+    'name': 0,
+    'material': 1,
+    'visible': 2,
+}
+
+
+class AzimuthalOverlayManager:
+
+    def __init__(self, parent=None):
+        loader = UiLoader()
+        self.ui = loader.load_file('azimuthal_overlay_manager.ui', parent)
+
+        flags = self.ui.windowFlags()
+        self.ui.setWindowFlags(flags | Qt.Tool)
+
+        self.material_combos = []
+        self.visibility_boxes = []
+
+        self.setup_connections()
+
+    def setup_connections(self):
+        self.ui.table.selectionModel().selectionChanged.connect(
+            self.selection_changed)
+        self.ui.table.itemChanged.connect(self.table_item_changed)
+        self.ui.add_button.pressed.connect(self.add)
+        self.ui.remove_button.pressed.connect(self.remove)
+        self.ui.edit_style_button.pressed.connect(self.edit_style)
+
+    def show(self):
+        self.update_table()
+        self.ui.show()
+
+    def create_materials_combo(self, v):
+        materials = list(HexrdConfig().materials.keys())
+
+        if v not in materials:
+            raise Exception(f'Unknown material: {v}')
+
+        cb = QComboBox(self.ui.table)
+        size_policy = QSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
+        cb.setSizePolicy(size_policy)
+        for mat in materials:
+            cb.addItem(mat, mat)
+
+        cb.setCurrentIndex(materials.index(v))
+        cb.currentIndexChanged.connect(self.update_config_materials)
+        self.material_combos.append(cb)
+        return self.create_table_widget(cb)
+
+    def create_visibility_checkbox(self, v):
+        cb = QCheckBox(self.ui.table)
+        cb.setChecked(v)
+        cb.toggled.connect(self.update_visibilities)
+        self.visibility_boxes.append(cb)
+        return self.create_table_widget(cb)
+
+    def create_table_widget(self, w):
+        # These are required to center the widget...
+        tw = QWidget(self.ui.table)
+        layout = QHBoxLayout(tw)
+        layout.addWidget(w)
+        layout.setAlignment(Qt.AlignCenter)
+        layout.setContentsMargins(0, 0, 0, 0)
+        return tw
+
+    def clear_table(self):
+        self.material_combos.clear()
+        self.visibility_boxes.clear()
+        self.ui.table.clearContents()
+
+    def update_table(self):
+        block_list = [
+            self.ui.table,
+            self.ui.table.selectionModel()
+        ]
+
+        with block_signals(*block_list):
+            prev_selected = self.selected_row
+
+            self.clear_table()
+            self.ui.table.setRowCount(len(self.overlays))
+            for i, overlay in enumerate(self.overlays):
+                w = QTableWidgetItem(overlay['name'])
+                self.ui.table.setItem(i, COLUMNS['name'], w)
+
+                w = self.create_materials_combo(overlay['material'])
+                self.ui.table.setCellWidget(i, COLUMNS['material'], w)
+
+                w = self.create_visibility_checkbox(overlay['visible'])
+                self.ui.table.setCellWidget(i, COLUMNS['visible'], w)
+
+            if prev_selected is not None:
+                select_row = (prev_selected if prev_selected < len(self.overlays)
+                              else len(self.overlays) - 1)
+                self.select_row(select_row)
+
+            self.ui.table.resizeColumnsToContents()
+
+            # The last section isn't always stretching automatically, even
+            # though we have setStretchLastSection(True) set.
+            # Force it to stretch manually.
+            last_column = max(v for v in COLUMNS.values())
+            self.ui.table.horizontalHeader().setSectionResizeMode(
+                last_column, QHeaderView.Stretch)
+
+            # Just in case the selection actually changed...
+            self.selection_changed()
+
+    def select_row(self, i):
+        if i is None or i >= self.ui.table.rowCount():
+            # Out of range. Don't do anything.
+            return
+
+        # Select the row
+        selection_model = self.ui.table.selectionModel()
+        selection_model.clearSelection()
+
+        model_index = selection_model.model().index(i, 0)
+        command = QItemSelectionModel.Select | QItemSelectionModel.Rows
+        selection_model.select(model_index, command)
+
+    @property
+    def overlays(self):
+        return HexrdConfig().azimuthal_overlays
+
+    @property
+    def selected_row(self):
+        selected = self.ui.table.selectionModel().selectedRows()
+        return selected[0].row() if selected else None
+
+    def selection_changed(self):
+        self.update_enable_states()
+
+    def update_enable_states(self):
+        row_selected = self.selected_row is not None
+        self.ui.remove_button.setEnabled(row_selected)
+        self.ui.edit_style_button.setEnabled(row_selected)
+
+    def update_config_materials(self):
+        any_changed = False
+        for i in range(self.ui.table.rowCount()):
+            w = self.material_combos[i]
+            overlay = self.overlays[i]
+            if overlay['material'] != w.currentData():
+                overlay['material'] = w.currentData()
+                any_changed = True
+
+        if any_changed:
+            # In case the material was renamed
+            self.update_table()
+
+    def update_visibilities(self):
+        for i in range(self.ui.table.rowCount()):
+            w = self.visibility_boxes[i]
+            self.overlays[i]['visible'] = w.isChecked()
+
+    @property
+    def active_material_name(self):
+        return HexrdConfig().active_material_name
+
+    @property
+    def active_overlay(self):
+        i = self.selected_row
+        if i is None or i >= len(self.overlays):
+            return None
+
+        return self.overlays[i]
+
+    def table_item_changed(self, item):
+        col = item.column()
+        if col == COLUMNS['name']:
+            return self.overlay_name_edited(item)
+        else:
+            raise Exception(f'Item editing not implemented for column: {col}')
+
+    def overlay_name_edited(self, item):
+        row = item.row()
+        new_name = item.text()
+        modified_overlay = self.overlays[row]
+        old_name = modified_overlay['name']
+
+        # If the name matches any other overlay names, revert back so that
+        # we can keep unique names.
+        for overlay in self.overlays:
+            if new_name == overlay['name']:
+                # This name already exists. Revert changes.
+                item.setText(old_name)
+                return
+
+        modified_overlay['name'] = new_name
+
+    def create_unique_name(self):
+        existing_names = [o['name'] for o in self.overlays]
+        return utils.unique_name(existing_names, self.active_material_name)
+
+    def add_azimuthal_overlay(self):
+        data = {
+            'name': self.create_unique_name(),
+            'material': self.active_material_name,
+            'visible': True,
+            'fwhm': 0.25,
+            'scale': 1.0,
+            'color': '#ff0000',
+            'opacity': 0.3,
+        }
+        self.overlays.append(data)
+
+    def add(self):
+        self.add_azimuthal_overlay()
+        self.update_table()
+        self.select_row(len(self.overlays) - 1)
+
+    def remove(self):
+        self.overlays.pop(self.selected_row)
+        self.update_table()
+
+    def edit_style(self):
+        pass
--- a/hexrd/ui/azimuthal_overlay_manager.py
+++ b/hexrd/ui/azimuthal_overlay_manager.py
@@ -6,6 +6,7 @@ from PySide2.QtWidgets import (
 from hexrd.ui import utils
 
 from hexrd.ui.hexrd_config import HexrdConfig
+from hexrd.ui.azimuthal_overlay_editor import AzimuthalOverlayEditor
 from hexrd.ui.ui_loader import UiLoader
 from hexrd.ui.utils import block_signals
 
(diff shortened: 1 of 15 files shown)


bnmajor:
Current demo:

https://user-images.githubusercontent.com/51238406/227745646-a6b15109-b888-4585-b1fc-7b6b4d337249.mp4

bnmajor:
@psavery @saransh13 Still a WIP pending some clean-up but if there are any questions/concerns in the meantime I'm happy to take note now and make changes as I wrap this up

saransh13:
> WIP
> 
> Still needs:
> 
> * [x]  Naming testing
> * [x]  Do not disable editor inputs (on first entry)
> * [x]  Make state persistent between sessions
> * [x]  Save selections to state file
> * [x]  Make sure plot is re-scaled to fit all data as overlays are added/changed
> * [ ]  Suggest default `fwhm`/`scale` values?
You can use default `fwhm = 2`, and `scale = azimuthal_lineout.max()/100`

> * [x]  Re-apply overlays when toggling away from/back to polar view
> * [ ]  Ability to save azimuthal plot
> * [x]  Refactor/clean-up the changes made in the plot update
> * [x]  Add legend
> 
> Most of these are pretty simple fixes and should be in soon. Just making this branch available in the meantime as a preview.
> 
> azimuthal_overlays_example

bnmajor:
Two questions before this is complete:
1. What are we saving out? Plot data (as it seems is the behavior for "save current view") or an image of the azimuthal plot?
2. I can't remember what was decided - are we providing a good estimate for scale based on the integral sum? Or on the simulated data?

saransh13:
> 1. What are we saving out? Plot data (as it seems is the behavior for "save current view") or an image of the azimuthal plot?

Yeah, we can leave the `save current view` as is and add an option to export an image with the overlays!

saransh13:
By the way, I'm not seeing an option to toggle the legends on (as shown in your image in the top right)

Screen Shot 2023-03-26 at 1 33 25 PM

bnmajor:
> By the way, I'm not seeing an option to toggle the legends on (as shown in your image in the top right)
> 
> Screen Shot 2023-03-26 at 1 33 25 PM

I just force pushed some changes so you may need to do a `fetch` and `pull` again to get the latest. Should be on SHA `
d3b5882` now.

hexrd/ui/image_canvas.py
```diff
@@ -998,6 +1002,43 @@ def compute_azimuthal_integral_sum(self, scaled=True):
...
+                    result,
+                    color=overlay['color'],
+                    alpha=overlay['opacity']
+                )
+                artist.set_label(overlay['name'])
+                self.azimuthal_overlay_artists.append({
+                    'name': overlay['name'],
+                    'material': overlay['material'],
+                    'artist': artist,
+                    'line': line,
```
psavery, line 1030: See comment above about potentially restructuring this.

hexrd/ui/image_canvas.py
```diff
@@ -109,6 +111,7 @@ def clear_azimuthal_integral_axis(self):
         self.clear_wppf_plot()
         self.azimuthal_integral_axis = None
         self.azimuthal_line_artist = None
+        self.azimuthal_overlay_artists.clear()
```
psavery, line 114: See other comment about creating a `clear_azimuthal_overlay_artists()` function. We would then call that here.

hexrd/ui/image_mode_widget.py
```diff
@@ -408,6 +415,17 @@ def on_eta_max_changed(self, max_value):
...
+        if hasattr(self, '_overlay_manager'):
+            self._overlay_manager.ui.reject()
+            del self._overlay_manager
+
+        self._overlay_manager = AzimuthalOverlayManager(self.ui)
+        self._overlay_manager.show()
+
+    def toggle_legend(self, value):
+        HexrdConfig().azimuthal_legend = value
+        HexrdConfig().azimuthal_overlay_modified.emit()
```
psavery, line 428: Can we move the legend and the button to save the plot into the overlay manager? We would need to create a section in that dialog for global options (as opposed to the overlay-specific options displayed at the bottom).

hexrd/ui/image_canvas.py
```diff
@@ -998,6 +1002,43 @@ def compute_azimuthal_integral_sum(self, scaled=True):
...
+                    'material': overlay['material'],
+                    'artist': artist,
+                    'line': line,
+                })
+        if HexrdConfig().azimuthal_legend and len(self.azimuthal_overlay_artists):
+            self.azimuthal_integral_axis.legend()
+        elif self.azimuthal_integral_axis:
+            try:
+                self.azimuthal_integral_axis.get_legend().remove()
+            except:
```
psavery, line 1037: Let's not use bare exceptions. Can we either catch specific exceptions or check things with `if` statements?

hexrd/ui/azimuthal_overlay_manager.py
```diff
@@ -0,0 +1,251 @@
...
+            # Out of range. Don't do anything.
+            return
+
+        # Select the row
+        selection_model = self.ui.table.selectionModel()
+        selection_model.clearSelection()
+
+        model_index = selection_model.model().index(i, 0)
+        command = QItemSelectionModel.Select | QItemSelectionModel.Rows
+        selection_model.select(model_index, command)
```
psavery, line 138: It might be good for us to make a common base class between this class and the overlay manager class that contains functions like these where the logic is identical. I might make a quick attempt at this.

hexrd/ui/image_mode_widget.py
```diff
@@ -408,6 +415,17 @@ def on_eta_max_changed(self, max_value):
                 HexrdConfig().set_polar_res_eta_min(min_value, rerender=False)
         HexrdConfig().polar_res_eta_max = max_value
 
+    def show_overlay_manager(self):
+        if hasattr(self, '_overlay_manager'):
+            self._overlay_manager.ui.reject()
+            del self._overlay_manager
+
+        self._overlay_manager = AzimuthalOverlayManager(self.ui)
+        self._overlay_manager.show()
```
psavery, line 424: Can we change all of the `overlay_manager` names to `polar_overlay_manager`? This is just to make it clear in the code that this is in the polar view, since the image mode widget has in it logic for all of the different views.

psavery (commented):
Looks really great, thank you for your work Brianna. Just a few suggestions here.

psavery:
Also, let's set the window titles on the dialogs now to "Overlay Manager" and "Azimuthal Overlay Manager", so it is clearer what we are looking at.

hexrd/ui/hexrd_config.py
```diff
@@ -239,6 +243,8 @@ def __init__(self):
         self.polar_angular_grid = None
         self._recent_images = {}
         self.max_cpus = None
+        self.azimuthal_overlays = []
+        self.azimuthal_legend = True
```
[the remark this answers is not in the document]
bnmajor: Makes sense! Fixed.

hexrd/ui/image_canvas.py
```diff
@@ -998,6 +1002,43 @@ def compute_azimuthal_integral_sum(self, scaled=True):
         masked = np.ma.masked_array(pimg, mask=np.isnan(pimg))
         return masked.sum(axis=0) / np.sum(~masked.mask, axis=0)
 
+    def update_azimuthal_plot_overlays(self):
+        while self.azimuthal_overlay_artists:
+            item = self.azimuthal_overlay_artists.pop(0)
+            item['artist'].remove()
+            item['line'].remove()
```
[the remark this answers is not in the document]
bnmajor: Makes sense! Fixed and fixed.

hexrd/ui/image_canvas.py
bnmajor: Done.

hexrd/ui/image_canvas.py
bnmajor: Updated.

hexrd/ui/image_canvas.py
```diff
@@ -998,6 +1002,43 @@ def compute_azimuthal_integral_sum(self, scaled=True):
...
+
+        # Apply new, visible overlays
+        tth, sum = HexrdConfig().last_unscaled_azimuthal_integral_data
+        for overlay in HexrdConfig().azimuthal_overlays:
+            if overlay['visible']:
```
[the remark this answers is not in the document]
bnmajor: Done!

hexrd/ui/image_mode_widget.py
bnmajor: Done.

saransh13:
@bnmajor Can we also add the density in the legends? This property sits in `material.unitcell.density` in g cm^-3. We should only display the density it to 2 decimal places for the legend to fit nicely.

bnmajor:
@saransh13 Sure thing!

psavery:
@bnmajor We are also going to need to make sure there is some signal emitted if the density is modified, and then perform an update on this legend text when the density is modified.

hexrd/ui/azimuthal_overlay_manager.py
bnmajor: This makes sense to me but I did not address this in this PR

hexrd/ui/materials_panel.py
```diff
@@ -276,6 +276,7 @@ def show_overlay_manager(self):
 
     def update_table(self):
         HexrdConfig().update_reflections_tables.emit(self.current_material())
+        HexrdConfig().azimuthal_overlay_modified.emit()
```
bnmajor, line 279: @psavery My thinking here is that I believe any changes that may update the density value will end up hitting this function so we can emit a modified signal just in case? But my logic may be flawed/misguided so I'm open to changes if needed!

saransh13:
Not sure if it's ready for testing but I tried it anyway. Overall looks awesome. Just two comments:

1. The `save plot` option is not working yet. I get the following error even though I used `.png` as recommended: 
Screen Shot 2023-03-28 at 3 34 36 PM

`Traceback (most recent call last):
  File "/Users/soderlind3/hexrdgui/hexrd/ui/image_canvas.py", line 1028, in save_azimuthal_plot
    self.figure.savefig(selected_file, bbox_inches=new_extent)
  File "/Users/soderlind3/opt/miniconda3/envs/hexrd/lib/python3.10/site-packages/matplotlib/figure.py", line 3274, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/Users/soderlind3/opt/miniconda3/envs/hexrd/lib/python3.10/site-packages/matplotlib/backends/backend_qtagg.py", line 81, in print_figure
    super().print_figure(*args, **kwargs)
  File "/Users/soderlind3/opt/miniconda3/envs/hexrd/lib/python3.10/site-packages/matplotlib/backend_bases.py", line 2285, in print_figure
    with cbook._setattr_cm(self, manager=None), \
  File "/Users/soderlind3/opt/miniconda3/envs/hexrd/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/Users/soderlind3/opt/miniconda3/envs/hexrd/lib/python3.10/site-packages/matplotlib/backend_bases.py", line 2192, in _switch_canvas_and_return_print_method
    raise ValueError(
ValueError: Format 'png,' is not supported (supported formats: eps, jpeg, jpg, pdf, pgf, png, ps, raw, rgba, svg, svgz, tif, tiff, webp)`

Are the other formats not supported because of the transparency in the plot?

The second thing is a slight modification to the units in the legend. We should use the `tex` interpreter in pyplot to get the units to show up correctly (same as in the properties tab of the material) Image below

Screen Shot 2023-03-28 at 3 38 55 PM

hexrd/ui/materials_panel.py
psavery: Hey @bnmajor! I think it's a little better design for the materials panel to not be aware of other parts of the program, and basically just emit a signal that says "This was modified!", and have other parts of the program listen for that signal to respond.

On that note, there is a `HexrdConfig().active_material_modified` signal. The active material is whatever one is currently displayed in the materials panel, which is usually the only one that gets modified since the user can only see that one in the panel. If you hook into that signal, you will get the same behavior as what you have here, and I think that might be better.

However, there are some cases where things like lattice parameters get modified by calibration. I'm thinking we should add a `HexrdConfig().material_modified(str)` signal, where the `str` is the name of the material that was modified (and we should emit this anytime we change something about the material behind the scenes). But we don't necessarily need to do that for this PR.

bnmajor:
@saransh13 Thanks! So that error was produced by a syntax error I made (no commas in the list of file extensions!) and should be fixed now. I've also removed that list altogether since that was based on what appears to be an incomplete list of supported file types. This leaves it a bit more open-ended for the user to save to the image format that they prefer.

The units have been corrected as well. Just a heads up that in the case of the legend the exponent was is kind of squished. Presumably users will understand that it is centimeters cubed though and that will not be an issue. 
legend

hexrd/ui/calibration/auto/powder_runner.py
```diff
@@ -222,14 +222,11 @@ def update_config(self):
             #        this array is the reduce lattice parameter set.
             refined_lattice_params = self.ic.full_params[self.ic.npi:]
             self.material.latticeParameters = refined_lattice_params
+            HexrdConfig().material_modified(self.material)
```
psavery, line 225: I think this should be `self.material.name` instead.

psavery: Oh, and we need the `emit()` too. It should also be `HexrdConfig().material_modified.emit(self.material.name)`

hexrd/ui/calibration/wppf_runner.py
```diff
@@ -82,9 +82,7 @@ def write_lattice_params_to_materials(self):
 
             mat.latticeParameters = lparms
             HexrdConfig().flag_overlay_updates_for_material(name)
-
-            if mat is HexrdConfig().active_material:
-                HexrdConfig().active_material_modified.emit()
+            HexrdConfig().material_modified(name)
```
psavery, line 85: Should be `HexrdConfig().material_modified.emit(mat_name)`

hexrd/ui/calibration/calibration_runner.py
```diff
@@ -390,8 +390,7 @@ def run_calibration(self):
                 mat = materials[mat_name]
                 mat.latticeParameters = calibrator.params
                 HexrdConfig().flag_overlay_updates_for_material(mat_name)
-                if mat is HexrdConfig().active_material:
-                    HexrdConfig().active_material_modified.emit()
+                HexrdConfig().material_modified(mat_name)
```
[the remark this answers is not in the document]
bnmajor: Missed the emit on all of them.... Fixed now.

hexrd/ui/calibration/auto/powder_runner.py
bnmajor: Both fixed now

hexrd/ui/calibration/wppf_runner.py
bnmajor: Fixed

hexrd/ui/hexrd_config.py
```diff
@@ -1579,6 +1582,10 @@ def _set_materials(self, materials):
     def material(self, name):
         return self.config['materials']['materials'].get(name)
 
+    def check_activate_material_changed(self, material_name):
```
psavery, line 1585: Can we change `activate` to `active`?

hexrd/ui/hexrd_config.py
```diff
@@ -1579,6 +1582,10 @@ def _set_materials(self, materials):
...
+        if material_name == self.active_material:
```
psavery, line 1586: This should be `self.active_material_name` instead (`self.active_material` returns the material object)

hexrd/ui/hexrd_config.py
```diff
@@ -1579,6 +1582,10 @@ def _set_materials(self, materials):
...
+            self.active_material_changed.emit()
```
psavery, line 1587: Should be `self.active_material_modified.emit()` instead.

hexrd/ui/materials_panel.py
psavery: Awesome, thank you @bnmajor!

hexrd/ui/hexrd_config.py
bnmajor: Sorry about that, fixed.

hexrd/ui/hexrd_config.py
bnmajor: Fixed

hexrd/ui/hexrd_config.py
bnmajor: Thanks, fixed.

saransh13:
The scale factor is limited to a maximum value of 1. This should not be the case.

bnmajor:
> The scale factor is limited to a maximum value of 1. This should not be the case.

It's now set to a max of 100. Easy to change to any arbitrary number though!

hexrd/ui/resources/ui/azimuthal_overlay_editor.ui
```diff
@@ -76,17 +76,17 @@
...
        <property name="singleStep">
-        <double>0.010000000000000</double>
+        <number>1</number>
        </property>
        <property name="value">
-        <double>1.000000000000000</double>
+        <number>1</number>
+       </property>
+       <property name="decimals" stdset="0">
+        <number>3</number>
```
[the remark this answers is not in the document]
bnmajor: Sounds, good. I'll change that.

saransh13:
> > > The scale factor is limited to a maximum value of 1. This should not be the case.
> > 
> > 
> > It's now set to a max of 100. Easy to change to any arbitrary number though!
> 
> @saransh13 Does 100 sound like a reasonable max? Or should we make it even higher, like 10000?

Could be higher. Maybe `1E9` sounds like it should be ok.

saransh13 (approved):
I have tested out the features, and everything seems to work. I just made a small change to the maximum value in the scale box. All the features work for me. I'll wait for Patrick to give the final nod, but LGTM!

hexrd/ui/image_canvas.py
```diff
@@ -998,6 +1005,69 @@ def compute_azimuthal_integral_sum(self, scaled=True):
...
+        if not selected_file:
+            return
+        # Find and invert the physical inches from the bottom left corner
+        size = self.figure.dpi_scale_trans.inverted()
+        # Get the bbox for the second plot
+        extent = self.azimuthal_integral_axis.get_window_extent().transformed(size)
+        # The bbox does not include the axis so manually scale it up so it does
+        new_extent = extent.from_extents([[0, 0], [extent.xmax*1.05, extent.ymax*1.05]])
+        # Save the clipped region of the figure
+        self.figure.savefig(selected_file, bbox_inches=new_extent)
```
psavery, line 1028: Interesting thing we had to do here!

psavery:
@bnmajor This ended up being more work than expected, but you did an awesome job with this. Thank you so much for all of your hard work! This is a great improvement.

PR #1377, opened 2023-03-28
Set Qt.Tool flag on masking dialogs

### LINKED ISSUE #1371: Masking tool hidden by main window
When you want to add a mask element, the window that pops up is hidden by the main window. It would save time if that was not the case (ie, have the mask tool window open up to the left of the main window). I usually end up chasing the main window across the screen and in a few clicks have to start over again. Is it possible to have the mask work more like Dioptas where there's a separate tab with dedicated controls in the gui?

bnmajor:
Fixes #1371


saransh13:
The `hand drawn mask` dialog still hides behind the main window. All other mask windows seem to stay on top for me!

psavery:
@bnmajor Looks good so far. Can you add this to the `HandDrawnMaskDialog` as well?

bnmajor:
Oof, completely overlooked that one, sorry! Fixed now.

psavery (approved):
LGTM

PR #1389, opened 2023-04-05
Allow state file argument to be instrument also

psavery:
This allows us to alternatively load in an instrument config file by command line instead of a state file.


bnmajor (approved):
Wonderful!

Issue #1368, opened 2023-03-27
Larger polar/azimuthal default ranges in powder/laue calibration tool


akrygier:
I suggest 60-90 degrees in azimuth/15 degrees in 2theta.

psavery:
@akrygier Do you mean for the zoom box?

PR #1390, opened 2023-04-05
Increase default zoom box ranges

psavery:
Two theta is now 15 degrees, and eta is now 60 degrees.

Fixes: #1368


bnmajor (approved):
Nice!

Issue #1365, opened 2023-03-27
round off displayed values in gui


akrygier:
The colormap limits and 2theta range text boxes currently include a lot of decimal points (more than 10). Can this be rounded off to 1-2 decimal points?

psavery:
This sounds good. I think probably the best solution is for us to just do some rounding for when these are automatically generated (such as [here](https://github.com/HEXRD/hexrdgui/blob/a21da7f8e00ba1bb120b40c617b2536fb67671e2/hexrd/ui/color_map_editor.py#L162-L175) and [here](https://github.com/HEXRD/hexrdgui/blob/a21da7f8e00ba1bb120b40c617b2536fb67671e2/hexrd/ui/brightness_contrast_editor.py#L186-L192)).

The `ScientificDoubleSpinBox` widgets throughout the program are set up to not display trailing zeros, so rounding the numbers themselves will make it display fewer digits.

akrygier:
Yes, I agree with that.

PR #1391, opened 2023-04-05
Round auto-generated values for nicer display

psavery:
The auto-generated values were being displayed to many decimal places.

Round them so that they look nicer. They don't really need that many decimal places anyways.

Fixes: #1365


bnmajor (approved):
Looks good!

Issue #1392, opened 2023-04-07
Normalize by minimum solid angle of all detectors in the instrument


saransh13:
After applying the `pixel solid angle correction`, the intensities become very large since the solid angles of each pixels are small numbers. To get around this, the solid angles are normalized by the minimum value [here](https://github.com/HEXRD/hexrd/blob/da7322e88498a1987ac85c36aa10d7cbf1448f39/hexrd/instrument.py#L4231-L4233). However, this normalizes each detector separately and can lead to one of the detectors having a much higher intensity than the rest. The normalization should be done for the whole collection of detectors in the instrument with the minimum value of the entire instrument.

PR #1393, opened 2023-04-07
Pixel solid angle correction fix

saransh13:
Fixes issue #1392. Also relies on [hexrd PR#510](https://github.com/HEXRD/hexrd/pull/510).

Fixes: #1392

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -736,9 +736,16 @@ class HexrdConfig(QObject, metaclass=QSingleton):
         instr = create_hedm_instrument()
 
         if HexrdConfig().apply_pixel_solid_angle_correction:
+            sangle = dict.fromkeys(images_dict.keys())
+            mi = 1E12 # some large number
+            # normalize by minimum of the entire instrument
+            # not each detector individually
             for name, img in images_dict.items():
                 panel = instr.detectors[name]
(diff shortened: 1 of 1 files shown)


hexrd/ui/hexrd_config.py
```diff
@@ -736,10 +736,17 @@ def intensity_corrected_images_dict(self):
...
+            # not each detector individually
             for name, img in images_dict.items():
                 panel = instr.detectors[name]
-                images_dict[name] = img / panel.pixel_solid_angles
-
+                sangle[name] = panel.pixel_solid_angles
+                mi = np.min((mi, sangle[name].min()))
+            for name, img in images_dict.items():
+                images_dict[name] = mi * img / sangle[name]
+            del sangle
```
psavery, line 749: You don't need to delete `sangle` here. It will be deleted automatically when it goes out of scope.

hexrd/ui/hexrd_config.py
```diff
@@ -736,10 +736,17 @@ def intensity_corrected_images_dict(self):
         instr = create_hedm_instrument()
 
         if HexrdConfig().apply_pixel_solid_angle_correction:
+            sangle = dict.fromkeys(images_dict.keys())
+            mi = 1E12 # some large number
```
psavery, line 740: I would typically do `np.finfo(np.float64).max`, but this is probably okay too

psavery (approved):

PR #1395, opened 2023-04-12
Cylindrical detector support

psavery:
This depends on hexrd/hexrd#504.

This supercedes #1360.


psavery:
I'm going to go ahead and merge this one since https://github.com/HEXRD/hexrd/pull/504 is now merged. This will prevent merge conflicts.

There are still some issues we need to fix with this PR, but we will fix them later. Right now, a warning is being printed when a cylindrical detector is being created noting that not everything works properly. That is good enough for now.

PR #1396, opened 2023-04-13
Make fixes for new make_reflection_patches()

psavery:
In hexrd/hexrd#511, `make_reflection_patches()` is now returning an iterator instead of a list.

Instead of getting the number of points based upon the length of the output, compute it based upon the input. From reading `make_reflection_patches()`, it appears that the number of patches is equal to the number of input angles.


bnmajor (approved):
LGTM

PR #1407, opened 2023-04-18
Simplify llnl import

### LINKED ISSUE #1359: IMAGE PLATE 3 for TARDIS import

### LINKED ISSUE #1363: Automatically read TARDIS hdf and automatically flip images.
TARDIS hdf files store the data in a similar way, ie, XXX/DATA/DATA. It would save clicks if we could just have that read automatically. 

I attached a compressed python method that I use in other code that may be a good starting place. Not all of the if statements are necessary for TARDIS.

[load_nif_im_from_h5.py.zip](https://github.com/HEXRD/hexrdgui/files/11083683/load_nif_im_from_h5.py.zip)

### LINKED ISSUE #1374: Draw the detector outlines in TARDIS import by default
It's an extra click to do this, and I do it every time.

bnmajor:
This adds a handful of features to simplify the LLNL workflow:

- The template is automatically applied when a detector is selected and an image is loaded (Fixes #1374)
- The cropped IP3 boundary uses the dashed line instead of a solid one (Fixes #1359)
- Automatically prompt user to apply the default transform associated with the current instrument/detector (Fixes #1363)
- Adds a help button to the documentation

### DIFF
--- a/hexrd/ui/llnl_import_tool_dialog.py
+++ b/hexrd/ui/llnl_import_tool_dialog.py
@@ -64,12 +64,10 @@ class LLNLImportToolDialog(QObject):
             self.instrument_selected)
         self.ui.load.clicked.connect(self.load_images)
         self.ui.detectors.currentIndexChanged.connect(self.detector_selected)
-        self.ui.add_template.clicked.connect(self.add_template)
         self.ui.trans.clicked.connect(self.setup_translate)
         self.ui.rotate.clicked.connect(self.setup_rotate)
         self.ui.add_transform.clicked.connect(self.add_transform)
-        self.ui.button_box.accepted.connect(self.crop_and_mask)
-        self.ui.button_box.rejected.connect(self.clear)
+        self.ui.accept_template.clicked.connect(self.crop_and_mask)
         self.ui.complete.clicked.connect(self.completed)
         self.ui.bb_height.valueChanged.connect(self.update_bbox_height)
         self.ui.bb_width.valueChanged.connect(self.update_bbox_width)
@@ -123,6 +121,7 @@ class LLNLImportToolDialog(QObject):
         return True
 
     def set_detector_options(self):
+        self.detectors.clear()
         for det, vals in self.defaults['detectors'].items():
             self.detector_defaults[det] = vals['transform']
             self.detectors.append(det)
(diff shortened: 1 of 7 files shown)


hexrd/ui/llnl_import_tool_dialog.py
```diff
@@ -138,6 +141,11 @@ def load_config(self):
         self.config_file = selected_file if selected_file else None
         self.ui.config_file_label.setText(os.path.basename(self.config_file))
         self.ui.config_file_label.setToolTip(self.config_file)
+        if self.ui.instrument.isEnabled():
+            # Only set the instrument config once
+            success = self.get_instrument_defaults()
+            if not success:
+                return
```
psavery, line 148: Is there supposed to be different behavior depending on the value of `success`? It looks like this function returns anyways after this block ends.

hexrd/ui/llnl_import_tool_dialog.py
```diff
@@ -177,6 +186,11 @@ def update_config_selection(self, checked):
         elif not self.ui.default_config.isChecked():
             self.enable_widgets(self.ui.load_config, self.ui.config_file_label,
                                 enabled=True)
+        if self.ui.instrument.isEnabled():
+            # Only set the instrument config once
+            success = self.get_instrument_defaults()
+            if not success:
+                return
```
psavery, line 193: Same here

hexrd/ui/llnl_import_tool_dialog.py
bnmajor: Nope, careless copy-paste. Fixed.

hexrd/ui/llnl_import_tool_dialog.py
bnmajor: Same careless copy-paste. Fixed.

psavery (approved):
LGTM. We will keep improving this in further iterations in the near future.

PR #1409, opened 2023-04-18
Add a function to try and guess the hdf5 path

bnmajor:
Currently designed for the LLNL workflow but could be expanded for any other general cases. Adapted from the script provided by Andy Krygier (see [comment](https://github.com/HEXRD/hexrdgui/issues/1363#issue-1642962930)). Addresses part of #1363


hexrd/ui/image_file_manager.py
```diff
@@ -143,14 +143,49 @@ def is_hdf(self, extension):
...
+            elif 'TIME_ADJUSTED_IMAGE' in h5['ATTRIBUTES']:
+                location = 'TIME_ADJUSTED_IMAGE'
+            elif 'PSL_IMAGE' in h5['ATTRIBUTES']:
+                location = 'PSL_IMAGE'
+        elif 'DATA' in h5:
+            location = 'DATA'
+        h5.close()
+        path=f'ATTRIBUTES/{location}/DATA'
+        imageseries.open(f, 'hdf5', path=path, dataname=dataname)
+        return path, dataname
```
psavery, line 173: Let's clean this up a little bit. I will suggest some changes soon.

hexrd/ui/image_file_manager.py
```diff
@@ -143,14 +143,49 @@ def is_hdf(self, extension):
...
             return True
         except:
-            return False
+            try:
+                path, dataname = self.guess_path(f)
+                HexrdConfig().hdf5_path = [path, dataname]
+                self.path = HexrdConfig().hdf5_path
+                return True
+            except:
+                return False
```
psavery, line 188:
```suggestion
            try:
                path, dataname = self.guess_path(f)
            except:
                return False
            else:
                HexrdConfig().hdf5_path = [path, dataname]
                self.path = HexrdConfig().hdf5_path
                return True
```
Let's minimize the section in the `try` block to whatever we are expecting to raise an exception.

We should also avoid catching bare exceptions, since there may be valid exceptions that need to be propagated to the user. We should figure out exactly what exceptions we are expecting and catch those.

bnmajor:
@psavery TIL that you could check for a path in an hdf5 file with a simple "in", so I've pushed a new commit. the logic is simplified and there are not try/catch statements to worry about.

hexrd/ui/constants.py
```diff
@@ -71,3 +71,14 @@ class OverlayType(Enum):
...
+KNOWN_HDF5_PATHS = [
+    ['ATTRIBUTES/TARGET_ORIENTED_IMAGE/DATA', 'DATA'],
+    ['ATTRIBUTES/PINHOLE_ORIENTED_IMAGE/DATA', 'DATA'],
+    ['ATTRIBUTES/PSL_FADE_CORR_IMAGE/DATA', 'DATA'],
+    ['ATTRIBUTES/HDR_IMAGE/DATA', 'DATA'],
+    ['ATTRIBUTES/CORR_IMAGE/DATA', 'DATA'],
+    ['ATTRIBUTES/TIME_ADJUSTED_IMAGE/DATA', 'DATA'],
+    ['ATTRIBUTES/PSL_IMAGE/DATA', 'DATA'],
+    ['DATA', 'DATA'],
+]
```
psavery, line 84: Yes, this was more so the kind of thing I was thinking of :slightly_smiling_face:

hexrd/ui/image_file_manager.py
```diff
@@ -143,49 +144,17 @@ def is_hdf(self, extension):
...
     def path_exists(self, f):
-        try:
-            path, dataname = HexrdConfig().hdf5_path
-            imageseries.open(f, 'hdf5', path=path, dataname=dataname)
-            self.path = HexrdConfig().hdf5_path
-            return True
-        except:
-            try:
-                path, dataname = self.guess_path(f)
+        all_paths = [*constants.KNOWN_HDF5_PATHS, HexrdConfig().hdf5_path]
```
psavery, line 148: Should we put the user's saved path first? It seems to me that should get priority.

hexrd/ui/image_file_manager.py
```diff
@@ -143,49 +144,17 @@ def is_hdf(self, extension):
...
+        h5 = h5py.File(f, 'r')
```
psavery, line 149: Let's use a context manager.

```python
with h5py.File(f, 'r') as hf:
    ...
```

Then indent all of the below code and remove `h5.close()` in each case.

The context manager is better here because if an exception is raised before the file is closed, it may never be closed, and it may make the user's HEXRDGUI session not work anymore.

hexrd/ui/image_file_manager.py
bnmajor: Yes, fixed!

hexrd/ui/image_file_manager.py
bnmajor: Fixed

psavery (approved):
LGTM!

PR #1411, opened 2023-04-18
Do not remember azimuthal overlays between sessions

bnmajor:


hexrd/ui/hexrd_config.py
```diff
@@ -335,7 +335,6 @@ def _attributes_to_persist(self):
             ('overlays_dictified', []),
             ('_polar_tth_distortion_overlay_name', None),
             ('_recent_images', {}),
-            ('azimuthal_overlays', []),
```
psavery, line 338: So, the tricky thing here is that I think this will also not save the azimuthal overlays in the state file, which we do want.

Can you verify that this does not save the azimuthal overlays between sessions, but *does* save the azimuthal overlays to state files? If not, we may need to think of something different.

bnmajor: Yeah, you're right. That would not save it to the state. I've update the PR now to just skip that particular key when loading the state from the QSettings. Appears to work as expected, forgetting between sessions but using the settings when loading from a state file.

hexrd/ui/hexrd_config.py
```diff
@@ -363,6 +363,8 @@ def _save_state_to_settings(self, state, settings):
     def _load_state_from_settings(self, settings):
         state = {}
         for name, default in self._attributes_to_persist():
+            if name == 'azimuthal_overlays':
+                continue
```
psavery, line 367: Can you leave a comment here that we are doing this so that `azimuthal_overlays` get saved to state but do not get saved to the QSettings in between sessions?

bnmajor: Done!

psavery (approved):
LGTM

PR #1413, opened 2023-04-26
Fix places where `ArtistList.remove()` was used

psavery:
`ArtistList.remove()` is no longer a method in matplotlib, as of 3.7.0 (it was made immutable, see [here](https://matplotlib.org/stable/api/axes_api.html#matplotlib.axes.Axes.ArtistList), and it may be turned into a tuple soon).

As such, we encounter an error in these places for new versions of matplotlib.

Instead, we can call `remove()` on the artist itself, and this removes it from the `ArtistList` automatically. Do that instead.

This fixes some of the issues mentioned in #1397 where the rectangle/ellipse masking was not working.

This also potentially fixes another issue with the LLNL Import tool removing boundaries.

### DIFF
--- a/hexrd/ui/interactive_template.py
+++ b/hexrd/ui/interactive_template.py
@@ -109,7 +109,7 @@ class InteractiveTemplate:
 
     def clear(self):
         if self.shape in self.raw_axes.patches:
-            self.raw_axes.patches.remove(self.shape)
+            self.shape.remove()
             self.redraw()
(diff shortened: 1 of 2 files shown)


bnmajor (approved):
Thank you!

psavery:
Thanks for your extra commits, @bnmajor! Hopefully this takes care of all of the issues.

Issue #1417, opened 2023-04-26
Remember last pinhole mask setting in widget


psavery:
See discussion here: https://github.com/HEXRD/hexrdgui/discussions/1397#discussioncomment-5736208

PR #1418, opened 2023-04-26
Make pinhole mask dialog more interactive

psavery:
The pinhole mask settings are now saved so that they persist both in the same session and in between sessions.

Also, the dialog now stays open with an "Apply" button. When this is clicked, the previous pinhole mask is replaced with a mask generated with the new settings. This allows users to modify the values, create the pinhole mask, and then repeat in case the values need to be modified.

https://user-images.githubusercontent.com/9558430/234713408-bb08f7ac-79b4-488d-b246-441da0a62e18.mp4

Fixes: #1417


bnmajor (approved):
LGTM

PR #1419, opened 2023-04-27
Add check for panel buffer size

psavery:
Sometimes, we have a panel buffer that does not match the size of the detector.

To prevent this from happening entirely, add a check that will delete the panel buffer if the size does not match.

It would be nice to not have this check, but it is sometimes difficult to track down the many places where this synchronization issue may occur. Let's keep this check unless we figure out a better way.

Fixes: #1381
Fixes: #1398


bnmajor (approved):
Nice!

PR #1421, opened 2023-04-27
Do not show frame for LLNL import tool

psavery:
The frame goes away if you dock it, and it is confusing because dragging the frame will not allow you to dock it.

Just remove it altogether.


bnmajor (approved):
Perfect!

PR #1423, opened 2023-04-27
Save to NPZ for HEDM workflow output images

psavery:
The unaggregated imageseries from the HEDM workflow were previously being saved as HDF5 files. This is a problem, because that data can be *very* large when it is fully expanded.

Instead, we should be saving it in the sparse frame-cache NPZ files.


psavery:
@donald-e-boyce What would be the best threshold to use when saving the NPZ files at the end of this workflow?

bnmajor (approved):
Awesome!

Issue #1405, opened 2023-04-18
Add ability to compute structure factor relative to another material


psavery:
This would be an option in the reflections table.

PR #1425, opened 2023-04-28
Add relative scaling for structure factor

psavery:
This adds the ability to the reflections table to scale the structure
factor based upon a different material's structure factor.
    
This can be helpful, for instance, to compare how strong the signal will
be of one material compared to another.

https://user-images.githubusercontent.com/9558430/235229614-8dd70335-5a4c-40f1-9cdc-7cf10e6b7e40.mp4

Fixes: #1405


bnmajor (approved):
LGTM

PR #1427, opened 2023-04-28
Add a reflections selection helper dialog

psavery:
This dialog allows conditions to be specified so that when "Apply" is clicked, reflections matching the specified conditions will be selected automatically.

image

Fixes: #1404


bnmajor (approved):
Looks great!

PR #1429, opened 2023-05-01
fix TARDIS image loading issue

saransh13:
an empty list was being appended causing loading error. discard HexrdConfig().hdf5_path if it is an empty lsit

### DIFF
--- a/hexrd/ui/image_file_manager.py
+++ b/hexrd/ui/image_file_manager.py
@@ -157,10 +157,8 @@ class ImageFileManager(metaclass=Singleton):
 
     def hdf5_path_exists(self, f):
         all_paths = []
-        if HexrdConfig().hdf5_path is not None:
-            # added by SS 05/01/23 to check against empty list
-            if HexrdConfig().hdf5_path:
-                all_paths.append(HexrdConfig().hdf5_path)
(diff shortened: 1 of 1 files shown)


hexrd/ui/image_file_manager.py
```diff
@@ -158,9 +158,10 @@ def hdf_path_exists(self, f):
     def hdf5_path_exists(self, f):
         all_paths = []
         if HexrdConfig().hdf5_path is not None:
-            all_paths.append(HexrdConfig().hdf5_path)
+            # added by SS 05/01/23 to check against empty list
+            if HexrdConfig().hdf5_path:
+                all_paths.append(HexrdConfig().hdf5_path)
```
bnmajor, line 163: This can be simplified a bit further by dropping the `if HexrdConfig().hdf5_path is not None` condition to just be:
```
if HexrdConfig().hdf5_path:
    all_paths.append(HexrdConfig().hdf5_path)
```

bnmajor:
Thanks @saransh13! I think we can drop the first conditional statement, otherwise looks good!

bnmajor (approved):
LGTM!

psavery (approved):
LGTM

Issue #1406, opened 2023-04-18
Add ability to set the minimum structure factor in the fit grains options


psavery:
This would automatically select HKLs in fit-grains based upon the selected structure factor.

Issue #1357, opened 2023-03-14
Improve Fit Grains HKL Selection


psavery:
![fit_grains_dialog](https://user-images.githubusercontent.com/9558430/224875001-0087c335-87ab-4e58-ba37-401268ff25d4.png)

Currently, in the `fit-grains` dialog (pictured above), if you want to mimic the behavior of the `hexrd` command line, you must click the `Choose HKLs` button and modify the HKLs at this point.

The reason you must modify the HKLs here is because earlier, when the eta omega maps are generated, you choose HKLs at that point for the eta omega map generation, and typically, you only choose a few that have low two theta values. Those same HKLs are what are selected when you get to the fit grains options.

However, in the `hexrd` command line, you specify a `min_sfac_ratio` that gets used to exclude some of the HKLs. You then typically pick HKLs to use for the eta omega map generation from this filtered down list, and then choose HKL seeds from that list. However, in the command line `hexrd`, fit grains will use all of the HKLs except those excluded by the `min_sfac_ratio`.

In the GUI, however, you select HKLs from the table separately for the two different steps (first for generating the eta omega maps, and then again for performing fit grains). If you want to mimic the `hexrd` command line behavior in `fit-grains`, once you get to the fit-grains options, you must click `Choose HKLs`, sort the reflections table by the `min_sfac_ratio` (that is the `|F|²`), and then shift-click to manually select all HKLs that are above a certain `min_sfac_ratio` (note that a `min_sfac_ratio` of `0.05` in the `hexrd` command line is equivalent to `5` in the GUI). If you do this, then the results between the `hexrd` command line and the GUI are essentially identical (except small floating point precision). See attached image below as an example of selecting all HKLs in the table above a `min_sfac_ratio` of `2.5`.

![f_squared_select](https://user-images.githubusercontent.com/9558430/224876825-0f00da6e-bb51-445b-9294-f0f873008f07.png)

I wonder if this is non-intuitive and if there is an easier way to do this. For instance, should we have an option at this point to enter in the `min_sfac_ratio`, and then automatically select HKLs at this point based upon what the user inputs?

Maybe this will help alleviate #1347 some as well.

What do you think, @donald-e-boyce?

Issue #1347, opened 2023-02-24
Make sure HEXRDGUI and HEXRD options are in sync


psavery:
Some options are available in one but not the other.

PR #1430, opened 2023-05-01
Add option to specify min sfac for fit grains

psavery:
In the HEXRD CLI for fit-grains, the HKLs are, right now, typically specified via min sfac.

Add the ability to set these in the GUI as well so that the settings are more consistent with those of the GUI.

Addresses part of: #1347

Fixes: #1357
Fixes: #1406


bnmajor (approved):
LGTM

PR #1431, opened 2023-05-02
Remove Change Image Folder button

bnmajor:
This was added in the very beginning when there was a specific file structure expected for data to be loaded. Now the Select Image Files option takes care of figuring out where all of the image files are located and this is no longer needed and just adds a bit of confusion.

New dialog layout:
new_simple_image_series


psavery (approved):
LGTM

Issue #1403, opened 2023-04-18
Add option to convert x-axis of polar view to scattering vector Q.


saransh13:
Add an option to convert to Q space in Angstrom^-1. The formula is Q = 4 * Pi * sin(theta)/lambda.
lambda is the wavelength in Angstrom, and theta = current x-axis value/2 in radians.

PR #1438, opened 2023-05-05
Add option to display polar x-axis as Q

psavery:
This adds an option to change the polar x-axis type to the Q scattering vector. This also adds infrastructure so that we can more easily add more polar x-axis types in the future if needed.

This also adds `Q` to the mouse hover information in all views.

image

Fixes: #1403


saransh13 (approved):
LGTM!

Issue #1367, opened 2023-03-27
Reflections not selected


akrygier:
I frequently find that only a subset of reflection planes are activate when I turn on a crystal for plotting. (e.g., the [110] reflection of a bcc/229 structure is not displayed when I add it as an overlay in the overlay manager window.

akrygier:
I should add that selecting/deselecting reflection planes is buggy. E.g., I try to turn off a specific reflection and 2 or 3 more get turned off for no apparent reason.

psavery:
> I should add that selecting/deselecting reflection planes is buggy. E.g., I try to turn off a specific reflection and 2 or 3 more get turned off for no apparent reason.

Andy, I think I have found a fix for this, and it is in PR #1439.

psavery:
> I frequently find that only a subset of reflection planes are activate when I turn on a crystal for plotting. (e.g., the [110] reflection of a bcc/229 structure is not displayed when I add it as an overlay in the overlay manager window.

As for this one: if you modify the space group, it will only automatically select HKLs that were selected in the previous space group. So any HKLs in the new space group that weren't a part of the previous space group won't be selected.

Maybe we should modify this behavior, though? When we change space groups, should we just auto-select like, the first 5 HKLs? What do you think @saransh13

PR #1439, opened 2023-05-05
Do not update reflections table during selection

psavery:
Reflections selection has been a bit buggy. It appears that the issue is that updating the reflections table during selection messes up something with Qt's selection model.

But we do not need to update the reflections table during selection. So do not do it. This appears to fix the bugginess.

Fixes: #1367


saransh13 (approved):
LGTM!

PR #1441, opened 2023-05-09
Clean up WPPF options dialog enable states

psavery:
The min/max are now disabled if "vary" is False, and the number of refinement steps is now disabled if "Rietveld" is selected.


saransh13 (approved):
changes LGTM!

PR #1442, opened 2023-05-10
Support varied file names when using 'image-files' adapter

bnmajor:
Fixes a bug revealed in testing HEXRD/hexrd#523. Files loaded with the `image-files` Image Series Adapter are passed to `glob` but will fail if they contain special characters. Use `glob.escape` to parse the string and prevent errors loading files.

### DIFF
--- a/hexrd/ui/image_file_manager.py
+++ b/hexrd/ui/image_file_manager.py
@@ -1,3 +1,4 @@
+import glob
 import os
 import numpy as np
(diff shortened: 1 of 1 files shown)


psavery (approved):
I think this makes sense, because the `files` argument here is intended to accept glob patterns, and since we are definitely not providing it glob patterns, we should escape them.

PR #1443, opened 2023-05-10
Perform deep rerender after calibration

psavery:
Otherwise, the current view may look incorrect.

This modifies several places in the main window to also perform a deep rerender when needed.


bnmajor (approved):
LGTM

PR #1445, opened 2023-05-11
Do not reset material exclusions for export HEDM

psavery:
As of hexrd/hexrd#513, the material exclusions are now reset by default when running fit-grains.

In hexrd/hexrd#520, they are now reset in the materials as well. However, that PR also introduces the option to not reset exclusions, both in fit-grains and in the material config.

Set that option when we export the config so that we keep the exclusions that we saved on the materials. This is so that we are able to reproduce what was set in the GUI when running the HEDM workflow from the command line.

Depends on: hexrd/hexrd#520

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -986,6 +986,9 @@ class HexrdConfig(QObject, metaclass=QSingleton):
 
         recursive_key_check(self.indexing_config, cfg)
 
+        # Make sure the exclusions do not get reset in fit-grains
+        cfg['fit_grains']['reset_exclusions'] = False
+
         current_material = self.indexing_config['_selected_material']
         selected_material = self.material(current_material)
         plane_data = selected_material.planeData
@@ -1001,7 +1004,7 @@ class HexrdConfig(QObject, metaclass=QSingleton):
             'active': current_material,
             'dmin': selected_material.dmin.getVal('angstrom'),
             'tth_width': tth_width,
-            'min_sfac_ratio': None
+            'reset_exclusions': False,
         }
 
         data = []
--- a/hexrd/ui/indexing/create_config.py
+++ b/hexrd/ui/indexing/create_config.py
@@ -34,9 +34,14 @@ def create_indexing_config():
         # Set the max number of CPUs
         indexing_config['multiprocessing'] = HexrdConfig().max_cpus
 
+    # Make sure exclusions are not reset
+    fit_grains_config = indexing_config.setdefault('fit_grains', {})
+    disable_exclusions_reset(fit_grains_config)
+
     # Set the active material on the config
-    tmp = indexing_config.setdefault('material', {})
-    tmp['active'] = material.name
+    material_config = indexing_config.setdefault('material', {})
+    material_config['active'] = material.name
+    disable_exclusions_reset(material_config)
 
     # Create the root config from the indexing config dict
     config = RootConfig(indexing_config)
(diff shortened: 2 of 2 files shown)


donald-e-boyce:
It looks like you need to do more than just modify the exported YAML config files. It seems you need to modify the runtime config file as well.  Here are the results from running in the GUI and then running the exported workflow.

I selected HKL by 10% cutoff for the structure factor, giving 27 HKLs.
GUI results.  
```
# grain ID    completeness  chi^2         exp_map_c[0]             exp_map_c[1]             exp_map_c[2]             t_vec_c[0]               t_vec_c[1]               t_vec_c[2]               inv(V_s)[0,0]            inv(V_s)[1,1]            inv(V_s)[2,2]            inv(V_s)[1,2]*sqrt(2)    inv(V_s)[0,2]*sqrt(2)    inv(V_s)[0,2]*sqrt(2)    ln(V_s)[0,0]             ln(V_s)[1,1]             ln(V_s)[2,2]             ln(V_s)[1,2]             ln(V_s)[0,2]             ln(V_s)[0,1]           
0             0.946759      5.168142e-04  6.6921017793913029e-01   -9.8632798711803560e-01  7.3664741577593373e-01   4.9752682755868890e-03   3.3522579406223044e-04   3.2536082001188922e-03   1.0001732407337502e+00   1.0001625076450116e+00   1.0001855440167606e+00   -2.3884731899696052e-05  1.1939429713575515e-06   -1.7858791855691052e-05  -1.7322564924433448e-04  -1.6249421979495530e-04  -1.8552666267274479e-04  1.6886111946847880e-05   -8.4398714650520851e-07  1.2625946128388609e-05 
```
CLI results for exported workflow.
```
# grain ID    completeness  chi^2         exp_map_c[0]             exp_map_c[1]             exp_map_c[2]             t_vec_c[0]               t_vec_c[1]               t_vec_c[2]               inv(V_s)[0,0]            inv(V_s)[1,1]            inv(V_s)[2,2]            inv(V_s)[1,2]*sqrt(2)    inv(V_s)[0,2]*sqrt(2)    inv(V_s)[0,1]*sqrt(2)    ln(V_s)[0,0]             ln(V_s)[1,1]             ln(V_s)[2,2]             ln(V_s)[1,2]             ln(V_s)[0,2]             ln(V_s)[0,1]           
0             1.000000      1.506552e-03  6.6916796368088238e-01   -9.8637568787414309e-01  7.3667013800667813e-01   4.8412067974358611e-03   1.0862225610760879e-03   3.7213447428033329e-03   1.0001876198791395e+00   1.0001492887619063e+00   1.0000878601448275e+00   -8.3049009018846905e-05  -4.9128345318310297e-05  -1.8355349512998343e-05  -1.8760159330045761e-04  -1.4927581138746992e-04  -8.7853958169964474e-05  5.8717780537333891e-05   3.4734582909044563e-05   1.2978025827895392e-05 
```

The exported workflow got it right I think, with 100% completeness.  The GUI output seems to be using default HKLs. In the GUI, I got these runtime log messages:
```
fit_grains:min_sfac_ratio not specified, defaulting to None
fit_grains:reset_exclusions not specified, defaulting to True
fit_grains:dmin not specified, defaulting to None
fit_grains:dmax not specified, defaulting to None
fit_grains:tthmin not specified, defaulting to None
fit_grains:tthmax not specified, defaulting to None
fit_grains:sfacmin not specified, defaulting to None
fit_grains:sfacmax not specified, defaulting to None
fit_grains:pintmin not specified, defaulting to None
fit_grains:pintmax not specified, defaulting to None
```
In particular, the reset_exclusions defaulted to True when the GUI ran fit-grains. It looks like you also need to update the runtime config file.

donald-e-boyce (approved):
Looks good.

psavery:
@donald-e-boyce ran this with tilts set to 0 and got identical results in the GUI and in the CLI. One source of precision loss is that the GUI stores the tilt values internally according to the user's Euler angle convention setting, and converting between the user's Euler angle convention to exponential map parameters causes a loss of precision. We should potentially explore always storing the tilts as exponential map parameters so we do not lose this precision.

donald-e-boyce:
Here is some documentation: [notes on exclusions](https://github.com/donald-e-boyce/hexrd3/wiki/exclusions).  We may want put this somewhere in the main hexrd repository.

I agree it makes sense to notify the user that things are being igored in the case `reset_exclusions` is False.  I suggest we put a warning [here](https://github.com/HEXRD/hexrd/blob/081b3da9498954a4a804d6dcb4de6543dc5c2bb6/hexrd/config/utils.py#L57-L59), something like: 
```
If not reset_exclusions:
   warning.warn("Caution: exclusions are being ignored. If this was not intentional, you may get surprising results")
```

PR #1454, opened 2023-05-17
Add tooltips for QActions when they are disabled

psavery:
This provides info as to why they are disabled.

disabled_qaction_tooltips

Fixes: #1449


bnmajor (approved):
Very nice!

Issue #1450, opened 2023-05-17
Enlarge cartesian view during interaction if needed


psavery:
If the user is interacting with detector parameters, and a detector is moved out of the current cartesian view image, the view should be enlarged so that they can still see the full detector plate.

PR #1456, opened 2023-05-17
Enlarge cartesian view in interactions if needed

psavery:
If interacting with a detector causes it to move outside of the current limits of the cartesian view, enlarge the cartesian view so that it can still be viewed. This makes interaction much nicer when moving detectors around.

Fixes: #1450


bnmajor (approved):
Awesome!!

Issue #1451, opened 2023-05-17
Polar view breaks if pixel size is modified


psavery:
We need to regenerate the dummy images (and remove any loaded images) if the pixel size is modified so that the polar view can be generated correctly. This would fix this exception:

```python
Traceback (most recent call last):
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 16, in polar_viewer
    return InstrumentViewer()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 25, in __init__
    self.draw_polar()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polar_plot.py", line 71, in draw_polar
    self.pv.warp_all_images()
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polarview.py", line 443, in warp_all_images
    self.create_warp_image(det)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polarview.py", line 224, in create_warp_image
    self.warp_dict[det] = self.warp_image(img, panel)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/polarview.py", line 281, in warp_image
    wimg = panel.interpolate_bilinear(
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/instrument/detector.py", line 936, in interpolate_bilinear
    assert is_2d and right_shape,\
AssertionError: input image must be 2-d with shape (3888, 3000)
```

PR #1457, opened 2023-05-17
Reload dummy images if detector shape changes

psavery:
If the detector shape changes, the images stored internally will not match the new shape. We must reset them. Otherwise, there are errors in the polar view.

Fixes: #1451


bnmajor (approved):
LGTM!

Issue #1453, opened 2023-05-17
Sometimes, navigation bar at the bottom gets duplicated


psavery:
I've found at least one situation where this happens: if you have multiple detectors, and you check "Tabbed View", switch tabs to a different tab at the top, and then uncheck "Tabbed View", duplicate navigation bars appear at the bottom:

![image](https://github.com/HEXRD/hexrdgui/assets/9558430/3f4253fa-fe54-46f4-a3a6-55f2e9862150)

I believe this is caused by the fact that when you go to the tabbed view, there are multiple canvases, and each one has their own navigation bar. When you leave the tabbed view, sometimes these extra navigation bars do not go away.

PR #1459, opened 2023-05-17
Hide all toolbars when clearing so none remain

psavery:
Even though extra toolbars were being deleted, they were not being hidden, and thus a reference to them were held and they remained visible.

Hide all toolbars when clearing so this does not happen.

Fixes: #1453


bnmajor (approved):
Works great for me!

PR #1466, opened 2023-05-18
Use nansum for zoom box subplots

psavery:
This prevents points on the lines from disappearing entirely when there are nans in the row/column.

This PR also adds support for enabling/disabling the entire image mode widget, as well as fixing mouse hovering on the zoom canvas in unsummed mode.


bnmajor (approved):
LGTM!

Issue #1452, opened 2023-05-17
The editor sometimes stays open when editing values in the tree view


psavery:
Sometimes, the editor doesn't close if you edit a value in the tree view. This has side effects, such as making the value uneditable until you switch views. Here is an example. If you start with `hexrdgui --ignore-settings`, and then you switch to the form view and back to the tree view, then edit a value, you encounter the issue:

https://github.com/HEXRD/hexrdgui/assets/9558430/3a676716-659d-448d-86f2-adfa2561692b

We should fix this.

PR #1468, opened 2023-05-19
Ensure tree view gets fully reset when clearing

psavery:
When we cleared the tree view before, we would call `beginRemoveRows()` and `endRemoveRows()`. However, some parts of the model were apparently not notified of the underlying data change, and they were not updating properly. Maybe a `dataChanged()` would have fixed this, but since we are deleting everything anyways, we should just do `beginResetModel()` and `endResetModel()` instead.

This appears to fix the issues.

Fixes: #1452


bnmajor (approved):
Works well for me!

PR #1470, opened 2023-05-19
Add a marker similar to IGOR's for line picking

psavery:
We generated this marker by drawing an SVG in inkscape and then converting it to matplotlib-friendly vertices and connection info.

This is now used by default for the line picker dialog.

example


saransh13 (approved):
LGTM!

Issue #1469, opened 2023-05-19
Line picker zoom box is incorrect near canvas borders


psavery:
You can see from the image here that the mouse on the canvas is not on the detector, but in the zoom box, it is.

![line_picker_border_issue](https://github.com/HEXRD/hexrdgui/assets/9558430/b0da0698-89bd-4a99-9a04-00dfcd9c33f7)

I think there is some issue when some of the zoom box is off of the canvas.

PR #1471, opened 2023-05-19
Fix line picker zoom box on canvas borders

psavery:
When the line picker zoom box was previously near canvas borders, the clipping to the x and y limits that would occur would make the drawn zoom box image incorrect.

Instead of clipping the x and y limits, we should instead clip the lines in the subplots. This fixes the issue.

Fixes: #1469


saransh13 (approved):
Tested with some data. LGTM!

Issue #1472, opened 2023-05-23
TARDIS importer for Image PLATE 3 broken


saransh13:
The image for IMAGE-PLATE-3 after the file is selected is a constant array of just 1. Images for IMAGE-PLATE-2 and IMAGE-PLATE-4 work fine. Some screenshots below. Super high priority.

Problem seems to be in `Merge pull request #1457 from HEXRD/reload-dummy-images-on-detector-shape-change	167caaf`. Since IMAGE-PLATE-3 changes shape while load, the image gets reset to only ones.

![Screen Shot 2023-05-22 at 6 00 09 PM](https://github.com/HEXRD/hexrdgui/assets/15834451/bb31ee35-3edb-4c40-ba5a-81c079070f65)
![Screen Shot 2023-05-22 at 6 01 00 PM](https://github.com/HEXRD/hexrdgui/assets/15834451/a50e836b-5a0b-482d-aa91-7c225db1575b)

PR #1473, opened 2023-05-23
Fix the LLNL import of the rind detector

psavery:
Previous updates to reload dummy images when pixel sizes change introduced an issue when loading the rind via the LLNL import tool.

This update fixes the code so that dummy images are not reloaded for the rind.

Fixes: #1472


saransh13 (approved):
LGTM

PR #1477, opened 2023-05-23
Fix exception that occurs when buffer is None

psavery:
This adds a check for when the buffer is `None`. If it is `None`, then we should just display zeros for the panel buffer border.


saransh13 (approved):
LGTM

PR #1483, opened 2023-05-25
Llnl tool improvements

### LINKED ISSUE #1420: LLNL Import Tool Improvements
- [x] Default to "Load Configuration" with "Default" checked (since that appears to be the most commonly used setting)
- [x] Skip showing the image load dialog that asks about the transform. This was mainly for pairing images with detectors. It is not needed since you are still able to transform the image after loading (via the `Add Transform` button).
- [x] Disable changing the image mode while the LLNL import tool is active. Currently, if you change the image mode, it breaks the application.
- [x] After saving a TARDIS detector boundary, when the current detector changes (and the user is no longer to make any edits to the current template), clear the selected image and the canvas. Alternatively, if we want to allow the user to fix their template after saving it (which may especially happen for image plate 3, where some extra cropping is done), we may need to add a way for them to modify it without loading the image again.
- [x] Consider making right-click rotate while left-click translates. We may need to keep the buttons, however, to determine what to do for key press events.
- [x] Add text in `Edit Detector Outline` specific for each detector that explains why the bounding box options are or are not grayed out.

bnmajor:
Fixes #1420

### DIFF
--- a/hexrd/ui/image_mode_widget.py
+++ b/hexrd/ui/image_mode_widget.py
@@ -99,6 +99,8 @@ class ImageModeWidget(QObject):
 
         HexrdConfig().enable_image_mode_widget.connect(
             self.enable_image_mode_widget)
+        HexrdConfig().set_image_mode_widget_tab.connect(
+            self.set_image_mode_widget_tab)
 
         self.ui.polar_show_snip1d.clicked.connect(self.polar_show_snip1d.emit)
 
@@ -133,7 +135,10 @@ class ImageModeWidget(QObject):
     def enable_image_mode_widget(self, b):
         self.ui.tab_widget.setEnabled(b)
 
-    def currentChanged(self, index):
+    def set_image_mode_widget_tab(self, idx):
+        self.ui.tab_widget.setCurrentIndex(idx)
+
+    def currentChanged(self, index = None):
         modes = {
             0: ViewType.raw,
             1: ViewType.cartesian,
--- a/hexrd/ui/image_mode_widget.py
+++ b/hexrd/ui/image_mode_widget.py
@@ -143,13 +143,13 @@ class ImageModeWidget(QObject):
     def enable_image_mode_widget(self, b):
         self.ui.tab_widget.setEnabled(b)
 
-    def set_image_mode_widget_tab(self, idx):
-        tab = VIEW_MODE_TO_TAB_INDEX[idx]
+    def set_image_mode_widget_tab(self, view_mode):
+        tab = VIEW_MODE_TO_TAB_INDEX[view_mode]
         self.ui.tab_widget.setCurrentIndex(tab)
 
     def currentChanged(self, index):
-        ind = TAB_INDEX_TO_VIEW_MODE[index]
-        self.tab_changed.emit(ind)
+        view_mode = TAB_INDEX_TO_VIEW_MODE[index]
+        self.tab_changed.emit(view_mode)
 
     def all_widgets(self):
         widgets = [
--- a/hexrd/ui/image_mode_widget.py
+++ b/hexrd/ui/image_mode_widget.py
@@ -13,6 +13,14 @@ from hexrd.ui.overlays import Overlay
 from hexrd.ui.ui_loader import UiLoader
 from hexrd.ui.utils import block_signals
 
+TAB_INDEX_TO_VIEW_MODE = {
+    0: ViewType.raw,
+    1: ViewType.cartesian,
+    2: ViewType.polar,
+    3: ViewType.stereo,
+}
+VIEW_MODE_TO_TAB_INDEX = {v: k for k, v in TAB_INDEX_TO_VIEW_MODE.items()}
+
 
 class ImageModeWidget(QObject):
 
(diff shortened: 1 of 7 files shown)


saransh13 (approved):
I tried it out with some TARDIS data and the changes look good to me. I haven't looked too deeply in the actual code, but the features seems to be working well.

hexrd/ui/image_mode_widget.py
```diff
@@ -136,17 +144,12 @@ def enable_image_mode_widget(self, b):
...
-    def currentChanged(self, index = None):
-        modes = {
-            0: ViewType.raw,
-            1: ViewType.cartesian,
-            2: ViewType.polar,
-            3: ViewType.stereo,
-        }
-        ind = self.ui.tab_widget.currentIndex()
-        self.tab_changed.emit(modes[ind])
+        tab = VIEW_MODE_TO_TAB_INDEX[idx]
```
psavery, line 147: The `idx` variable should be renamed to `view_mode`, to emphasize it is actually a view mode, and not an idx.

hexrd/ui/image_mode_widget.py
```diff
@@ -136,17 +144,12 @@ def enable_image_mode_widget(self, b):
...
+        self.ui.tab_widget.setCurrentIndex(tab)
+
+    def currentChanged(self, index):
+        ind = TAB_INDEX_TO_VIEW_MODE[index]
+        self.tab_changed.emit(ind)
```
psavery, line 152: The `ind` here can also be renamed to `view_mode`.

psavery (commented):
A few issues to fix:

- [ ] Switching the instrument to TARDIS and then back to `None` should cause all of the same options to be disabled as before (the configuration and detector combo boxes do not become disabled). I think we should make switching the instrument type call exactly the same logic as pressing the "Cancel LLNL Import" button, except that it won't hide the tool, and the instrument type gets changed to whatever was selected.
- [ ] If you shift-left-click and rotate the template too much, it starts bouncing around. I'm not sure why, but it should just rotate like normal.
- [ ] If you shift-left-click, and then release the shift key while dragging, the template starts translating, but then when you release, it ends up somewhere other than where you left it.  We should make sure the behavior of pressing shfit while left-click is held and releasing shift while left-click is held is defined.
- [ ] When the detector boundary is saved, we should disable `Save detector boundary` and `Transform Image` and the options associated with it (if you modify these after saving the detector boundary, you start getting exceptions)
- [ ] If you select `Currently loaded configuration` and then switch to a different instrument type, your previous instrument configuration settings get lost (the detectors say "default" rather than the detectors you previously had).

hexrd/ui/image_mode_widget.py
bnmajor: Fixed!

hexrd/ui/image_mode_widget.py
bnmajor: Fixed!

bnmajor:
> If you shift-left-click, and then release the shift key while dragging, the template starts translating, but then when you release, it ends up somewhere other than where you left it. We should make sure the behavior of pressing shift while left-click is held and releasing shift while left-click is held is defined.

This is the only item that I am not sure about. Right now the (updated) behavior is to ignore changes to shift while interacting - so if translating and shift is pressed without releasing the mouse it will continue translating (and vice versa). Does this seem okay or do we want different behavior? @psavery @saransh13

bnmajor:
Converting to draft for the moment - there is weird behavior happening now where images are sometimes cleared after save & reload:
- Set only Image Plate 2 -> Save & Reload -> No issues
- Set Image Plate 3 and/or Image Plate 4 alone or with one or more detectors -> Save & Reload -> Images cleared

bnmajor:
> Converting to draft for the moment - there is weird behavior happening now where images are sometimes cleared after save & reload:
> 
> * Set only Image Plate 2 -> Save & Reload -> No issues
> * Set Image Plate 3 and/or Image Plate 4 alone or with one or more detectors -> Save & Reload -> Images cleared

Fixed in [8e49461](https://github.com/HEXRD/hexrdgui/pull/1483/commits/8e49461bd0345b7ea30617cc607708cb40ec1694) - this seems to have been an issue with the order of operations when resetting the UI (and the signals those resets emit).

psavery:
It's looking really nice, @bnmajor! Only two more issues I think need to be fixed:

- [ ] If you are in a different image mode (such as polar view) when you start using the tool, there is an exception at the beginning:
```python
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/image_canvas.py", line 831, in finish_show_polar
    self.azimuthal_line_artist, = axis.plot(*data)
  File "/home/patrick/virtualenvs/hexrd/lib/python3.10/site-packages/matplotlib/axes/_axes.py", line 1688, in plot
    lines = [*self._get_lines(*args, data=data, **kwargs)]
  File "/home/patrick/virtualenvs/hexrd/lib/python3.10/site-packages/matplotlib/axes/_base.py", line 311, in __call__
    yield from self._plot_args(
  File "/home/patrick/virtualenvs/hexrd/lib/python3.10/site-packages/matplotlib/axes/_base.py", line 504, in _plot_args
    raise ValueError(f"x and y must have same first dimension, but "
ValueError: x and y must have same first dimension, but have shapes (1023,) and (1250,)
```

And then when we hover the mouse over the image, we get a different one:
```python
  File "/home/patrick/virtualenvs/hexrd/lib/python3.10/site-packages/matplotlib/cbook/__init__.py", line 304, in process
    func(*args, **kwargs)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/image_tab_widget.py", line 348, in on_motion_notify_event
    dpanel = iviewer.instr.detectors[key]
KeyError: 'default'
```

This only seems to occur if you are in a different image mode when you start.

- [ ] In PXRDIP, when you change detectors, the `Transform Image` and `Save detector boundary` should become enabled, but it looks like those are only set up to become enabled if you load a new image (but you don't load a new image for PXRDIP).

bnmajor:
>  In PXRDIP, when you change detectors, the Transform Image and Save detector boundary should become enabled, but it looks like those are only set up to become enabled if you load a new image (but you don't load a new image for PXRDIP).

Since the detector bounds will be incorrect if the image transform is applied when it shouldn't be (or vice versa) we will only re-enable the "save" option once the first template has been applied.

psavery:
> > In PXRDIP, when you change detectors, the Transform Image and Save detector boundary should become enabled, but it looks like those are only set up to become enabled if you load a new image (but you don't load a new image for PXRDIP).
> 
> Since the detector bounds will be incorrect if the image transform is applied when it shouldn't be (or vice versa) we will only re-enable the "save" option once the first template has been applied.

Awesome, thanks so much @bnmajor! It looks good to me. Ready to merge when CI passes!

psavery (approved):

Issue #1440, opened 2023-05-06
power calibration without constraining to a structure


saransh13:
For some cases in TARDIS and PXRDIP, the structure is not known. having a calibration routine which can just makes points straight is useful for these cases. Add functionality for these cases.

In addition, add ability to specify some constraints between variables as well as bound constraints on each variable.

PR #1484, opened 2023-05-26
Add first pass at structureless calibration

psavery:
This adds a first pass at performing structureless calibration. First,
points are either picked/loaded from a file. These points define
powder lines that should be straight.

Next, a dialog appears where you can edit refinable parameters (including
the min/max), draw the picks, edit the picks, etc.

You can press `Run` to run a calibration, and `Undo run` to undo what you 
did if it did not work out very well.

Still need to do: 

- [x] color picks on different lines differently in the picks editor
- [x] add pinhole distortion options
- [x] possibly disable more things in the main window, including the instrument config
- [x] don't reset table options if engineering constraints are modified
- [x] verify that applying panel buffers from pinhole dimensions is working 

Depends on the `calibration-new` branch in HEXRD/HEXRD.

structureless_picks

image

https://github.com/HEXRD/hexrdgui/assets/9558430/c787068c-56bf-482f-a3e5-cfa0ff715c00

Fixes: #1440


saransh13 (approved):
All the features work for some example cases I looked at. So, LGTM.

Issue #1369, opened 2023-03-27
Mask range issue


akrygier:
If I make a mask and then change the 2theta or eta range such that it intersects the mask, the mask will sometimes extend across the entire displayed azimuthal range.

saransh13:
@akrygier can you post a screenshot with this issue?

akrygier:
Masked polar image with azimuthal range -90:270
![image](https://github.com/HEXRD/hexrdgui/assets/59886900/274eec75-3ce2-4db0-993a-aefba11adb62)

Same image with azimuthal range changed to -270:90
![image](https://github.com/HEXRD/hexrdgui/assets/59886900/a24ea6a3-528e-4a24-a8ba-fb44f1966b85)

This happens when you simply narrow the range sometimes too, which is the real failure mode since users plausibly do that from time to time. (ie, -270:90 is not a typical azimuthal range, but it does demonstrate the issue)

psavery:
Yeah, we need to fix this

psavery:
I have a fix for this in #1486. Basically, we did the following:

1. Add a check to see if the mask in the polar view is being split on top and bottom (this is being done by checking if two consecutive points on the line around the mask have an eta difference greater than 180 degrees, which should never happen).
2. If it is split, then break it up into two separate polygons on the top and the bottom.
3. Add a couple of extra points to the polygons so that they extend to the edge of the eta border. We interpolate in two theta so that these points are going in the right direction.

akrygier:
👍 🙏
- Andy

PR #1486, opened 2023-05-31
Fix polar mask when eta is split

### LINKED ISSUE #1458: Masks not working as expected
Once masks are applied to the cylindrical detector and view is switched, or if state file is saved, then the masks vanish.

psavery:
If a polar mask is split in eta so that part of it is at the top and
part of it is at the bottom, it is now split into two polygons. These
polygons get some extra padding (so they cover all pixels to the eta
border) and some interpolation is done in the two theta pixel to
ensure the padding runs in the correct direction.

This PR also fixes #1458, where masks would disappear if any nan coordinates were present.
    
Fixes: #1369
Fixes: #1458

### DIFF
--- a/hexrd/ui/create_polar_mask.py
+++ b/hexrd/ui/create_polar_mask.py
@@ -10,9 +10,9 @@ from hexrd.ui.utils.conversions import pixels_to_angles
 
 
 def convert_raw_to_polar(det, line):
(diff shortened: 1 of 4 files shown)


bnmajor (approved):
Fantastic!

Issue #1481, opened 2023-05-23
Pinhole corrections are undone when polar view mask is drawn


saransh13:
If `Apply 2\theta distortion` is checked and then another mask is drawn in the polar view, then the polar image gets shifted back to the same state as if the distortions are not applied. If the `Apply 2\theta distortion` button is checked off and on again, then the new mask in the polar view vanishes.

Issue #1448, opened 2023-05-16
mask location when applying sample and pinhole corrections


saransh13:
The mask locations don't move when the pinhole or sample corrections are applied to the polar view. The mask needs to move by the same amount. I assumed that the mask was applied in the raw view, so any distortion applied to the polar view will automatically move correctly in the polar view.

PR #1487, opened 2023-05-31
Apply masks before applying tth distortion

psavery:
We want the tth distortion to also move the masks, so make sure the masks are applied before applying the tth distortion.

This also fixes the `reapply_masks()` function to also apply the tth distortion again (otherwise, it would be lost).

Need to fix:

- [x] If tth distortion is applied, and then you draw a mask in the polar view, it acts as if you drew a mask in the non-distorted polar view (so the pixels get shifted a little bit compared to what you drew). We should fix this so drawing masks in a polar view with tth distortion applied is more intuitive.

Fixes: #1448
Fixes: #1481

### DIFF
--- a/hexrd/ui/calibration/polarview.py
+++ b/hexrd/ui/calibration/polarview.py
@@ -287,7 +287,13 @@ class PolarView:
             data=wimg, mask=nan_mask, fill_value=0.
         )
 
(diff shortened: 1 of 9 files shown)


bnmajor (approved):
Looks great!

PR #1488, opened 2023-05-31
Masking improvements

### LINKED ISSUE #1414: Clarify text for hand-drawn masks
See discussion: https://github.com/HEXRD/hexrdgui/discussions/1397#discussioncomment-5736208

Maybe it should say instead:

```
Add points around a region to mask with left-click. 

Start a new masked region with a single right-click.
```

But feel free to give another suggestion, @bnmajor

### LINKED ISSUE #1465: Finish hand drawn mask with double click

### LINKED ISSUE #1479: button to turn all masks on/off
Can we add a button in the `Mask management` menu to toggle all masks on or off.

bnmajor:
Fixes #1414
Fixes #1479
Fixes #1465

### DIFF
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -132,7 +132,7 @@ class InstrumentViewer:
             if name not in HexrdConfig().visible_masks:
                 continue
 
-                data[f'mask_{name}'] = mask
+            data[f'mask_{name}'] = mask
 
         # Delete the file if it already exists
         if filename.exists():
--- a/hexrd/ui/calibration/polar_plot.py
+++ b/hexrd/ui/calibration/polar_plot.py
@@ -132,7 +132,7 @@ class InstrumentViewer:
             if name not in HexrdConfig().visible_masks:
                 continue
 
-            data[f'mask_{name}'] = mask
+                data[f'mask_{name}'] = mask
 
         # Delete the file if it already exists
         if filename.exists():
--- a/hexrd/ui/hand_drawn_mask_dialog.py
+++ b/hexrd/ui/hand_drawn_mask_dialog.py
@@ -134,10 +134,11 @@ class HandDrawnMaskDialog(QObject):
         self.add_line()
 
     def button_pressed(self, event):
-        if event.dblclick:
-            self.accepted()
-            self.ui.close()
         if event.button == 1:
+            if event.dblclick:
+                self.accepted()
+                self.ui.close()
+                return
             self.drawing = True
         if event.button == 3:
             self.line_finished()
--- a/hexrd/ui/hand_drawn_mask_dialog.py
+++ b/hexrd/ui/hand_drawn_mask_dialog.py
@@ -95,8 +95,10 @@ class HandDrawnMaskDialog(QObject):
 
     def axes_exited(self, event):
         if not self.drawing:
-            self.linebuilder.disconnect()
-            self.lines.pop()
+            if self.linebuilder:
+                self.linebuilder.disconnect()
+            if self.lines:
+                self.lines.pop()
 
     def add_line(self):
         if self.drawing:
--- a/hexrd/ui/hand_drawn_mask_dialog.py
+++ b/hexrd/ui/hand_drawn_mask_dialog.py
@@ -134,6 +134,9 @@ class HandDrawnMaskDialog(QObject):
         self.add_line()
 
     def button_pressed(self, event):
+        if event.dblclick:
+            self.accepted()
+            self.ui.close()
         if event.button == 1:
             self.drawing = True
         if event.button == 3:
--- a/hexrd/ui/mask_manager_dialog.py
+++ b/hexrd/ui/mask_manager_dialog.py
@@ -85,7 +85,8 @@ class MaskManagerDialog(QObject):
         self.ui.import_masks.clicked.connect(self.import_masks)
         self.ui.panel_buffer.clicked.connect(self.masks_to_panel_buffer)
         self.ui.view_masks.clicked.connect(self.show_masks)
-        self.ui.hide_masks.toggled.connect(self.hide_masks)
+        self.ui.hide_all_masks.clicked.connect(self.hide_all_masks)
+        self.ui.show_all_masks.clicked.connect(self.show_all_masks)
 
         HexrdConfig().mode_threshold_mask_changed.connect(
             self.update_masks_list)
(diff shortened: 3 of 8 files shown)


hexrd/ui/calibration/polar_plot.py
```diff
@@ -132,7 +132,7 @@ def write_image(self, filename='polar_image.npz'):
             if name not in HexrdConfig().visible_masks:
                 continue
 
-            data[f'mask_{name}'] = mask
+                data[f'mask_{name}'] = mask
```
psavery, line 135:
```suggestion
            data[f'mask_{name}'] = mask
```

I think this should be switched back to the previous indentation. Otherwise, this is unreachable code.

hexrd/ui/hand_drawn_mask_dialog.py
```diff
@@ -134,6 +134,9 @@ def line_finished(self):
         self.add_line()
 
     def button_pressed(self, event):
+        if event.dblclick:
+            self.accepted()
+            self.ui.close()
```
psavery, line 139: Right now, this `event.dblclick` includes double left-click, double middle-click, and double right-click. Do we want this to be just for double left-click?

Also, we should return after `self.ui.close()`, otherwise, the code below this continues to run (and double right-click produces some errors).

hexrd/ui/mask_manager_dialog.py
```diff
@@ -136,6 +137,22 @@ def toggle_threshold_visibility(self, status):
         if self.threshold_cb is not None:
             self.threshold_cb.setChecked(status)
 
+    def masks_changed(self):
+        if self.image_mode == ViewType.polar or self.image_mode == ViewType.stereo:
```
psavery, line 141:
```suggestion
        if self.image_mode in (ViewType.polar, ViewType.stereo):
```

hexrd/ui/mask_manager_dialog.py
```diff
@@ -136,6 +137,22 @@ def toggle_threshold_visibility(self, status):
         if self.threshold_cb is not None:
             self.threshold_cb.setChecked(status)
 
+    def masks_changed(self):
```
psavery, line 140: Awesome! This is a great example of using a function to reduce duplicated code.

hexrd/ui/mask_manager_dialog.py
```diff
@@ -136,6 +137,22 @@ def toggle_threshold_visibility(self, status):
...
+
+    def update_export_options(self):
+        hidden = self.ui.hide_masks.isChecked()
+        visible = len(HexrdConfig().visible_masks)
+        enabled = (not hidden) and visible
+        self.ui.export_masks.setEnabled(enabled)
+        self.ui.panel_buffer.setEnabled(enabled)
+        tip = '' if enabled else 'Masks hidden. Only visible masks are used.'
+        self.ui.export_masks.setToolTip(tip)
+        self.ui.panel_buffer.setToolTip(tip)
```
psavery, line 154: It looks to me that exporting the masks actually writes out all masks (both visible and invisible), and it also saves the visibilities. I think this is desired behavior.

However, that means that it doesn't make sense to disable the `export_masks` button depending on whether masks are visible or not. I think the only time we'd want to disable the `export_masks` button is if there are no masks at all.

Also, applying the masks to the panel buffer also uses all masks (visible and invisible). I don't know if this is the desired behavior (like, it might make more sense to only apply **visible** masks to the panel buffers), but unless we change that, we also shouldn't disable the panel buffer button based upon the visibilities of the masks. I think this one also should only be disabled if there are no masks at all.

But it is really great to have tooltips when buttons are disabled that explain why they are disabled!

hexrd/ui/mask_manager_dialog.py
```diff
@@ -388,3 +400,12 @@ def show_masks(self):
             axis.imshow(HexrdConfig().raw_masks_dict[det])
         fig.canvas.draw_idle()
         fig.show()
+
+    def hide_masks(self, checked):
+        if checked:
+            HexrdConfig().visible_masks.clear()
+        else:
+            HexrdConfig().visible_masks = list(self.masks.keys())
```
psavery, line 408: It looks to me that `HexrdConfig().visible_masks` is paired with the checkboxes in the table. And since we are trying to make those checkboxes and this checkbox separate, I think the `Hide all masks` checkbox should not modify the same variables.

However, to make the internal code design simpler, I think for now it might be easier to just add two buttons: `Hide all masks` and `Show all masks`. These would modify `HexrdConfig().visible_masks`, and then we would update all checkboxes in the table to reflect their new state.

The only reasons why a separate checkbox would be preferable are:

1. It is only one widget (a checkbox) instead of two buttons (so takes up less space in the UI)
2. It could allow global control of mask visibilities that is separate from individual control (the checkboxes in the table). But I think if we wanted this, we should add a separate variable for the global mask visibilities (similar to `HexrdConfig().show_overlays`), and save that in the state as well.

Instead of the checkbox, I think buttons are fine for now unless someone asks for the checkbox.

psavery (changes requested):
Left a few comments.

Also, can we make sure the dialog has a good default size? It looks like the default width is now too small after these changes:

image

We should make the default width larger so that part of the table isn't hidden.

hexrd/ui/calibration/polar_plot.py
bnmajor: Thanks, fixed.

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: Right now I've updated this to only handle left-click double-click but that can be changed if it should be. Return statement was added as well.

hexrd/ui/mask_manager_dialog.py
bnmajor: For now I have just removed the code that disables the export but I agree, I think it makes sense to only use visible masks. To keep things consistent I'll skip that logic in this PR and I can open another PR with the changes to disable export/add tooltips/only use visible masks.

bnmajor:
> Also, can we make sure the dialog has a good default size? It looks like the default width is now too small after these changes

Can you let me know if you are still seeing this? I am not seeing this in my testing

hexrd/ui/mask_manager_dialog.py
bnmajor: Fixed

hexrd/ui/mask_manager_dialog.py
```diff
@@ -390,10 +391,18 @@ def show_masks(self):
         fig.canvas.draw_idle()
         fig.show()
 
-    def hide_masks(self, checked):
-        if checked:
-            HexrdConfig().visible_masks.clear()
-        else:
-            HexrdConfig().visible_masks = list(self.masks.keys())
+    def toggle_all_masks(self):
```
psavery: This is a great function! But I am changing the name to `update_visibility_checkboxes` instead. The name `toggle_all_masks` makes me thing it is going to go through the checkboxes and do this: `cb.setChecked(not cb.isChecked())` (which is what toggling is).

psavery (approved):
I fixed the default width of the dialog (just needed to go into QtCreator and modify the geometry).

I also pushed a couple more small fixes.

With these changes, it looks good to me!

bnmajor:
> I fixed the default width of the dialog (just needed to go into QtCreator and modify the geometry).

Weird to me that I also bumped this size and was not seeing the issue you had. If this fix worked for you though then glad it's resolved!

PR #1489, opened 2023-05-31
Add the loaded file info to the toolbar

### LINKED ISSUE #1460: Display/record image file name
It would be very useful to have a record of the filenames of the images accessible either in the Instrument panel (id) and/or displayed somewhere else.

Much of the data looks similar and I am frequently forgetting which shot I am analyzing, etc.

Maybe the save state name can default to the same name but with .h5 extension (or similar since it doesn't work for NIF obviously...)

bnmajor:
Adds the list of file names to a tooltip in the toolbar at the bottom of the window. The full path is still available in the menu: `View>Loaded Images`

files_tooltip

Fixes #1460

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -392,13 +392,19 @@ class HexrdConfig(QObject, metaclass=QSingleton):
 
     # Load the state from QSettings
     def _load_state_from_settings(self, settings):
+        # Skip these when loading the QSettings.
+        # These need to be saved in state files, but we do not want them
+        # to persist in between regular sessions.
+        skip = [
+            'azimuthal_overlays',
+            '_recent_images',
+        ]
+
         state = {}
         for name, default in self._attributes_to_persist():
-            if name == 'azimuthal_overlays':
-                # We need to keep this value in the settings so that it is
-                # saved to state files, but we want to forget selections
-                # between sessions.
+            if name in skip:
                 continue
+
             state[name] = settings.value(
                 self._attribute_to_settings_key(name), default)
 
@@ -412,8 +418,6 @@ class HexrdConfig(QObject, metaclass=QSingleton):
         state = {}
         for name, _ in self._attributes_to_persist():
             state[self._attribute_to_settings_key(name)] = getattr(self, name)
-
-        state['euler_angle_convention'] = self.euler_angle_convention
 
         return state
 
@@ -871,7 +875,8 @@ class HexrdConfig(QObject, metaclass=QSingleton):
     def clear_images(self, initial_load=False):
         self.reset_unagg_imgs()
         self.imageseries_dict.clear()
-        self._recent_images.clear()
+        # Set this instead of clearing so the signal gets emitted
+        self.recent_images = {}
         if self.load_panel_state is not None and not initial_load:
             self.load_panel_state.clear()
             self.load_panel_state_reset.emit()
--- a/hexrd/ui/image_series_toolbar.py
+++ b/hexrd/ui/image_series_toolbar.py
@@ -45,7 +45,7 @@ class ImageSeriesToolbar(QWidget):
         self.omega_label.setVisible(False)
 
         data = resource_loader.load_resource(hexrd.ui.resources.icons,
-                                        'file.svg', binary=True)
+                                             'file.svg', binary=True)
         pixmap = QPixmap()
         pixmap.loadFromData(data, 'svg')
         self.file_label = QLabel(self.parent())
@@ -113,7 +113,7 @@ class ImageSeriesToolbar(QWidget):
         ome_range = HexrdConfig().omega_ranges
 
         enable = not is_aggregated and ome_range is not None
-        self.toggle_widget_visibility(enable)
+        self.omega_label.setVisible(enable)
         if not enable:
             return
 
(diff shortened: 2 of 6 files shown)


hexrd/ui/image_series_toolbar.py
```diff
@@ -37,16 +44,29 @@ def create_widget(self):
         self.omega_label = QLabel(self.parent())
         self.omega_label.setVisible(False)
 
+        data = resource_loader.load_resource(hexrd.ui.resources.icons,
+                                        'file.svg', binary=True)
```
psavery, line 48:
```suggestion
        data = resource_loader.load_resource(hexrd.ui.resources.icons,
                                             'file.svg', binary=True)
```

hexrd/ui/image_series_toolbar.py
```diff
@@ -93,9 +113,18 @@ def update_omega_label_text(self):
         ome_range = HexrdConfig().omega_ranges
 
         enable = not is_aggregated and ome_range is not None
-        self.omega_label.setVisible(enable)
+        self.toggle_widget_visibility(enable)
```
psavery, line 116: I think we should change this one back. It is possible for an image stack to be loaded with no omega information associated with it. If that's the case, the omega label should be invisible, but the slider and frame number should be visible.

hexrd/ui/image_series_toolbar.py
```diff
@@ -93,9 +113,18 @@ def update_omega_label_text(self):
...
         self.omega_label.setText(f'  Omega range: [{ome_min}°, {ome_max}°]')
+
+    def update_file_tooltip(self):
+        tips = []
+        for det, images in HexrdConfig().recent_images.items():
+            fnames = [Path(img).name for img in images]
+            tips.append(
+                f'{det}: {(", ").join(fnames)}'
+            )
+        self.file_label.setToolTip(('\n').join(tips))
```
psavery, line 130:
```suggestion
        self.file_label.setToolTip('\n'.join(tips))
```

hexrd/ui/image_series_toolbar.py
```diff
@@ -93,9 +113,18 @@ def update_omega_label_text(self):
...
 
         ome_min, ome_max = ome_range
         self.omega_label.setText(f'  Omega range: [{ome_min}°, {ome_max}°]')
+
+    def update_file_tooltip(self):
+        tips = []
+        for det, images in HexrdConfig().recent_images.items():
+            fnames = [Path(img).name for img in images]
+            tips.append(
+                f'{det}: {(", ").join(fnames)}'
```
psavery, line 128:
```suggestion
                f'{det}: {", ".join(fnames)}'
```

psavery (changes requested):

psavery (commented):
@bnmajor Things to fix:

- [ ] Can we try to make the icon the same size as the other icons in the toolbar?
- [ ] Can we move the icon to be to the left of the matplotlib navigation toolbar icons, instead of to the right?
- [ ] We should eliminate the menu item that displays the file names as well.
- [ ] The image series toolbar doesn't keep sliding (see #1490). It appears this is because `set_visible()` is for some reason being called while the slider is moving, and there is a `set_visible(False)` followed by a `set_visible(True)`. Before, this didn't cause an issue because the `QWidget` parent had its visibility modified. But now that the QSlider has its visibility modified, it is turning off the interaction.

bnmajor:
> * [x]  Can we move the icon to be to the left of the matplotlib navigation toolbar icons, instead of to the right?

This is fixed in [e81e540](https://github.com/HEXRD/hexrdgui/pull/1489/commits/e81e540142b5bfbe73b5afcf4922d44f6c29d54b) by creating a third toolbar that is now positioned before the matplotlib toolbar. This also fixes the slider bug because the `ImageSeriesToolbar` behavior has been reverted and we are no longer toggling the visibility of individual widgets.

bnmajor:
> Can we try to make the icon the same size as the other icons in the toolbar?

I tried to match up the size a bit better but I think part of the reason they look so different is because of the button (in the nav bar) vs label (our tooltip) formatting. Happy to tweak it some more if that seems helpful.
toolbars

hexrd/ui/hexrd_config.py
```diff
@@ -413,8 +419,6 @@ def state_to_persist(self):
         for name, _ in self._attributes_to_persist():
             state[self._attribute_to_settings_key(name)] = getattr(self, name)
 
-        state['euler_angle_convention'] = self.euler_angle_convention
```
psavery, line 416: I am deleting this because it seems repetitious. It should already be in the state.

hexrd/ui/image_series_toolbar.py
```diff
@@ -8,6 +8,54 @@
 from hexrd.ui.hexrd_config import HexrdConfig
 
 
+class ImageSeriesInfoToolbar(QWidget):
```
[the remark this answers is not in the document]
psavery: Looks fine to me for now!

psavery (approved):
LGTM! I pushed a couple of minor changes as well (including making the icon a little bigger so it better matches with the other icons).

bnmajor:
> LGTM! I pushed a couple of minor changes as well (including making the icon a little bigger so it better matches with the other icons).

And here I made it smaller because I thought it looked bigger than the others... I do not have a good eye for these things :upside_down_face:

PR #1491, opened 2023-06-01
Add material name to HKL info for mouse hover

psavery:
The currently selected material in the materials panel is the one that is used for mouse hover info. However, this may not be explicit to users, who may expect whatever overlay is displayed to be the material that is used.

The material name makes it more explicit which material is being used to compute the HKL.

image


bnmajor (approved):
Nice!

PR #1495, opened 2023-06-08
Pxrdip bugs

bnmajor:
- Fix issue seen when changing image rotation using quick load dialog (found while debugging)
- Fix bug that was resetting the config when the LLNL workflow was completed (seen when trying to view PXRDIP in polar view after completing because detector shape was no longer correct).


psavery (approved):
Looks great, thanks @bnmajor!

Issue #1493, opened 2023-06-08
Masking issue near borders


akrygier:
I frequently (always) want to mask the full azimuthal range at low 2 theta; this is usually high background on TARDIS/PXRDIP.

When I try to apply a rectangular or "by hand" mask, there is often a few pixels on the edge that are difficult to select without zooming in significantly. When I try to draw a rectangle that extends beyond the range of the displayed region the mask fails (I guess due to extending beyond a valid range of 2 theta or azimuthal angles).

I tried just changing the displayed 2theta range, but the azimuthal integration (the lineout at the bottom) scale appears to be set by the full range.

psavery:
@akrygier About the "by hand mask", can you update to the latest and try again? I believe I fixed at least part of this issue recently (you can release your mouse outside of the canvas window and it keeps the drawn mask - it doesn't discard it).

And yes, adjusting the two theta range will adjust both the main polar canvas and the azimuthal lineout.

If it is helpful, we could potentially add a menu item where you can specify a two theta range where a mask will get created over the whole azimuthal range. What do you think?

akrygier:
I tried updating using the command at the documentation site and there is an error.

The error happens both when I update and when I clean install into a new conda environment.

Here is the error:
[HEXRD_error.txt](https://github.com/HEXRD/hexrdgui/files/11692059/HEXRD_error.txt)

That would definitely work, but I also think a rectangle mask is sufficient if it is possible to click and drag outside the 2theta/azimuthal range.

akrygier:
everything installs without issue; the error comes up when I try to run hexrdgui

psavery:
Yeah, @akrygier Will take a look!

PR #1496, opened 2023-06-08
Snap rectangle mask to borders if it is close

psavery:
If you draw a rectangle mask and you exit the axes close to the canvas borders (with 1%), snap the mask to the borders. This makes masking pixels at the borders easier.

Fixes: #1493


bnmajor (approved):
Very nice!

PR #1502, opened 2023-06-09
Don't reset unaggregated images until "OK" clicked

psavery:
I believe I ran into an issue where, when I was switching between the aggregations, something got reset in the config, and I believe this was the issue.

We should make sure we don't reset the aggregations on HexrdConfig() unless the "OK" button is clicked.

Does this look good to you, @bnmajor?


bnmajor (approved):
LGTM!

PR #1503, opened 2023-06-09
Ensure update of selected material in grains table

psavery:
These were sometimes getting out-of-sync and this was causing `pull_spots()` to run with the wrong material when visualizing the spots.

This issue is now fixed by adding this extra synchronization.


bnmajor (approved):
Awesome!

Issue #1499, opened 2023-06-09
Masks made in polar view are applied flipped up-down in stereograph


akrygier:
Here are screen shots. The clear evidence is the straight-through beam (bottom part of northern panel); the corresponding mask is on the bottom panel.

![image](https://github.com/HEXRD/hexrdgui/assets/59886900/681eb457-3847-4cde-9a0c-b9734fedd5a6)

![image](https://github.com/HEXRD/hexrdgui/assets/59886900/84aab79f-9021-4a3f-88c5-45b62ac6da4e)

psavery:
@akrygier I think we have found the issue and will have a fix soon

Issue #1333, opened 2023-02-04
Stereo view has inverted y compared to Cartesian view


psavery:
This isn't necessarily an issue unless it makes the stereo view more intuitive to view.

To fix this, we would need to modify the code used to generate the stereo image, and then modify the code where we convert from polar angles to stereo pixels and back (this conversion is used for many things, such as overlays, detector borders, pick points, mouse coordinates, etc.).

psavery:
This is being fixed in #1504 by simply changing the origin of the matplotlib axis to be in the lower left corner instead of in the upper left corner.

PR #1504, opened 2023-06-12
Fix a few issues with the stereographic view

psavery:
The following issues were fixed:

1. Masks are now reapplied correctly when toggling their visibility in the mask manager
2. The masks are now correctly converted to the VISAR view positions

Fixes: #1333
Fixes: #1499


saransh13 (approved):
LGTM

Issue #1497, opened 2023-06-09
Change 2th range during composite calibration point picking


akrygier:
Just had an issue pop up while picking points for a calibration in the "Composite (Laue and Powder)" tool.

While picking points, I changed the viewable 2th range in the "Image mode" window. (I was accidentally zoomed in).

When I increased the range, several things broke. First, in the main window (polar view), the box showing the ROI disappeared. Second, the ROI that is displayed in the point picking tool also disappeared. Then, when I clicked ok the calibration ran, but the main image was completely white (again in polar view).

psavery:
Yeah, we should disable the navigation toolbar and the image mode during point picking, unless we decide to support them.

akrygier:
I think what I was doing is not that common so disabling makes sense.

It might be useful to warn that you are zoomed in.

PR #1506, opened 2023-06-12
Disable most of the app during hand picking

psavery:
If the user modifies certain settings while the hand picking is
being applied, the GUI can get into an unusable state.
    
Disable most of the app during hand picking, except what is relevant
and needed.
    
Fixes: #1497


bnmajor (approved):
Works well, very nice!

PR #1507, opened 2023-06-13
Make several fixes to polar masking

psavery:
This adds several fixes, including:

1. Slightly adjusting the "snap" values in the rectangular mask (so that they don't go out of bounds on conversion)
2. Increasing the required gap to assume the mask is split across eta
3. Adding logic to handle a single gap (there must be two gaps, so we then look for the second largest gap)


bnmajor (approved):
Works great for me!

PR #1509, opened 2023-06-13
Move structureless params yaml file to resources

psavery:
We are currently packaging all of the files in the `resources` directory,
but we are not packaging non-python files outside of that directory. That
means that this structureless params file was being excluded.
    
We could add logic to include it, but to keep things simple, let's just
move it into the resources directory, so it is packaged with the rest
of the program.

Also add help buttons for structureless calibration


bnmajor (approved):
Awesome!

PR #1510, opened 2023-06-13
Change structureless params to plain text

psavery:
These were not being displayed correctly in Windows. We should be able to figure out a way to make this work, but I haven't figured out a way yet.

For now, use plain text characters so that the release will not be blocked.


bnmajor (approved):
LGTM

PR #1519, opened 2023-06-14
Import stack improvements

bnmajor:


hexrd/ui/image_stack_dialog.py
```diff
@@ -194,6 +209,9 @@ def select_omega_file(self):
             HexrdConfig().images_dir, 'NPY files (*.npy)')
         self.ui.omega_file.setText(omega_file)
         self.state['omega'] = omega_file
+        wedges = np.load(omega_file, allow_pickle=False)
```
psavery, line 212: `allow_pickle` is `False` by default, so I don't think we need to specify it.

psavery (approved):
LGTM

Issue #1512, opened 2023-06-13
Improvements for Windows MSI


psavery:
We can make several improvements for the MSI file:

- Add a logo in the places where logos are displayed
- HEXRD appears in the Windows start menu after you install it, but it also should have a logo, and should be renamed to HEXRDGUI
- Add the option for a desktop shortcut

The fact that HEXRD appears in the Windows start menu is also kind of hidden. So it might not be clear to users right away how to run HEXRDGUI after the installation finishes.

PR #1523, opened 2023-06-27
Add hexrd logo and desktop shortcut to MSI

psavery:
Fixes: #1512


saransh13 (approved):
LGTM

psavery:
Here is a summary of how this changes the MSI installer. First of all, note that there is an MSI icon that cannot be changed, so this remains the same:

msi_icon

However, we can change images in most of the other places.

Here was the original background image in the HEXRDGUI setup wizard:

msi_first_page

I have changed this using the `CPACK_WIX_UI_DIALOG` variable to be the HEXRD logo. However, it appears to take up the entire background, and you cannot see the text anymore due to the text remaining black:

new_logo

We might want to construct an image where it is mostly white with the HEXRD icon on the left.

Secondly, the `CPACK_WIX_UI_BANNER` controls the banner at the top. Previously, it was this:

install_page

Now, it appears as this:

banner

Note that once again, it takes up all available space, and we should probably add some white space to the left so that we can see the text.

Good news so far, though, is that the desktop icon appears, although we need to assign it an image (I'm working on that now). The old start menu also did not have an image, as shown:

win_old_start_menu

I am working on fixing that too (hopefully my last commit fixed this).

psavery:
I resized the images and fixed the icons for the desktop icon and the start menu.

Here are the resized images in the MSI installer:

msi_current

msi_install

We can, of course, modify these as needed.

And here are the icons on the desktop and in the start menu:

hexrdgui_icon

start_menu

PR #1524, opened 2023-06-27
Show the detector in the mouse hover info

psavery:
Display the detector that the mouse is hovering over in the mouse hover info.


saransh13 (approved):
Tested for TARDIS data. LGTM!

Issue #1515, opened 2023-06-13
Add graceful exit for HEDM workflow


psavery:
If you start something, such as fit-grains, and you realize that you made a mistake in the options and the run will take forever, there should be a way to gracefully exit.

Maybe create a backup of the state right before running so that if the program exits/crashes, it can be restored.

Or maybe add a `Cancel` button to the progress bar.

PR #1530, opened 2023-07-08
Add the option to cancel fit grains

psavery:
Sometimes, users realize after they start fit-grains that the options selected will take a *very* long time.

Rather than having to kill the application and start the HEDM workflow all over again, we now provide an option to cancel fit-grains. This stops running fit-grains and it goes back to the fit-grains options dialog.

example1

Depends on hexrd/hexrd#543
Fixes part of #1515


psavery:
This was approved verbally by @donald-e-boyce

PR #1538, opened 2023-07-11
Ensure zoom canvas is deleted on cleanup

psavery:
Otherwise, some of the Qt connections do not get disconnected.


bnmajor (approved):
LGTM

PR #1539, opened 2023-07-12
Use non-VISAR instrument for stereo distortion

psavery:
In the stereo view, we were previously using the VISAR instrument to perform pinhole distortion, but this was incorrect and would produce errors.

Now, we are using the non-VISAR instrument. This works correctly.


saransh13 (approved):
The changes looks good. One question: are the overlays using the correct distortion field?

psavery:
> The changes looks good. One question: are the overlays using the correct distortion field?

Yes they are!

PR #1543, opened 2023-07-18
Support hand-drawn masking on tabbed canvases

### LINKED ISSUE #1542: Polygon masking doesn't work in the raw tabbed image view for tabs other than the first

bnmajor:
Fixes #1542

### DIFF
--- a/hexrd/ui/hand_drawn_mask_dialog.py
+++ b/hexrd/ui/hand_drawn_mask_dialog.py
@@ -6,8 +6,8 @@ import numpy as np
 
 import matplotlib.pyplot as plt
 from matplotlib.widgets import Cursor
+from hexrd.ui.hexrd_config import HexrdConfig
 
-from hexrd.ui.constants import ViewType
 from hexrd.ui.ui_loader import UiLoader
 from hexrd.ui.utils import add_sample_points
 
@@ -17,7 +17,7 @@ class HandDrawnMaskDialog(QObject):
     # Emits the ring data that was selected
     finished = Signal(list, list)
 
-    def __init__(self, canvas, parent):
+    def __init__(self, parent):
         super().__init__(parent)
 
         loader = UiLoader()
@@ -25,12 +25,13 @@ class HandDrawnMaskDialog(QObject):
         flags = self.ui.windowFlags()
         self.ui.setWindowFlags(flags | Qt.Tool)
 
-        self.canvas = canvas
         self.ring_data = []
         self.linebuilder = None
         self.lines = []
         self.drawing = False
         self.dets = []
+        self.canvas_ids = []
+        self.det = None
 
         prop_cycle = plt.rcParams['axes.prop_cycle']
         self.color_cycler = cycle(prop_cycle.by_key()['color'])
@@ -42,12 +43,21 @@ class HandDrawnMaskDialog(QObject):
     def setup_connections(self):
         self.ui.accepted.connect(self.accepted)
         self.ui.rejected.connect(self.rejected)
-        self.bp_id = self.canvas.mpl_connect('button_press_event',
-                                             self.button_pressed)
-        self.enter_id = self.canvas.mpl_connect('axes_enter_event',
-                                                self.axes_entered)
-        self.exit_id = self.canvas.mpl_connect('axes_leave_event',
-                                               self.axes_exited)
+        for canvas in self.parent().image_tab_widget.active_canvases:
+            press = canvas.mpl_connect('button_press_event',
+                                                self.button_pressed)
+            enter = canvas.mpl_connect('axes_enter_event',
+                                                    self.axes_entered)
+            exit = canvas.mpl_connect('axes_leave_event',
+                                                self.axes_exited)
+            self.canvas_ids.append([press, enter, exit])
+        self.parent().image_tab_widget.tabBarClicked.connect(
+            self.tabbed_view_changed)
+        HexrdConfig().tab_images_changed.connect(self.tabbed_mode_changed)
+
+    @property
+    def canvas(self):
+        return self.parent().image_tab_widget.active_canvas
 
     def move_dialog_to_left(self):
         # This moves the dialog to the left border of the parent
@@ -70,13 +80,13 @@ class HandDrawnMaskDialog(QObject):
         self.linebuilder = None
         self.cursor = None
 
-        self.canvas.mpl_disconnect(self.bp_id)
-        self.canvas.mpl_disconnect(self.enter_id)
-        self.canvas.mpl_disconnect(self.exit_id)
+        for idx, (press, enter, exit) in enumerate(self.canvas_ids):
+            canvas = self.parent().image_tab_widget.active_canvases[idx]
+            canvas.mpl_disconnect(press)
+            canvas.mpl_disconnect(enter)
+            canvas.mpl_disconnect(exit)
 
-        self.bp_id = None
-        self.enter_id = None
-        self.exit_id = None
+        self.canvas_ids.clear()
         self.canvas.draw()
 
     def start(self):
@@ -89,6 +99,7 @@ class HandDrawnMaskDialog(QObject):
             return
 
         self.ax = event.inaxes
+        self.det = self.ax.get_title()
         # fire up the cursor for this tool
         self.cursor = Cursor(self.ax, useblit=True, color='red', linewidth=1)
         self.add_line()
@@ -131,7 +142,7 @@ class HandDrawnMaskDialog(QObject):
         ring_data = add_sample_points(ring_data, 300)
 
         self.ring_data.append(ring_data)
-        self.dets.append(self.ax.get_title())
+        self.dets.append(self.det)
         self.drawing = False
         self.add_line()
 
@@ -156,6 +167,13 @@ class HandDrawnMaskDialog(QObject):
 
     def show(self):
         self.ui.show()
+
+    def tabbed_view_changed(self):
+        if self.linebuilder:
+            self.line_finished()
+
+    def tabbed_mode_changed(self):
+        self.accepted()
 
 
 class LineBuilder(QObject):
--- a/hexrd/ui/hand_drawn_mask_dialog.py
+++ b/hexrd/ui/hand_drawn_mask_dialog.py
@@ -99,6 +99,10 @@ class HandDrawnMaskDialog(QObject):
             return
 
         self.ax = event.inaxes
+        # Get the detector name on mask creation since completion may have been
+        # triggered by the canvas being changed. No title is returned unless
+        # we're in the raw view but that is the only time it is needed. See:
+        # https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/main_window.py#L727-L745
         self.det = self.ax.get_title()
         # fire up the cursor for this tool
         self.cursor = Cursor(self.ax, useblit=True, color='red', linewidth=1)
--- a/hexrd/ui/hand_drawn_mask_dialog.py
+++ b/hexrd/ui/hand_drawn_mask_dialog.py
@@ -11,13 +11,14 @@ from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.ui_loader import UiLoader
 from hexrd.ui.utils import add_sample_points
 
+# TODO: How to handle image mode? Mode has changed byt the time the signal has been emitted.
 
 class HandDrawnMaskDialog(QObject):
 
     # Emits the ring data that was selected
     finished = Signal(list, list)
 
-    def __init__(self, parent):
+    def __init__(self, canvas, parent):
         super().__init__(parent)
 
         loader = UiLoader()
@@ -25,12 +26,12 @@ class HandDrawnMaskDialog(QObject):
         flags = self.ui.windowFlags()
         self.ui.setWindowFlags(flags | Qt.Tool)
 
+        self.canvas = canvas
         self.ring_data = []
         self.linebuilder = None
         self.lines = []
         self.drawing = False
         self.dets = []
-        self.canvas_ids = []
         self.det = None
 
         prop_cycle = plt.rcParams['axes.prop_cycle']
@@ -39,25 +40,19 @@ class HandDrawnMaskDialog(QObject):
         self.move_dialog_to_left()
 
         self.setup_connections()
+        self.setup_canvas_connections()
 
     def setup_connections(self):
         self.ui.accepted.connect(self.accepted)
         self.ui.rejected.connect(self.rejected)
-        for canvas in self.parent().image_tab_widget.active_canvases:
-            press = canvas.mpl_connect('button_press_event',
-                                                self.button_pressed)
-            enter = canvas.mpl_connect('axes_enter_event',
-                                                    self.axes_entered)
-            exit = canvas.mpl_connect('axes_leave_event',
-                                                self.axes_exited)
-            self.canvas_ids.append([press, enter, exit])
-        self.parent().image_tab_widget.tabBarClicked.connect(
-            self.tabbed_view_changed)
-        HexrdConfig().tab_images_changed.connect(self.tabbed_mode_changed)
 
-    @property
-    def canvas(self):
-        return self.parent().image_tab_widget.active_canvas
+    def setup_canvas_connections(self):
+        self.bp_id = self.canvas.mpl_connect('button_press_event',
+                                            self.button_pressed)
+        self.enter_id = self.canvas.mpl_connect('axes_enter_event',
+                                                self.axes_entered)
+        self.exit_id = self.canvas.mpl_connect('axes_leave_event',
+                                            self.axes_exited)
 
     def move_dialog_to_left(self):
         # This moves the dialog to the left border of the parent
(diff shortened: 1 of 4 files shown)


hexrd/ui/main_window.py
```diff
@@ -717,9 +717,7 @@ def on_action_edit_euler_angle_convention(self):
 
     def on_action_edit_apply_hand_drawn_mask_triggered(self):
         # Make the dialog
-        canvas = self.ui.image_tab_widget.image_canvases[0]
-        self._apply_drawn_mask_line_picker = (
-            HandDrawnMaskDialog(canvas, self.ui))
+        self._apply_drawn_mask_line_picker = (HandDrawnMaskDialog(self.ui))
```
psavery, line 720:
```suggestion
        self._apply_drawn_mask_line_picker = HandDrawnMaskDialog(self.ui)
```

hexrd/ui/hand_drawn_mask_dialog.py
```diff
@@ -42,12 +43,21 @@ def __init__(self, canvas, parent):
     def setup_connections(self):
         self.ui.accepted.connect(self.accepted)
         self.ui.rejected.connect(self.rejected)
-        self.bp_id = self.canvas.mpl_connect('button_press_event',
-                                             self.button_pressed)
-        self.enter_id = self.canvas.mpl_connect('axes_enter_event',
-                                                self.axes_entered)
-        self.exit_id = self.canvas.mpl_connect('axes_leave_event',
-                                               self.axes_exited)
+        for canvas in self.parent().image_tab_widget.active_canvases:
```
psavery, line 46: I think that expecting the parent to be a particular object can cause trouble if we move the dialogs around. It might be better to just pass the active canvases to this class' `__init__` method directly (similar to what we were doing before).

hexrd/ui/hand_drawn_mask_dialog.py
```diff
@@ -42,12 +43,21 @@ def __init__(self, canvas, parent):
...
+            press = canvas.mpl_connect('button_press_event',
+                                                self.button_pressed)
+            enter = canvas.mpl_connect('axes_enter_event',
+                                                    self.axes_entered)
+            exit = canvas.mpl_connect('axes_leave_event',
+                                                self.axes_exited)
+            self.canvas_ids.append([press, enter, exit])
+        self.parent().image_tab_widget.tabBarClicked.connect(
```
psavery, line 54: Same here. I think it's better for this connection to be made in the parent object instead. That can be done in the `on_action_edit_apply_hand_drawn_mask_triggered` function.

hexrd/ui/hand_drawn_mask_dialog.py
```diff
@@ -157,6 +168,13 @@ def rejected(self):
     def show(self):
         self.ui.show()
 
+    def tabbed_view_changed(self):
+        if self.linebuilder:
+            self.line_finished()
+
+    def tabbed_mode_changed(self):
+        self.accepted()
```
psavery, line 176:
```suggestion
        self.accept()
```

hexrd/ui/hand_drawn_mask_dialog.py
```diff
@@ -89,6 +99,7 @@ def axes_entered(self, event):
             return
 
         self.ax = event.inaxes
+        self.det = self.ax.get_title()
```
psavery, line 102: The hand drawn mask works also in the polar view, right? What does this do in the polar view?

hexrd/ui/hand_drawn_mask_dialog.py
psavery: You can also change the `canvas` there too on this object when the tabbed view changes.

psavery (commented):
Looking good so far! A few requested changes.

If I draw a mask in the polar view using this class, and then I switch to the raw view and toggle  `Tabbed View`, I get this error:
```python
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/hand_drawn_mask_dialog.py", line 176, in tabbed_mode_changed
    self.accepted()
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/hand_drawn_mask_dialog.py", line 161, in accepted
    self.line_finished()
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/hand_drawn_mask_dialog.py", line 132, in line_finished
    ring_data = np.vstack([linebuilder.xs, linebuilder.ys]).T
AttributeError: 'NoneType' object has no attribute 'xs'
```

You might need to disconnect a signal somewhere, or make sure the object gets destroyed.

Also, if I am in a tabbed view, on a different canvas than the first, and I start drawing a mask, then uncheck "Tabbed View" without accepting it, I get this error:

```python
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/hand_drawn_mask_dialog.py", line 176, in tabbed_mode_changed
    self.accepted()
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/hand_drawn_mask_dialog.py", line 163, in accepted
    self.clear()
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/hand_drawn_mask_dialog.py", line 84, in clear
    canvas = self.parent().image_tab_widget.active_canvases[idx]
IndexError: list index out of range
```

I think it's great that you took into account what should happen if the tabbed view is toggled during mask drawing or if tabs are changed during mask drawing! But we need to fix a couple of things.

hexrd/ui/main_window.py
bnmajor: Fixed.

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: The list of active canvases will change every time that we toggle the tabbed mode on/off or change the image mode. We can pass the list into the init function like we were before but it seems to me that we will still need a signal that emits the list of active canvases on image mode and tabbed mode change, unless we want to close the dialog and force the user to re-open it each time these are changed?

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: Connection moved.

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: I don't understand this suggestion? We want to call `accepted` to complete the current line and there is no `accept` function.

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: This is just moving the existing logic that was [here](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/hand_drawn_mask_dialog.py#L134) so that the information is grabbed when the masked is created rather than when it is completed (as this may have happened after the mode has changed). For the polar view no title is returned but since it is not used unless we're in the raw view (see [signal](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/main_window.py#L724-L725) and [function](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/main_window.py#L727-L745)) this does not matter.

hexrd/ui/hand_drawn_mask_dialog.py
psavery: Oh I see, sorry. I was thinking about `QDialog`s where `accepted()` is a signal but `accept()` is a slot, and we would typically prefer calling the slot.

No need to change anything here.

hexrd/ui/hand_drawn_mask_dialog.py
psavery: Sounds good! But I think you need to push your changes? I don't see them here.

hexrd/ui/hand_drawn_mask_dialog.py
psavery: I see, that's good to know! Can you leave a short comment here about that?

hexrd/ui/hand_drawn_mask_dialog.py
psavery: So, this `setup_connections()` function is only called in the `__init__` method, so it looks like the code is currently looping over the active canvases at the time of `__init__`, and it is not currently updating the list of canvases if tabbed mode is toggled on/off or the image mode is changed.

Looking at the `ImageTabWidget` code, it looks like the first canvas is never deleted (it is used for all views and for the first tab in the tabbed raw view), but the other canvases (which are only used for non-first tabs in the tabbed raw view) are re-created when `clear()` is called (which happens several times).

I think in the main window function that creates this object, we can connect to any signals like `HexrdConfig().tab_images_changed` that result in the canvases changing, and then we can update the active canvases on this object from there.

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: Added :+1:

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: Pushed in the latest WIP (questions in comment)

hexrd/ui/hand_drawn_mask_dialog.py
bnmajor: Sounds good, I think we're talking about the same approach. I am having some issues though so I've broken them out into their own comment for help/clarification.

bnmajor:
> Looking at the ImageTabWidget code, it looks like the first canvas is never deleted (it is used for all views and for the first tab in the tabbed raw view), but the other canvases (which are only used for non-first tabs in the tabbed raw view) are re-created when clear() is called (which happens several times).
>
> I think in the main window function that creates this object, we can connect to any signals like HexrdConfig().tab_images_changed that result in the canvases changing, and then we can update the active canvases on this object from there.

In the latest WIP commit that I have pushed you can see what I tried to set up, but with no success. The list of active canvases is updated correctly but there are issues with the signal connections that I don't understand. For example:

1. I toggle tabbed mode on and open the polygon masking tool. I am able to draw masks on any of the tabs. The ids produced by setting up the connections are `[[12, 13, 14], [12, 13, 14], [12, 13, 14]]`.
2. If I toggle the tabbed mode off I am able to mask any of the detectors on the single canvas. The ids produced by setting up the connections are `[[18, 19, 20]]`.
3. If I toggle the tabbed mode back on I am able to mask the first canvas but not the last two.The ids produced by setting up the connections are `[[21, 22, 23], [12, 13, 14], [12, 13, 14]]`.

It's not clear to me why when all the canvases signals are set correctly they all have the same id (I had assumed these would be unique ids returned) or why the same results are not produced when we toggle from tab mode off to on?

hexrd/ui/main_window.py
```diff
@@ -718,6 +718,18 @@ def on_action_edit_euler_angle_convention(self):
     def on_action_edit_apply_hand_drawn_mask_triggered(self):
         # Make the dialog
         self._apply_drawn_mask_line_picker = (HandDrawnMaskDialog(self.ui))
+        self.ui.image_tab_widget.tabBarClicked.connect(
+            self._apply_drawn_mask_line_picker.tabbed_view_changed)
+        HexrdConfig().tab_images_changed.connect(
+            lambda: self._apply_drawn_mask_line_picker.tabbed_mode_changed(
+                self.ui.image_tab_widget.active_canvases
+            )
+        )
```
psavery, line 727: I think I have an idea that will simplify the design.

What if we let the `HandDrawnMaskDialog` only hold on to a reference to a single canvas: the one that is currently active? It can then have a `change_canvas()` function that switches the canvas to a different one (it will call `mpl_disconnect()` for the old canvas and `mpl_connect()` for the new canvas).

Then, we just need to keep track somewhere in the code when the active canvas changes. Maybe the best place to do that is in the image tab widget? It can emit a signal anytime the active canvas changes. Then this main window function can call `change_canvas()` on the `HandDrawnMaskDialog`. We could also probably use this signal from the image tab widget that the active canvas changed in other places in the code too.

Here is my current understanding about the active canvas: 

1. If the user does not change to the tabbed view in raw mode, but switches between view modes (cartesian, polar, etc.), then the active canvas does not change. 
2. But if the user changes tabs in the tabbed raw view, then the active canvas changes. And if they then switch tabs or from a non-first tab to a different view mode, the active canvas will change again.
3. It looks like the non-first tab canvases get deleted and remade for every rerender (maybe we should redesign it so that isn't the case). But right now, if the user is on a non-first tab canvas, and a rerender happens, then the active canvas changes (because the old one was deleted and a new one was made).

We would need to take into account that information when emitting the signal when the active canvas changes. But I think it will be easier for the image tab widget to keep track of which canvas is active, and emit a signal when the active canvas changes.

bnmajor: So I did actually try this approach (only tracking the active canvas since that is all we need) but I was encountering the same issues as with the list of active canvases. Your breakdown pointed me at the real culprit though:

> It looks like the non-first tab canvases get deleted and remade for every rerender (maybe we should redesign it so that isn't the case). But right now, if the user is on a non-first tab canvas, and a rerender happens, then the active canvas changes (because the old one was deleted and a new one was made).

I did not catch this initially! Basically what was happening was the connections were made to the new canvases, but then any existing in-progress masks would be completed (the raw masks would be updated) which would trigger a re-render and new canvases and the connections that were created would be useless. So thanks for catching this!

This has been cleaned up now and we now have an `active_canvas_changed` function in the main window that can be used for anything else that may need to be updated when the active canvas changes.

hexrd/ui/main_window.py
```diff
@@ -715,9 +718,17 @@ def on_action_edit_euler_angle_convention(self):
         self.update_all()
         self.update_config_gui()
 
+    def active_canvas_changed(self):
+        if hasattr(self, '_apply_drawn_mask_line_picker'):
+            self._apply_drawn_mask_line_picker.canvas_changed(
+                self.ui.image_tab_widget.active_canvas
+            )
```
psavery, line 725: Let's keep the `active_canvas_changed` function, but can we put this code in a different function named `update_drawn_mask_line_picker_canvas()`? Then, `active_canvas_changed()` will call this function.

Then, we should set up the code here so that `active_canvas_changed()` is only actually called when the active canvas changes (there are a few other situations where it is being called right now where the active canvas is not being changed).

hexrd/ui/main_window.py
```diff
@@ -294,6 +296,7 @@ def setup_connections(self):
         HexrdConfig().raw_masks_changed.connect(self.update_all)
         HexrdConfig().enable_canvas_toolbar.connect(
             self.on_enable_canvas_toolbar)
+        HexrdConfig().tab_images_changed.connect(self.active_canvas_changed)
```
psavery, line 299: This will be connected to `update_drawn_mask_line_picker_canvas()`, since the tab images changing doesn't necessarily mean that the active canvas has changed.

hexrd/ui/main_window.py
```diff
@@ -901,6 +912,10 @@ def on_show_raw_zoom_dialog(self):
         dialog.zoom_height = int(img.shape[0] / 5)
 
     def change_image_mode(self, mode):
+        # The active canvas change needs to be triggered *before* the image
+        # mode is changed. This makes sure that in-progress masks are completed
+        # and associated with the correct image mode.
+        self.active_canvas_changed()
```
psavery, line 918: This will call `update_drawn_mask_line_picker_canvas()` since the active canvas did not necessarily change.

This is just helping us to make sure that if we have a function called `active_canvas_changed()`, that it only gets called when the active canvas actually changes.

hexrd/ui/image_tab_widget.py
```diff
@@ -53,24 +56,13 @@ def setup_connections(self):
         HexrdConfig().detectors_changed.connect(self.reset_index)
 
     def clear(self):
-        # This removes all canvases except the first one,
-        # and it calls super().clear()
-
-        for canvas, cid in zip(self.image_canvases[1:],
-                               self.mpl_connections[1:]):
-            canvas.mpl_disconnect(cid)
-            canvas.deleteLater()
```
psavery, line 62: Do you know why we were originally doing this? Is removing it going to have any side effects for other parts of the code?

psavery (approved):
Can we change the window title name from "Dialog" to be something else, like "Draw Polygonal Mask"?

Other than the name changes, this looks like it is working, and LGTM! Thank you @bnmajor!

hexrd/ui/main_window.py
bnmajor: Changed

hexrd/ui/main_window.py
bnmajor: Fixed

hexrd/ui/main_window.py
bnmajor: Fixed

hexrd/ui/image_tab_widget.py
bnmajor: According to git blame this was to prevent a bug where signals were sometimes still connected to the canvases when the tabs were not being used ([here](https://github.com/HEXRD/hexrdgui/commit/043f1ef0d1006e090af02ec3b8eb9397d2407be2)). Without this change we face other issues (like we need to re-render the canvas with the masks but this re-creates the canvases and the signals for masking are broken). I think that if there is an issue with signals not being disconnected properly we should figure that out rather than just deleting and re-creating every time though.

bnmajor:
> Can we change the window title name from "Dialog" to be something else, like "Draw Polygonal Mask"?

Done :+1:

hexrd/ui/image_tab_widget.py
psavery: Yes I agree! Thanks for looking that up!

psavery (approved):
LGTM

PR #1544, opened 2023-07-18
Threshold masking

### LINKED ISSUE #1528: Threshold mask
Currently we have functionality to view pixels over a certain value. We should add functionality to mask pixels based on threshold values.

The user can specify the min and max and any pixels (in the raw view) outside that range would be made a `nan`. The use case right now is images recorded on the Rigaku at LLNL. There are a lot of dead pixels on the detector which have values ~`~1E10` and `+1E10`. Could also be useful for masking Laue spots since they are way brighter than powder spots.

bnmajor:
Fixes #1528


hexrd/ui/threshold_mask_dialog.py
```diff
@@ -0,0 +1,82 @@
...
+            functools.partial(self.setup_gui, reset=True))
+
+        HexrdConfig().mgr_threshold_mask_changed.connect(self.setup_gui)
+
+    def gather_input(self):
+        self.comparisons.clear()
+        self.values.clear()
+
+        val = self.ui.first_value.value()
+        if val != math.inf and val != -math.inf:
```
psavery, line 65: This could also be `np.isinf(val)` (which returns `True` for both positive and negative infinity), but it doesn't matter.

hexrd/ui/threshold_mask_dialog.py
```diff
@@ -0,0 +1,82 @@
...
+            idx = self.ui.second_comparison.currentIndex()
+            self.values.append(val)
+            self.comparisons.append(idx)
+
+    def accept(self):
+        self.gather_input()
+        HexrdConfig().threshold_comparisons = self.comparisons
+        HexrdConfig().threshold_values = self.values
+        apply_threshold_mask()
+        HexrdConfig().threshold_mask_changed.emit('threshold')
```
psavery, line 81: I wonder if this should be emitted in `apply_threshold_mask()` instead?

hexrd/ui/resources/ui/image_mode_widget.ui
```diff
@@ -60,71 +60,19 @@
...
-           </property>
-           <property name="maximum">
-            <double>100000.000000000000000</double>
-           </property>
-          </widget>
-         </item>
-        </layout>
-       </item>
        <item row="0" column="0">
-        <layout class="QVBoxLayout" name="verticalLayout_3">
```
psavery, line 116: I think it might be better if these checkboxes are stacked vertically instead of horizontally, with `Tabbed View` being on top.

hexrd/ui/threshold_mask_dialog.py
```diff
@@ -0,0 +1,82 @@
...
+
+        idx = comps[1] if len(comps) > 1 else 1
+        self.ui.second_comparison.setCurrentIndex(idx)
+        val = vals[1] if len(vals) > 1 else math.inf
+        self.ui.second_value.setValue(val)
+
+    def setup_connections(self):
+        self.ui.button_box.accepted.connect(self.accept)
+        apply_button = self.ui.button_box.button(QDialogButtonBox.Apply)
+        apply_button.clicked.connect(self.accept)
```
psavery, line 53: Can we remove the "OK/Cancel" buttons entirely? Users can just click "Apply" to apply the mask, and then "X" in the top corner to exit.

psavery (commented):
Looking great so far! A few things to fix:

- [ ] Toggling visibility in the mask manager dialog for the threshold mask doesn't appear to do anything
- [ ] If you remove the threshold mask from the mask manager and click "Apply" in the threshold dialog, the threshold mask doesn't come back (it looks like you have to exit the program to be able to make a new threshold mask)
- [ ] I think we should consider eliminating the combo boxes and just having it display two spinboxes that are a range of values that *will not* be thresholded. This could default to `-inf to inf`, and if you wanted to, for instance, threshold all values below 0 and above 50, it would be `0 to 50`. But let's see if this is okay with Saransh first. I think the text is a little confusing anyways - does "Less than" mean that all values less than that value will be thresholded (masked out) or kept? Looks like it means those values should be masked out. But if I try to keep only pixels between a certain range such as 0 and 50, it doesn't look like I can do that for some reason.

hexrd/ui/create_raw_mask.py
```diff
@@ -10,29 +10,33 @@
...
+    masks = []
+    for (comparison, value) in zip(comparisons, values):
+        if comparison == constants.UI_THRESHOLD_LESS_THAN:
+            masks.append(img < value)
+        elif comparison == constants.UI_THRESHOLD_GREATER_THAN:
+            masks.append(img > value)
+        elif comparison == constants.UI_THRESHOLD_EQUAL_TO:
+            masks.append(img == value)
+    mask = np.logical_and.reduce(masks)
+    return ~mask
```
psavery, line 39: This might not be working properly, because when I try to threshold values so that I only keep pixels between, for instance, 0 and 50, it appears that nothing gets thresholded.

bnmajor:
> If you remove the threshold mask from the mask manager and click "Apply" in the threshold dialog, the threshold mask doesn't come back (it looks like you have to exit the program to be able to make a new threshold mask)

This seems to work for me, but please let me know if I am misunderstanding what you mean:
thresholding

> I think we should consider eliminating the combo boxes and just having it display two spinboxes that are a range of values that _will not_ be thresholded. This could default to `-inf to inf`, and if you wanted to, for instance, threshold all values below 0 and above 50, it would be `0 to 50`. But let's see if this is okay with Saransh first.

Sounds good! So something like this maybe?
new_thresholding
My only concern is the default range -  if we are defaulting to `(-inf, inf)` doesn't this kind of impy that everything is being masked out? Maybe that's not really an issue since that's not how it would ever be used (and we can note the behavior in a tooltip like we are now) but I'm just wondering if it may be confusing. @saransh13 Any thoughts on behavior/dialog/defaults? This simplification would eliminate support for masking values `equal to` a certain value but maybe that is a non-issue.

> I think the text is a little confusing anyways - does "Less than" mean that all values less than that value will be thresholded (masked out) or kept? Looks like it means those values should be masked out. But if I try to keep only pixels between a certain range such as 0 and 50, it doesn't look like I can do that for some reason.

Yes, after [asking Saransh](https://github.com/HEXRD/hexrdgui/issues/1528#issuecomment-1638868117) we should be `and`-ing masks but this is confusing with the current dialog, I agree. Another point in favor of your simplified suggestion honestly :+1:

hexrd/ui/mask_manager_dialog.py
```diff
@@ -69,11 +69,18 @@ def update_masks_list(self, mask_type):
...
+                name = unique_name(self.masks, 'threshold')
+                self.masks[name] = ('threshold', HexrdConfig().threshold_masks)
+                HexrdConfig().visible_masks.append(name)
+                self.threshold = name
+                self.toggle_visibility(True, self.threshold)
+            elif self.threshold and not HexrdConfig().threshold_masks:
+                items = self.ui.masks_table.findItems(
+                    self.threshold, Qt.MatchExactly)
+                if items:
+                    self.remove_mask(items[0].row(), self.threshold)
```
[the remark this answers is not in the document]
bnmajor: This is already happening [here](https://github.com/HEXRD/hexrdgui/blob/f748a9ff0431a9ab2c86a595bc5ee6779664721c/hexrd/ui/mask_manager_dialog.py#L145) (called from [remove_mask](https://github.com/HEXRD/hexrdgui/blob/f748a9ff0431a9ab2c86a595bc5ee6779664721c/hexrd/ui/mask_manager_dialog.py#L163))

hexrd/ui/threshold_mask_dialog.py
bnmajor: That's much cleaner, thank you :slightly_smiling_face:

hexrd/ui/resources/ui/image_mode_widget.ui
bnmajor: Sounds good, fixed.

hexrd/ui/threshold_mask_dialog.py
bnmajor: Done :+1:

hexrd/ui/create_raw_mask.py
bnmajor: Yes, see the [comment below](https://github.com/HEXRD/hexrdgui/pull/1544#issuecomment-1662262587).

psavery:
> > If you remove the threshold mask from the mask manager and click "Apply" in the threshold dialog, the threshold mask doesn't come back (it looks like you have to exit the program to be able to make a new threshold mask)
> 
> This seems to work for me, but please let me know if I am misunderstanding what you mean: 

I'll try it again tonight! (tomorrow for me)

> Sounds good! So something like this maybe? 

Yes that looks good!

> My only concern is the default range - if we are defaulting to `(-inf, inf)` doesn't this kind of impy that everything is being masked out? Maybe that's not really an issue since that's not how it would ever be used (and we can note the behavior in a tooltip like we are now) but I'm just wondering if it may be confusing. @saransh13 Any thoughts on behavior/dialog/defaults? This simplification would eliminate support for masking values `equal to` a certain value but maybe that is a non-issue.

So, the range we are defining is the range of pixel values **that will not be masked out**, not the range of pixel values that will be masked out. I confirmed this with Saransh (and he thinks the range is better too).

So that means if it defaults to `(-inf, inf)`, no pixels get masked out. Then if we change it to be `(0, 50)`, then all pixels that do not have a value between `0` and `50` will be masked out. Does that sound good?

bnmajor:
> So, the range we are defining is the range of pixel values **that will not be masked out**, not the range of pixel values that will be masked out. I confirmed this with Saransh (and he thinks the range is better too).
> 
> So that means if it defaults to `(-inf, inf)`, no pixels get masked out. Then if we change it to be `(0, 50)`, then all pixels that do not have a value between `0` and `50` will be masked out. Does that sound good?

Ah, okay. This is the opposite of what I thought he was asking for. In that case this makes sense. Thanks!

bnmajor:
> * [ ]  Toggling visibility in the mask manager dialog for the threshold mask doesn't appear to do anything

This should be fixed now!

hexrd/ui/threshold_mask_dialog.py
bnmajor: Makes sense, moved it.

hexrd/ui/create_raw_mask.py
bnmajor: This is fixed now in [daf738f](https://github.com/HEXRD/hexrdgui/pull/1544/commits/daf738f00d2677c614b1fa13de63ff01edf6404d)

bnmajor:
@psavery @saransh13 I think this is all fixed up now and ready for a final review

psavery:
One more issue I see:

If I set the range to be `0` to `inf`, I get this error:

```python
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/threshold_mask_dialog.py", line 68, in accept
    apply_threshold_mask()
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/create_raw_mask.py", line 21, in apply_threshold_mask
    mask = create_threshold_mask(img)
  File "/Users/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/create_raw_mask.py", line 29, in create_threshold_mask
    lt_val, gt_val = HexrdConfig().threshold_values
ValueError: not enough values to unpack (expected 2, got 1)
```

We should also make sure setting one of the values in the range to `inf` or `-inf` works for saving/loading the mask too.

We also need to modify the cartesian and polar generation to take this masking into account, but I will look into that in a separate PR!

bnmajor:
@saransh13 I tested the example that you shared with me but it seems to work for me. Can you make sure your branch is up-to-date and try again?

@psavery Thanks for that! We're now saving the range regardless of value which fixes that bug.

hexrd/ui/create_raw_mask.py
psavery: Looks like it is working now!

hexrd/ui/resources/ui/threshold_mask_dialog.ui
```diff
@@ -0,0 +1,96 @@
...
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Threshold Masking</string>
+  </property>
+  <layout class="QGridLayout" name="gridLayout">
+   <item row="0" column="0">
+    <widget class="QLabel" name="comparison_label">
+     <property name="text">
+      <string>Range</string>
```
psavery, line 20:
```suggestion
      <string>Unmasked Range</string>
```

Can we change the text to "Unmasked Range" for clarity?

hexrd/ui/mask_manager_dialog.py
```diff
@@ -250,6 +239,8 @@ def write_all_masks(self, h5py_group=None):
             for i, (det, mask) in enumerate(data):
                 parent = d.setdefault(det, {})
                 parent.setdefault(name, {})[str(i)] = mask
+        if self.threshold:
+            d['threshold'] = HexrdConfig()._threshold_data
```
psavery, line 243: I tried writing out just the masks (via "Export Masks"), and sometimes, this 'threshold' group in the HDF5 file is just an empty group. This produces an error when we load in the masks.

Maybe there are some situations where `HexrdConfig()._threshold_data` is invalid?

But I can't actually reproduce this issue now. Let's keep an eye out for it.

psavery (approved):
Let's just do the name change ("Range" -> "Unmasked Range") and then this PR looks good to me!

psavery (approved):
LGTM

PR #1553, opened 2023-08-02
Ensure overlay visibility checkbox is updated on state load

### LINKED ISSUE #1552: Overlays visible on state load
After loading a state file, the `show overlay` check box is off but the overlays are visible on the canvas. Screenshot below
Screen Shot 2023-08-02 at 10 16 53 AM

bnmajor:
Fixes #1552

### DIFF
--- a/hexrd/ui/main_window.py
+++ b/hexrd/ui/main_window.py
@@ -333,6 +333,7 @@ class MainWindow(QObject):
 
     def on_state_loaded(self):
         self.update_action_check_states()
+        self.materials_panel.update_gui_from_config()
 
     def update_action_check_states(self):
         checkbox_to_hexrd_config_mappings = {


psavery:
This looks like it works, but emitting this signal might have other side effects like regenerating overlays (the signal is primarily intended to communicate that overlays need to be regenerated).

Could we instead try adding `self.materials_panel.update_gui_from_config()` to [here](https://github.com/HEXRD/hexrdgui/blob/ebbd67518a3fb09e243f9f7fc68e88c9291a4b03/hexrd/ui/main_window.py#L335)?

bnmajor:
> Could we instead try adding `self.materials_panel.update_gui_from_config()` to [here](https://github.com/HEXRD/hexrdgui/blob/ebbd67518a3fb09e243f9f7fc68e88c9291a4b03/hexrd/ui/main_window.py#L335)?

Sure thing.

psavery:
> > Could we instead try adding `self.materials_panel.update_gui_from_config()` to [here](https://github.com/HEXRD/hexrdgui/blob/ebbd67518a3fb09e243f9f7fc68e88c9291a4b03/hexrd/ui/main_window.py#L335)?
> 
> Sure thing.

That needs to be changed to `self.materials_panel.update_gui_from_config()`. Can you also verify that fixes the issue?

bnmajor:
> That needs to be changed to `self.materials_panel.update_gui_from_config()`. Can you also verify that fixes the issue?

I'm sorry, I read your original comment too fast. This is fixed now and that does work.

psavery:
> > That needs to be changed to `self.materials_panel.update_gui_from_config()`. Can you also verify that fixes the issue?
> 
> I'm sorry, I read your original comment too fast. This is fixed now and that does work.

Great! I think this makes more sense too, since updating the GUI is what we are really doing.

psavery (approved):
LGTM

PR #1554, opened 2023-08-03
Do not rescale the cartesian and polar views

psavery:
We have been rescaling the cartesian and polar views for quite a while, but it does not actually appear to be necessary.

The pixel values from the raw view seem to be maintained when they are transformed to the cartesian/polar view. Rescaling the values in the cartesian/polar view is sometimes causing problems when there are, for instance, very high and low values.

### DIFF
--- a/hexrd/ui/calibration/cartesian_plot.py
+++ b/hexrd/ui/calibration/cartesian_plot.py
@@ -289,15 +289,6 @@ class InstrumentViewer:
 
         # In case there were any nans...
         nan_mask = np.isnan(img)
-        img = np.ma.masked_array(img, mask=nan_mask, fill_value=0.)
-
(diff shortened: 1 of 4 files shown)


psavery:
@saransh13 Can you try this out for a few datasets and confirm that there are no issues in how it appears?

This might change the appearance of the cartesian/polar views in some cases. However, I think it is better to just not rescale the intensity at all, so this is hopefully better.

psavery:
If this looks good for a few datasets when you try them out, we can merge it!

saransh13:
I tried it out and looks like it's fixing the scaling issue. This was quite an annoyance and I'm glad you figured it out!

saransh13 (approved):
LGTM

PR #1558, opened 2023-08-08
Mask with NaNs in the raw view as well

### LINKED ISSUE #1555: Mask with nans in raw view
We should mask with NaN values [instead of 0](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/hexrd_config.py#L892) in the raw view to keep it consistent with the other views.

bnmajor:
Fixes #1555

### DIFF
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -890,21 +890,25 @@ class HexrdConfig(QObject, metaclass=QSingleton):
 
         return masks_dict
 
+    @property
+    def masked_images_dict(self):
+        return self.create_masked_images_dict()
+
     def create_masked_images_dict(self, fill_value=0):
         """Get an images dict where masks have been applied"""
         images_dict = self.images_dict
-        have_masks = len(self.visible_masks)
+        if not self.visible_masks:
+            # Force a fill_value of 0 if there are no visible masks
+            fill_value = 0
+
         for det, mask in self.raw_masks_dict.items():
             for name, img in images_dict.items():
-                if (have_masks and
-                        np.issubdtype(type(fill_value), np.floating) and
+                if (np.issubdtype(type(fill_value), np.floating) and
                         not np.issubdtype(img.dtype, np.floating)):
                     img = img.astype(float)
+                    images_dict[name] = img
                 if det == name:
-                    # Masks always exist but may not be masking any regions out
-                    # Only use the fill value if any of the masks are actually
-                    # being applied (visible).
-                    img[~mask] = fill_value if have_masks else 0
+                    img[~mask] = fill_value
 
         return images_dict
 
--- a/hexrd/ui/hexrd_config.py
+++ b/hexrd/ui/hexrd_config.py
@@ -896,12 +896,25 @@ class HexrdConfig(QObject, metaclass=QSingleton):
 
     def create_masked_images_dict(self, fill_value=0):
         """Get an images dict where masks have been applied"""
+        from hexrd.ui.create_hedm_instrument import create_hedm_instrument
+
         images_dict = self.images_dict
-        if not self.visible_masks:
+        instr = create_hedm_instrument()
+
+        has_masks = bool(self.visible_masks)
+        has_panel_buffers = any(panel.panel_buffer is not None
+                                for panel in instr.detectors.values())
+
+        if not has_masks and not has_panel_buffers:
             # Force a fill_value of 0 if there are no visible masks
+            # and no panel buffers.
             fill_value = 0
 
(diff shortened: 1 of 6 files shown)


psavery (approved):
As long as you tried it out (including switching to different views), then it looks good to me!

bnmajor:
@psavery Marking as WIP for now, I'm getting bugs in testing. I'll tag you again once they're fixed.

bnmajor:
@psavery @saransh13 Using NaNs in the raw view requires that all raw image data be floats. Will there be issues if we cast data to float?

saransh13:
I think we implicitly assume floats for all our other operations, so I
don’t see a problem doing that. Some testing should tell if that’s not the
case.

bnmajor:
@psavery This "fixes" the issue... although I'm not sure if it is the solution we want. Right now, before masking, we check to see if the image data type in integer and we cast it to float if it is. Should this maybe just automatically happen on image load so we only store the raw image data as floats?

psavery:
@bnmajor I'll check on this in more detail soon. Maybe one solution would be to used a [masked array](https://numpy.org/doc/stable/reference/maskedarray.html) instead, and different parts of the code can treat the mask differently (like the raw image viewer can convert the masked array to a plain array with nans).

hexrd/ui/hexrd_config.py
```diff
@@ -888,8 +888,10 @@ def masked_images_dict(self):
         images_dict = self.images_dict
         for det, mask in self.raw_masks_dict.items():
             for name, img in images_dict.items():
+                if np.issubdtype(img.dtype, np.integer):
+                    img = img.astype(float)
```
psavery, line 892: FYI: this currently isn't working, because `img.astype(float)` is creating a new array (not modifying the existing array in the `images_dict`) and creating a new `img` variable from that new array. So the `images_dict` that is returned doesn't contain the modified array.

psavery:
So here are the places where that `HexrdConfig().masked_images_dict` is being used. We would need to confirm that all of them work properly if we modify `HexrdConfig().masked_images_dict`:

* The fast powder calibration
* Laue and powder calibration (used during auto-picking for both powder and Laue, and used for the actual calibration)
* Stereo view drawing from raw (although that is currently not working anyways)
* The raw view

If we modify the output of `masked_images_dict`, we need to verify that all of these work properly.

Alternatively, I think we can do something else that will require less checking. Can we take the logic in `masked_images_dict`, and put it inside a new function on `HexrdConfig()` called `create_masked_images_dict`? That `create_masked_images_dict` will have a single argument `fill_value` that defaults to `0`. `HexrdConfig().masked_images_dict` will return the output of that function with the default argument.

Then, in the image canvas code [here](https://github.com/HEXRD/hexrdgui/blob/69f7df28906f5017ec2464409903792f7cb59800/hexrd/ui/image_canvas.py#L200), we can return `HexrdConfig().create_masked_images_dict(fill_value=np.nan)`. Then in `create_masked_images_dict()`, we can have that logic to convert to a float if the fill value is a float (including `np.nan`, but only if there are masks present).

That's just a possibility for a less-invasive change. We can still modify the output of `HexrdConfig().masked_images_dict`, but if we do, we need to verify that everything is working properly.

By the way, about the masked arrays, it looks like you can't set a fill value of `np.nan` if the dtype of the underlying array is an integer. So if we want to display nans for the masked regions in the raw view, then we definitely need to convert the array to float64 at some point. I think it's fine if we do that only during the generation of the raw view, if there are masks and the fill value is `np.nan`. The less invasive changes above would do that.

saransh13:
Is this ready for merge?

psavery:
I made some minor modifications. I think it looks good other than one thing: the panel buffer regions should also have `np.nan` in them, to match the other views. Try this state file, for example: https://github.com/HEXRD/examples/blob/master/state_examples/Epix_fast_powder/epix_fast_powder.h5

bnmajor:
> I made some minor modifications. I think it looks good other than one thing: the panel buffer regions should also have `np.nan` in them, to match the other views. Try this state file, for example: https://github.com/HEXRD/examples/blob/master/state_examples/Epix_fast_powder/epix_fast_powder.h5

Should this change be a part of this PR or part of another one where we've re-thought how to better mesh panel buffers with the masking tools?

psavery:
In the polar view, the panel buffer values are already nans. We might want to go ahead and do it here too for consistency. Or it can be in another PR. What do you think, @saransh13?

saransh13:
Let’s go ahead and do it in this PR.

bnmajor:
> Let’s go ahead and do it in this PR.

@psavery pushed the additional changes we need, the branch looks good in my testing :white_check_mark:

PR #1559, opened 2023-08-08
Fix threshold masking in polar and stereo views

bnmajor:


saransh13 (approved):
Tested this with the Rigaku dataset. The big negative numbers are correctly masked in the polar view to nans, so the lineout looks like it should.
LGTM

psavery:
> The last approach was not masking the dead pixels in the polar view. The new approach works but I think we can improve it in the future.

Sounds good. There might have been some things we could do to get the last approach to work, but if this one works fine then let's do it!

PR #1561, opened 2023-08-23
Make sure threshold masks are appended to existing masks

### LINKED ISSUE #1560: New rectangular/ellipse mask when threshold mask is applied.
When I have the threshold mask on and I try to draw a new rectangular/ellipse mask, then the new region is not masked off. If I turn off the threshold mask, then those masks show up.

bnmajor:
Previously threshold masks were overwriting any other masks in the polar view.

Fixes #1560


saransh13 (approved):
LGTM

psavery (approved):
LGTM

PR #1562, opened 2023-08-24
Create a copy of threshold masks when saving to state

bnmajor:
Otherwise if threshold masks are saved out to the currently loaded state file they were wiped from the state.


saransh13 (approved):
Fixes the issue.
LGTM!

PR #1569, opened 2023-09-06
Masking boundaries

bnmajor:
NEEDS:
- [x] Re-test LLNL workflow to make sure InteractiveTemplate changes have not broken anything
- [x] Re-test mask regions by rectangle/ellipse
  - [x] Toggle tabbed mode while masking
  - [x] Toggle view mode while masking
  - [x] Saving/loading state files with masks

KNOWN ISSUES:
- [x] In the raw, untabbed view mode rotation does not work (the template disappears) on any detector that is not the first. Rotation by arrow key works, translate works, and everything works in tabbed mode.
- [x] Rotation in the polar view skews the template
- [x] Create template(s) -> interact w/ another dialog (i.e. mask manager) -> try to draw another mask -> nothing happens (works again if the shape is toggled and then mask is drawn)
- [x] ~Draw template(s) in raw -> switch to tabbed view -> draw template but do not save it -> switch to polar view -> cannot immediately draw on canvas~


hexrd/ui/hexrd_config.py
```diff
@@ -874,7 +874,8 @@ def raw_masks_dict(self):
                 for det, mask in data:
                     if det == name:
                         final_mask = np.logical_and(final_mask, mask)
-            if self.threshold_mask_status:
+            if (self.threshold_mask_status and
+                    self.threshold_masks.get(name, None) is not None):
```
psavery, line 878:
```suggestion
                    self.threshold_masks.get(name) is not None):
```

If `self.threshold_masks` is a dictionary, then `get()` will return `None` by default if `name` is not present.

hexrd/ui/interactive_template.py
```diff
@@ -27,11 +23,48 @@ def __init__(self, parent=None):
         self.translation = [0, 0]
         self.complete = False
         self.event_key = None
-        self.parent.setFocusPolicy(Qt.ClickFocus)
+        self.detector = detector
+        self.instrument = instrument
+        self._static = True
+
+        for canvas in self.image_tab_widget.active_canvases:
+            canvas.setFocusPolicy(Qt.ClickFocus)
```
psavery, line 31: Maybe we should just put `self.setFocusPolicy(Qt.ClickFocus)` in the `ImageCanvas.__init__()` method instead?

hexrd/ui/interactive_template.py
```diff
@@ -8,17 +8,13 @@
...
 from hexrd.ui.utils import has_nan
 
 
 class InteractiveTemplate:
-    def __init__(self, parent=None):
-        self.parent = parent.image_tab_widget.image_canvases[0]
-        self.ax = self.parent.axes_images[0]
-        self.panels = create_hedm_instrument().detectors
+    def __init__(self, parent, detector, instrument=None):
+        self.image_tab_widget = parent.image_tab_widget
```
psavery, line 17: If `parent` isn't actually used for anything, then it might be better for us to just change that argument to `image_tab_widget` instead, and pass the `image_tab_widget`.

hexrd/ui/interactive_template.py
```diff
@@ -27,11 +23,48 @@ def __init__(self, parent=None):
...
+        self.instrument = instrument
+        self._static = True
+
+        for canvas in self.image_tab_widget.active_canvases:
+            canvas.setFocusPolicy(Qt.ClickFocus)
+
+    @property
+    def current_canvas(self):
+        idx = self.image_tab_widget.current_index
+        return self.image_tab_widget.active_canvases[idx]
```
psavery, line 36: This is very nice!

hexrd/ui/interactive_template.py
```diff
@@ -27,11 +23,48 @@ def __init__(self, parent=None):
...
+
+    @property
+    def ax(self):
+        idx = HexrdConfig().current_imageseries_idx
+        return self.current_canvas.axes_images[idx]
```
psavery, line 41: I am under the impression that this might not be correct. Isn't the `current_imageseries_idx` used for viewing different frames in an unaggregated image series? But those `axes_images` are instead the separate images for each detector in the untabbed raw view.

hexrd/ui/interactive_template.py
```diff
@@ -27,11 +23,48 @@ def __init__(self, parent=None):
...
 
     @property
     def raw_axes(self):
-        return list(self.parent.raw_axes.values())[0]
+        if not self.current_canvas.raw_axes:
+            return self.current_canvas.axis
```
psavery, line 46: Maybe the property `raw_axes` should just be renamed to `axis` since it is not necessarily raw?

hexrd/ui/interactive_template.py
```diff
@@ -41,20 +74,17 @@ def rotate_shape(self, angle):
         self.rotate_template(self.shape.xy, angle)
         self.redraw()
 
-    def create_shape(self, module, file_name, det, instr):
+    def create_polygon(self, verts, **kwargs):
```
psavery, line 77: Can you rename `kwargs` to `polygon_kwargs` for clarity?

hexrd/ui/interactive_template.py
```diff
@@ -27,11 +23,48 @@ def __init__(self, parent=None):
...
+
+    @static_mode.setter
+    def static_mode(self, mode):
+        if mode == self._static:
+            return
+
+        if mode:
+            self.disconnect()
+        else:
+            self.connect_translate_rotate()
```
psavery, line 66: Just to be overly code-defensive (and prevent potential bugs), can we call `disconnect()` at the beginning of `self.connect_translate_rotate()` to ensure any previous connections are disconnected?

Additionally, we can make sure `disconnect()` can be safely called if 1. no connections were made or 2. `disconnect()` has already been called once before.

hexrd/ui/interactive_template.py
```diff
@@ -41,20 +74,17 @@ def rotate_shape(self, angle):
...
             # This template contains more than one polygon and the last point
             # should not be connected to the first. See Tardis IP for example.
             self.shape.set_closed(False)
-        self.shape_styles.append({'line': '-', 'width': 1, 'color': 'cyan'})
-        self.update_position(instr, det)
-        self.connect_translate_rotate()
+        self.shape_styles.append(kwargs)
+        self.update_position()
+        if not self.static_mode:
+            self.connect_translate_rotate()
```
psavery, line 87: You could potentially put the `if self.static_mode` check inside the `self.connect_translate_rotate()` function as well. But you can decide where it goes.

hexrd/ui/interactive_template.py
```diff
@@ -149,11 +183,11 @@ def toggle_boundaries(self, show):
...
-        self.parent.mpl_disconnect(self.button_press_cid)
-        self.parent.mpl_disconnect(self.button_release_cid)
-        self.parent.mpl_disconnect(self.motion_cid)
-        self.parent.mpl_disconnect(self.key_press_cid)
-        self.parent.mpl_disconnect(self.button_drag_cid)
+        self.current_canvas.mpl_disconnect(self.button_press_cid)
+        self.current_canvas.mpl_disconnect(self.button_release_cid)
+        self.current_canvas.mpl_disconnect(self.motion_cid)
+        self.current_canvas.mpl_disconnect(self.key_press_cid)
+        self.current_canvas.mpl_disconnect(self.button_drag_cid)
```
psavery, line 190: You might want to put `canvas = self.current_canvas` in the first line so that the `self.current_canvas` property only gets evaluated once. Same for `self.connect_translate_rotate()`.

bnmajor:
@psavery @saransh13 While I wrap this branch up I'd welcome any thoughts or opinions on the best way to interact with the rectangle/ellipse masking. Right now the user:

1. Selects the shape
2. Draws the bounds
3. Translates/rotates the bounds
4. Right-clicks to complete the current mask and enable drawing the next

Do we need an additional or different option for indicating that the current mask is complete? Something that indicates that the template is in "interactive" (can be moved) mode still? I was thinking about the current mask being drawn in red and completed masks being drawn in black or something along those lines.

mask_region_dialog

hexrd/ui/mask_regions_dialog.py
```diff
@@ -99,42 +98,27 @@ def update_patch(self, event):
...
             center = [(width / 2 + x0), (height / 2 + y0)]
-            self.patch.set_center(center)
-            self.patch.height = height
-            self.patch.width = width
+            shape = patches.Ellipse(center, width, height)
+        verts = shape.get_patch_transform().transform(
+            shape.get_path().vertices[:-1])
+        verts = add_sample_points(verts, 300)
+        self.patch.template.set_xy(verts)
+        self.patch.center = self.patch.get_midpoint()
```
psavery, line 109: Looking very nice!

hexrd/ui/mask_regions_dialog.py
```diff
@@ -86,11 +88,8 @@ def create_patch(self):
             'fill': False,
             'animated': True,
         }
-        if self.selection == 'Rectangle':
-            self.patch = patches.Rectangle((0, 0), 0, 0, **kwargs)
-        elif self.selection == 'Ellipse':
-            self.patch = patches.Ellipse((0, 0), 0, 0, **kwargs)
-        self.axes.add_patch(self.patch)
+        self.patch = InteractiveTemplate(self.parent, self.det)
```
psavery, line 91: Should we rename `self.patch` to something else, since it's actually an interactive template and not a patch?

psavery (commented):
Looking really nice so far! I have mostly minor suggestions for the code. I'll give it a try soon too.

hexrd/ui/mask_regions_dialog.py
bnmajor: Yeah, that makes sense! I think the simplest is best: `self.patch` -> `self.interactive_template`, or we can just use `self.it` like we do in the LLNL import tool. Any preference?

hexrd/ui/hexrd_config.py
bnmajor: Cleaned up :+1:

hexrd/ui/interactive_template.py
bnmajor: Sounds good

hexrd/ui/interactive_template.py
bnmajor: This is actually removed altogether with [8ebecb9](https://github.com/HEXRD/hexrdgui/pull/1569/commits/8ebecb923eb7a5016946d2aab5ea2e43561a3bc3) now because we are using signals instead of passing around the main window (like we did with the `HandDrawnMaskDialog`)

hexrd/ui/interactive_template.py
bnmajor: You are correct, thanks for catching this! This was in fact the reason for the rotation bug as well. Thanks!

hexrd/ui/interactive_template.py
bnmajor: That makes sense, updated!

hexrd/ui/interactive_template.py
bnmajor: Sure thing, done.

hexrd/ui/interactive_template.py
bnmajor: Done

hexrd/ui/interactive_template.py
bnmajor: Done

hexrd/ui/interactive_template.py
bnmajor: No longer applicable

hexrd/ui/mask_regions_dialog.py
bnmajor: Yup, updated.

psavery:
@bnmajor Did you try rotating the masks in the polar view? The behavior appears a little odd.

Ellipses in the raw view don't look as smooth as they used to. Do you know why?

If I draw a mask and then right-click to save it, then click "Undo Last Selection", I get an exception.

Also, can we make sure all of the signals get disconnected when the mask manager is hidden? It looks to me that if I use the mask manager several times in one session, the program slows down.

Also, we should get feedback from @saransh13, but it might be more intuitive to change the clicking behavior to be like this:

1. If the user left-clicks inside any template, then translate that template
2. If the user left-clicks outside of any template, then save the previous template and start a new template
3. Right-click won't do anything

However, I wonder if it will be difficult to allow the user to translate templates that were already saved (like allow them to translate any template they have already created). If that would be difficult, maybe we can skip that part.

Overall, it is looking very nice, though!

bnmajor:
> @bnmajor Did you try rotating the masks in the polar view? The behavior appears a little odd.

I did not test rotate in the polar view but there is indeed some strange behavior... Rotation causes some strange skewing. I will get that sorted out.

> Ellipses in the raw view don't look as smooth as they used to. Do you know why?

We're now drawing the ellipses with the more generic [`patches.Polygon`](https://github.com/HEXRD/hexrdgui/pull/1569/files#diff-2b95b6927504eaee327e5da150cda9d5f158b0bfa1c610041217c730cc3f8494R67-R77) in the interactive template. We still use the `Eliipses` polygon for the initial drawing so that we can easily draw the shape and then grab the coordinates from it. The `InteractiveTemplate` uses Polygon for everything to stay as general as possible.

> Also, can we make sure all of the signals get disconnected when the mask manager is hidden? It looks to me that if I use the mask manager several times in one session, the program slows down.

I've added a commit with a lot more precautionary `disconnect` calls that will hopefully help with this problem!

bnmajor:
> If I draw a mask and then right-click to save it, then click "Undo Last Selection", I get an exception.
> 
> Also, we should get feedback from @saransh13, but it might be more intuitive to change the clicking behavior to be like this:
> 
> 1. If the user left-clicks inside any template, then translate that template
> 2. If the user left-clicks outside of any template, then save the previous template and start a new template
> 3. Right-click won't do anything
> 
> However, I wonder if it will be difficult to allow the user to translate templates that were already saved (like allow them to translate any template they have already created). If that would be difficult, maybe we can skip that part.

I think that this is all much smoother now! I've made a few updates and the behavior is now as follows:
1. User draws a shape in a region
2. Any existing template can be manipulated by selecting it
    1. 1. Click on a template and it will turn red if selected
    2. Translate and rotate as before
4. Click away from the selected template or simply begin drawing to create a new template shape
5. All templates can be undone now, not just interactive templates

psavery (approved):
This is working well for me, thanks @bnmajor!

The one thing we might want to do in a future PR: if users are interacting with these masks on a high-resolution polar canvas, we probably want to animate the interactions so that they are much faster. But we will work on that only if someone requests it.

@saransh13, can you try this out and let us know if you have any suggested changes?

saransh13:
👍🏽 I’ll have feedback on this PR later today.

bnmajor:
After speaking with @saransh13 this branch needs the following:
- [ ] Larger step size for rotate by arrow key. This should only apply to masking, not the LLNL import tool
- [ ] Bug fix: in polar view draw mask A -> move mask A -> draw mask B -> move mask B -> mask A moves as well

saransh13:
Forgot to mention, the bug was with the arrow key. I did not test the same
behavior with the Mouse.

bnmajor:
> * [x]  Larger step size for rotate by arrow key. This should only apply to masking, not the LLNL import tool
> * [x]  Bug fix: in polar view draw mask A -> move mask A -> draw mask B -> move mask B -> mask A moves as well

These are both fixed, although the angle of rotation was bumped an arbitrary amount. This value can be set to anything [here](https://github.com/HEXRD/hexrdgui/pull/1569/files#diff-9d49f0b9c78902698358671f3b2652c8363eea7c0c34f21e4fd8674a4740e424R90). @saransh13 @psavery

psavery (approved):
Feel free to merge when ready. We will animate the interactive template in a separate PR, and fix any more issues we encounter in future PRs.

Issue #1566, opened 2023-08-29
Calibration point picker lineout behaviot


akrygier:
Currently when you pick points by hand in the "Composite" calibration mode, there is a vertical and horizontal lineout. I think it is averaging over the full displayed range, but I was thinking it might be more useful to have the lineout correspond to a single pixel of average over just a few. I think this is a vestige of the previously narrower default range.

<img width="622" alt="image" src="https://github.com/HEXRD/hexrdgui/assets/59886900/6815c395-e1ac-49c2-a83e-6edda0de9432">

psavery:
Hi @akrygier. Yes, by default those lineout plots are plotting the sums of the pixel values across the corresponding rows/columns.

Can you try unchecking "Display sums in subplots"? Then, hover your mouse over the zoom canvas. Instead of plotting the sums, the lineouts will plot the pixel values that intersect with the cursor. 

Note [here](https://youtu.be/1897xY_4WSk?si=-OS5Wud5-7eLa2e9&t=252) in one of the tutorial videos where this is discussed (please excuse that I said "left" once where I should have said "right"!).

Do you have any suggestions for further changes to this?

saransh13:
@psavery We should uncheck the `display sum in subplots` by default. I like that option much better myself.

PR #1570, opened 2023-09-13
Default to not display sums in subplots in zoom

psavery:
For the Composite (Laue and Powder) calibration, default to not display sums in subplots. For this workflow, it is more helpful to not display the sums.

For the raw view zoom dialog, however, we are leaving it on for now.

Fixes: #1566


saransh13 (approved):
LGTM.

Issue #1567, opened 2023-08-29
Calibration point picker issue 2


akrygier:
When I want to cycle through the active reflection/detector click in the "Composite" calibration mode, sometimes I double click twice or accidentally double click, which causes the picker to go past the reflection/detector I would like. Is there a way to go backwards? If not can that be added?

psavery:
If you right-click, it moves on to the next reflection/detector, and there is currently not a way to go back to the previous reflection/detector. This is something we should add!

saransh13:
Yes, agreed. The process is tedious and prone to mistakes.

PR #1573, opened 2023-09-14
Allow going back a line while hand-picking points

psavery:
Previously, if the user accidentally right-clicked, we would move on
to the next line, and there would be no way to go back to the previous
line. This would mess up their hand-picking.
    
Now, the "Back" button is enabled, even when the current line has no
points, and clicking "Back" when the current line has no points results
in going back a line instead.

This PR also adds a couple of extra fixes where Laue spots were not
always being drawn correctly on the canvas while hand-picking
(they were not properly updated when the back button was pressed,
and they were not drawn on the detector after the first one).
    
Fixes: #1567


saransh13 (approved):
LGTM

PR #1574, opened 2023-09-14
Background refine

saransh13:
This is not compete yet. The background parameters are not initialized on first run but show up in subsequent runs. 
Requires PR https://github.com/HEXRD/hexrd/pull/558 to be merged.


psavery:
I believe this should work now.

psavery:
Waiting for packages from https://github.com/HEXRD/hexrd/pull/558 to build, and then we will merge!

psavery (approved):

Issue #1578, opened 2023-09-19
Add option to change multiprocessing settings


psavery:
This should probably go somewhere in the "Edit" menu.

PR #1580, opened 2023-09-23
Add a dialog where config settings can be changed

psavery:
This adds a dialog where the user can modify HEXRDGUI configuration settings.

Right now, the only option is to modify the number of processes used for multiprocessing.

The new menu item is under `Edit->Configuration`, and the dialog appears as the following:

image

We can likely add more configuration settings to this in the future, as needed.

Fixes: #1578

### DIFF
--- a/hexrd/ui/config_dialog.py
+++ b/hexrd/ui/config_dialog.py
@@ -1,0 +1,49 @@
+from hexrd.ui.hexrd_config import HexrdConfig
+from hexrd.ui.ui_loader import UiLoader
+
+
+class ConfigDialog:
+    def __init__(self, parent=None):
+        self.ui = UiLoader().load_file('config_dialog.ui', parent)
+
+        self.update_gui()
+        self.setup_connections()
+
+    def setup_connections(self):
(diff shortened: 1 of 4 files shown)


bnmajor (approved):
LGTM! My only question is if we want to remember the user's settings between sessions? It seems to me that if you need to limit resource usage for a machine you may want to limit it every time, but I also may be misunderstanding the use case here

psavery:
> LGTM! My only question is if we want to remember the user's settings between sessions? It seems to me that if you need to limit resource usage for a machine you may want to limit it every time, but I also may be misunderstanding the use case here

The user that requested this had mentioned that they start hexrdgui with a high number of CPUs on a shared machine, and then they want to reduce it later when they need to share the machine with others. So, they were specifically asking for a way to modify the number of CPUs in the middle of a session.

We can ask if we should remember this setting in between sessions. The potential problem, though, is that if a user sets it to something low in one session, they may forget about it in future sessions, and then they would not know why HEXRDGUI is performing slowly. Maybe a solution is to add a checkbox that says "Save as Default?" that will use it as the default. But we can ask the users first if this is something they want.

Issue #1575, opened 2023-09-15
Add ability to change the font size of the UI


psavery:
For high-resolution displays, the font size is too small, and cannot be easily read.

We should figure out a way to change the font size if needed.

PR #1582, opened 2023-10-02
Add an option to modify the font size of the app

psavery:
This option is present in the new "Edit"->"Configuration" dialog. This modifies the font size of both Qt and matplotlib.

When the user modifies the font size in the dialog, most (but not all) widgets in the application are updated immediately. It appears that any widgets with custom style sheets do not get updated. Perhaps some other types of widgets don't get updated either. But restarting the application normally appears to update everything to the new font size.

It would be nice to get all widgets to update immediately when the user modifies the font size. But for now, we can just present a warning that the user should restart the application normally in order to get all widgets to update.

Fixes: #1575

### DIFF
--- a/hexrd/ui/config_dialog.py
+++ b/hexrd/ui/config_dialog.py
@@ -1,3 +1,5 @@
+from PySide2.QtWidgets import QMessageBox
+
 from hexrd.ui.hexrd_config import HexrdConfig
 from hexrd.ui.ui_loader import UiLoader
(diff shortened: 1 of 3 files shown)


bnmajor (approved):
LGTM! Not a big deal but the default font size seems larger than it used to be - changing it to 9 seemed closer to what it was to me. But that may also be a platform specific thing

PR #1583, opened 2023-10-02
Remove tth max options from Fit Grains Options

psavery:
These options were actually not doing anything before. The reason is that the user is actually selecting these manually via the "Choose HKLs" button. That also provides the option to filter HKLs by their tth max value.

So the tth max options in the Fit Grains Options dialog were not doing anything, and their functionality is already available in the "Choose HKLs" button. Thus, they are removed.


psavery:
Here is what the dialog used to appear as:

prev_dialog

And here are the options that are being removed:

removed_options

psavery:
This also adds help buttons to various dialogs in "Fit Grains". See images below.

with_help

help2


help_button2

PR #1588, opened 2023-10-12
Mask boundaries

bnmajor:
WIP

### DIFF
--- a/hexrdgui/create_polar_mask.py
+++ b/hexrdgui/create_polar_mask.py
@@ -142,6 +142,7 @@ def create_polar_mask_from_raw(name, value):
     for det, data in value:
         line_data.extend(convert_raw_to_polar(det, data))
     create_polar_mask(name, line_data)
+    HexrdConfig().polar_bounds[name] = line_data
 
 
 def rebuild_polar_masks():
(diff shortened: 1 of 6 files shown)


bnmajor:
This is still a WIP but some questions I've come across so far:

1. My first pass separates toggling the boundary from toggling the mask which allows for any combination of mask and boundary visibility. Does this make sense or should these be exclusive modes (i.e. mask values, only show boundaries, hide all)?
2. Is it fair to assume the boundary option will only exist for polygon/rectangle/ellipse masks? Am I missing any?
3. Do we need to provide an option to set the styling for the boundary lines? For example if you're using the Greys colormap and masking out a dark feature the default black like is not very clear.

mask_bounds

@saransh13 @psavery Any thoughts?

saransh13:
> This is still a WIP but some questions I've come across so far:
> 
> 1. My first pass separates toggling the boundary from toggling the mask which allows for any combination of mask and boundary visibility. Does this make sense or should these be exclusive modes (i.e. mask values, only show boundaries, hide all)?
The way you have it set up is perfect. I think having both options works best.
> 2. Is it fair to assume the boundary option will only exist for polygon/rectangle/ellipse masks? Am I missing any?
Yep, those would be the only cases I can think of where this option is useful.
> 3. Do we need to provide an option to set the styling for the boundary lines? For example if you're using the Greys colormap and masking out a dark feature the default black like is not very clear.
Yes, I think that will be useful.
> 
> mask_bounds [ mask_bounds ](https://user-images.githubusercontent.com/51238406/274593884-b5f9fb3c-7673-42f6-94ac-fa0efc68c2c6.gif) [ ](https://user-images.githubusercontent.com/51238406/274593884-b5f9fb3c-7673-42f6-94ac-fa0efc68c2c6.gif)
> 
> @saransh13 @psavery Any thoughts?

bnmajor:
Closing in favor of #1653

Issue #1597, opened 2023-10-24
Error window size with Pyside6


saransh13:
During a fast powder calibration run, I encountered an error, which showed the an error message. The error message was long and the `ok` button was hidden. I could not drag the button to view and click it. I figured out that I can hit `return` to get back to the main window, but could be confusing to users.

<img width="1757" alt="Screenshot 2023-10-24 at 10 19 00 AM" src="https://github.com/HEXRD/hexrdgui/assets/15834451/5e9ecc96-a2e0-420f-8856-db9468629d04">

saransh13:
Looks like this is repeat of #1595. Closing this issue.

Issue #1595, opened 2023-10-24
Error window bigger than screen, can't be closed


akrygier:
I was doing a powder/laue calibration, using the automatic peak finder.  It failed due to too narrow 2 theta range (I think). When it failed an error window came up that I cannot close or ignore.

![image](https://github.com/HEXRD/hexrdgui/assets/59886900/35b436f2-630f-42fd-9ab0-78ce48ceb378)

akrygier:
This is the larger than normal image plate, if that matters...

saransh13:
I found the same problem. Hit `return` or `Enter` to close the window if you encounter this again. We'll work on a fix in the meantime.

psavery:
I think the easiest solution is to only put the final line of the error message in the message box, but keep the whole error message in the messages widget and terminal. I will do that.

PR #1599, opened 2023-10-24
Do not set detailed text in message box

psavery:
On Linux, the detailed text is hidden by default and can be shown via a button.

On Mac, however, the detailed text is always shown, and if it is too long, the buttons at the bottom of the dialog box will be unreachable on the user's monitor.

Don't put the detailed text in the message box. We already print it out in the messages widget, and that should be good enough. Only keep the abbreviated message in the message box.

Fixes: #1595 
Fixes: #1597


saransh13 (approved):
Generated error on purpose. The box appears smaller with a condensed message.

Issue #1598, opened 2023-10-24
Change label for line out


saransh13:
Change `Azimuthal Integration` to `Azimuthal average`

<img width="744" alt="Screenshot 2023-10-24 at 10 44 21 AM" src="https://github.com/HEXRD/hexrdgui/assets/15834451/eb8823fe-e3c7-40b2-930f-068b77eba699">

psavery:
So the y value is actually `Azimuthal Average`?

saransh13:
Yes, I think so. We divided by the number of non-NaN pixels in the azimuth,
correct?

psavery:
Yes that's right! See [here](https://github.com/HEXRD/hexrdgui/blob/05b53ce9d132b6d08c1141f7f75cc7ccca830fed/hexrdgui/image_canvas.py#L1126). I'll fix the label.

PR #1600, opened 2023-10-24
Change azimuthal label to "Azimuthal Average"

psavery:
We are indeed computing the azimuthal integration, but we are then dividing by the number of non-nan pixels, which results in an average.

Fixes: #1598


saransh13 (approved):
LGTM!

PR #1602, opened 2023-10-25
Change executable name in plist to match

psavery:
The name of the executable in the Info.plist file must exactly match the executable filename that is in `Contents/MacOS`. Otherwise, the code signing won't work.


bnmajor (approved):
:champagne: Thanks @psavery !

psavery:
I verified that the DMG still opens after making this change. So hopefully this change does not affect anything other than helping fix the codesigning.

Issue #1607, opened 2023-10-31
Increase width of display


saransh13:
The Image directory and filename are hard to read in the `Simple Image Series` dialog. Can we increase the width of the text field. We seem to have ample space.

<img width="742" alt="Screenshot 2023-10-31 at 10 27 03 AM" src="https://github.com/HEXRD/hexrdgui/assets/15834451/0afb487c-94c9-4afc-878c-329a01e11a0e">

PR #1609, opened 2023-11-02
Change info group to be a QGridLayout

psavery:
The QFormLayout that it was before was causing some resizing issues on Mac.

Fixes: #1607


saransh13 (approved):
LGTM.

Issue #1608, opened 2023-10-31
Export is still in 2 theta when X-axis type is Q


saransh13:
As the title says. I did `Export > Current View > *.xy` but the exported file had 2theta vs Intensity.

PR #1610, opened 2023-11-02
Export "Q" from the polar view

psavery:
This makes two changes:

1. For HDF5/npz formats, Q is always exported as "q_coordinates" (along with "tth_coordinates" and other arrays)
2. For all formats, the x-axis of the azimuthal integration is exported as "Q" if the polar x-axis type is set to "Q"

Fixes: #1608


saransh13 (approved):
LGTM!

PR #1612, opened 2023-11-09
Simplify powder overlay generation

psavery:
This skips swapping columns in the powder overlay generation so that we do not have to swap columns again when we are drawing. This simplifies the logic and speeds it up a little bit.


bnmajor (approved):
LGTM

PR #1613, opened 2023-11-09
Fix a few issues due to the flipped overlay data

psavery:
This fixes a couple of issues due to the flipped overlay data. Namely:

1. The sorting of HKL lines in the composite calibration
2. The masking by powder lines

I believe these should be the only two issues produced by #1612


bnmajor (approved):
Nice! Masking looks good now :+1:

PR #1614, opened 2023-11-10
Refactor overlay drawing to use fewer artists

psavery:
For all overlay types, we were using more artists than we needed to. For powder overlays, for instance, we would have one artist per line. For Laue and rotation series overlays, we would have one artist per spot and one artist per range! This could easily result in hundreds of artists.

However, if everything about the artists (including the style) are identical, except for the data, we are able to merge them together into a single artist. This can make matplotlib run much faster.

For line artists, we can have a single artist draw every powder overlay line - we just insert a `[nan, nan]` row in between lines in the data. Different artists are used for different styles (i. e., merged ranges are red instead of green, so they get a different artist, and highlighted data/ranges are a different color too, so they get a different artist). But we end up using only about 5 line artists in total for the main canvas, per overlay and per detector (each detector is currently rendered separately).

For scatter (path) artists, we ought to just pass every single spot to a single call to `scatter()`.

These changes speed up rendering of the overlays significantly, especially when there were many lines or spots being drawn. This is a great step toward better interactivity.

### DIFF
--- a/hexrdgui/create_raw_mask.py
+++ b/hexrdgui/create_raw_mask.py
@@ -47,9 +47,10 @@ def convert_polar_to_raw(line_data, reverse_tth_distortion=True):
         line_data[i] = line
 
     raw_line_data = []
+    instr = create_hedm_instrument()
     for line in line_data:
-        for key, panel in create_hedm_instrument().detectors.items():
-            raw = angles_to_pixels(line, panel)
+        for key, panel in instr.detectors.items():
+            raw = angles_to_pixels(line, panel, tvec_s=instr.tvec)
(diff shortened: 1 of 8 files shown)


bnmajor:
I am testing with the `tardis_simulated_powder_and_laue` state file from the examples repo and I am seeing weirdness with the masking:

powder_masking

Seems like everything between the lines is being masked now. Additionally the `Azimuthal Average` plot is also being masked it seems - I don't think that is the expected behavior though, am I correct?

psavery:
@bnmajor Can you give it a try again? It should be fixed now.

bnmajor (approved):
Awesome!

PR #1615, opened 2023-11-10
Fill nan lineout columns with zero

psavery:
Otherwise, they produce gaps in the lineout plot

This was only an issue if you had all nans in a column (such as when you apply a powder overlay mask in the polar view).


bnmajor (approved):
Awesome! LGTM :slightly_smiling_face:

PR #1616, opened 2023-11-13
Update "qt5agg" to "qtagg"

psavery:
In matplotlib, "qt5agg" is just a reference to "qtagg" now.

Let's update ours to "qtagg" so that it doesn't appear that we are using a qt5-specific import.


bnmajor (approved):
LGTM

Issue #1617, opened 2023-11-14
Add ability to copy selected reflections to clipboard


psavery:
In the reflections table, we should add a "right-click" context menu option where the users can copy highlighted rows to their clipboard. They will then be able to paste these as either plain text or in something like a CSV format (to more easily paste into Excel).

PR #1618, opened 2023-11-14
Add ability to copy reflections table to clipboard

psavery:
If a user right-clicks a row in the reflections table, a context menu appears with a single option: "Copy to Clipboard".

If the user clicks "Copy to Clipboard", the selected rows in the reflections table will be copied to the clipboard in CSV format. This may then be pasted anywhere (including to Excel).

An example:

image

The copied CSV text:
```csv
ID,{hkl},d-spacing (Å),2θ (°),|F|²,Iₚ,Multiplicity
0,1 1 1,3.12,3.17,100.00,100.00,8
1,2 0 0,2.71,3.66,44.98,25.29,6
2,2 2 0,1.91,5.18,52.19,29.30,12
3,3 1 1,1.63,6.08,19.52,15.93,24
4,2 2 2,1.56,6.35,10.37,2.59,8
```

Fixes: #1617

### DIFF
--- a/hexrdgui/reflections_table.py
+++ b/hexrdgui/reflections_table.py
@@ -1,10 +1,13 @@
 from contextlib import contextmanager
+import csv
+from io import StringIO
 import math
 
 import numpy as np
 
 from PySide6.QtCore import Qt, QItemSelectionModel
-from PySide6.QtWidgets import QTableWidgetItem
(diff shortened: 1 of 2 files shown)


bnmajor (approved):
Very cool!

PR #1622, opened 2023-11-16
First pass at new MaskManager class

bnmajor:


hexrdgui/mask_manager.py
```diff
@@ -0,0 +1,281 @@
...
+            h5py_group.create_group('masks')
+
+        self.write_all_masks(h5py_group['masks'])
+
+    def load_masks(self, h5py_group):
+        # TODO: Handle case of detector name mismatch (loading wrong mask file)
+        items = h5py_group.items()
+        if '_visible' in h5py_group.values():
+            # This is a file using the old format
+            items = load_old_mask_file(h5py_group)
```
psavery, line 256: I would say we should keep all of the logic as to whether we should convert an old format or not in the other compatibility file. Here, you can just do this instead:

```python
items = old_mask_compatibility(h5py_group)
```

That lets us keep all compatibility logic in that separate file (including checking the version). That `old_mask_compatibility()` function would be responsible for either converting an old format or just returning the items the way that they are if it is in the current format.

And then we should also start keeping a version in the h5py group instead of checking for `_visible`. So maybe have a `_version = 2` attribute. If the `_version` is not present, then assume it is version 1 (which means it would require converting the `_visible`). That will make future logic simpler if we have to change the mask format again.

hexrdgui/mask_compatability.py
```diff
@@ -0,0 +1,31 @@
+from hexrd.utils.compatibility import h5py_read_string
+
+
+def load_old_mask_file(self, h5py_group):
```
psavery, line 4: This can be changed to `load_masks_v1_to_v2` instead since it is converting masks from version 1 to version 2. Also, `self` shouldn't be an argument.

hexrdgui/mask_compatability.py
```diff
@@ -0,0 +1,31 @@
...
+            for name, masks in data.items():
+                items.setdefault(name, {
+                    'name': name,
+                    'mtype': 'unknown',
+                    'visible': name in visible,
+                    'border': False,
+                })
+                for mask in masks.values():
+                    # Load the numpy array from the hdf5 file
+                    items[name].setdefault(key, []).append(mask[()])
```
psavery, line 31: Need to return `items` at the end.

hexrdgui/mask_manager.py
```diff
@@ -0,0 +1,281 @@
...
+    def mask_image(self, status):
+        self._mask_image = status
+
+    @property
+    def show_border(self):
+        return self._show_border
+
+    @show_border.setter
+    def show_border(self, status):
+        self._show_border = status
```
psavery, line 63: These don't need to be properties. We can remove the properties for now (it's easy to change attributes to properties in the future if we need them).

hexrdgui/mask_manager.py
```diff
@@ -0,0 +1,281 @@
...
+        pass
+
+    @abstractmethod
+    def deserialize(self, data):
+        pass
+
+
+class RegionMask(Mask):
+    def __init__(self):
+        self._polar = None
```
psavery, line 93: We might want to consider just storing the coordinates in raw, especially if we might run into synchronization issues (such as the user changing polar view parameters), and just convert to polar on-the-fly. It's better to be bug-free than fast! But if we determine that the performance is an issue, we can think of things to do to speed it up.

hexrdgui/mask_manager.py
```diff
@@ -0,0 +1,281 @@
...
+    @abstractmethod
+    def get_data(self):
+        pass
+
+    @abstractmethod
+    def set_data(self, data):
+        pass
+
+    @abstractmethod
+    def get_mask_arrays(self):
```
psavery, line 75: What's the difference between `get_data` and `get_mask_arrays`?

hexrdgui/mask_manager.py
```diff
@@ -0,0 +1,281 @@
...
+            items = load_old_mask_file(h5py_group)
+        actual_view_mode = self.view_mode
+        self.view_mode = ViewType.raw
+        for key, data in items:
+            if data['mtype'] == 'threshold':
+                new_mask = ThresholdMask(None, None)
+                new_mask.deserialize(data)
+            else:
+                new_mask = RegionMask(None, None)
+                new_mask.deserialize(data)
```
psavery, line 265: It might be worth considering making `deserialize` a `classmethod` instead, so that you can just do this:

```python
new_mask = RegionMask.deserialize(data)
```

psavery (commented):
Overall, I think this is looking very nice! We need this refactoring improvement! Please see the suggestions, and keep up the good work! :+1:

hexrdgui/mask_compatability.py
psavery, line 1: Also, we might want to make a `masks` directory and put this inside, along with other mask-related files.

hexrdgui/mask_manager.py
bnmajor: Sounds good, updated

hexrdgui/mask_compatability.py
bnmajor: Done

hexrdgui/mask_compatability.py
bnmajor: Fixed

hexrdgui/mask_manager.py
bnmajor: Cleaned up

hexrdgui/mask_manager.py
bnmajor: `get_data` returns the actual point data describing the region to mask (needed for mask boundaries) and `get_mask_arrays` returns the list of masked images (one for each mask so that they can be quickly combined and re-combined if visibility is toggled).

hexrdgui/mask_manager.py
bnmajor: Sounds good! Just holding onto raw for now

hexrdgui/mask_manager.py
bnmajor: After some further simplification `get_mask_arrays` is now just an attribute: `masked_arrays`.

hexrdgui/main_window.py
```diff
@@ -749,21 +749,18 @@ def on_action_edit_apply_hand_drawn_mask_triggered(self):
...
-                name = unique_name(
-                    HexrdConfig().raw_mask_coords, 'polar_mask_0')
+                name = unique_name(MaskManager().mask_names, 'polar_mask_0')
                 raw_line = convert_polar_to_raw([line])
-                HexrdConfig().raw_mask_coords[name] = raw_line
-                HexrdConfig().visible_masks.append(name)
-                create_polar_mask_from_raw(name, raw_line)
-            HexrdConfig().polar_masks_changed.emit()
+                mask = MaskManager().add_mask(name, raw_line, MaskType.polygon)
+                mask.update_masked_arrays(self.image_mode)
```
psavery, line 755: Can you put `update_masked_arrays` inside `MaskManager().add_mask` so you don't need to call it after every `add_mask` call?

Or, alternatively, you could make a `get_masked_arrays()` function that takes the image mode as an argument, and updates lazily (when the data is requested). Something like this:

```python
def get_masked_arrays(self, image_mode):
    if self.masked_arrays is None or self.masked_arrays_image_mode != image_mode:
        self.update_masked_arrays(image_mode)

    return self.masked_arrays
```

I think it might be a good idea anyways for masks to keep track of which image mode their `masked_arrays` are currently in.

Lazily computing could also prevent repeated computations from happening (in case it were updated several times before the data were retrieved).

hexrdgui/masking/mask_manager.py
```diff
@@ -0,0 +1,390 @@
...
+from hexrdgui.constants import ViewType
+from hexrdgui.masking.constants import MaskType
+from hexrdgui.masking.create_polar_mask import (
+    create_polar_mask_from_raw, rebuild_polar_masks
+)
+from hexrdgui.masking.create_raw_mask import (
+    recompute_raw_threshold_mask, create_raw_mask, rebuild_raw_masks
+)
+from hexrdgui.hexrd_config import HexrdConfig
+from hexrdgui.masking.mask_compatability import load_masks_v1_to_v2
```
psavery, line 16: Spelling correction:

```suggestion
from hexrdgui.masking.mask_compatibility import load_masks_v1_to_v2
```

hexrdgui/masking/mask_manager.py
```diff
@@ -0,0 +1,390 @@
...
+
+class Mask(ABC):
+    def __init__(self, name='', mtype='', visible=True):
+        self.type = mtype
+        self.name = name
+        self.visible = visible
+        self.masked_arrays = None
+
+    def update_mask_visibility(self, visibility):
+        self.visible = visibility
```
psavery, line 33: This could alternatively be a property setter instead of a function.

hexrdgui/masking/mask_manager.py
```diff
@@ -0,0 +1,390 @@
...
+
+    # Abstract methods
+    @abstractmethod
+    def get_data(self):
+        pass
+
+    @abstractmethod
+    def set_data(self, data):
+        pass
```
psavery, line 42: `data` could be a property getter/setter as well, unless you anticipate more than one argument in the future.

hexrdgui/masking/mask_manager.py
```diff
@@ -0,0 +1,390 @@
...
+        if visible == self.visible:
+            return
+
+        self.visible = visible
+        if visible and self._hidden_mask_data:
+            self.set_data(self._hidden_mask_data)
+            self._hidden_mask_data = None
+        elif not visible:
+            self._hidden_mask_data = self.get_data()
+            self.set_data([-math.inf, math.inf])
```
psavery, line 114: I think we should always keep `self.min_val` and `self.max_val` to be whatever they were set as, regardless of the visibility. Elsewhere, functions like `recompute_raw_threshold_mask()` can just check if the mask is visible, and not apply it if it is not visible.

hexrdgui/masking/mask_manager.py
```diff
@@ -0,0 +1,390 @@
...
+    def write_masks_to_group(self, data, h5py_group):
+        unwrap_dict_to_h5(h5py_group, data, asattr=False)
+
+    def write_single_mask(self, name):
+        d = {'_version': 2}
+        d[name] = self.masks[name].serialize()
+        self.export_masks_to_file.emit(d)
+
+    def write_all_masks(self, h5py_group=None):
+        d = {'_version': 2}
```
psavery, line 267: We can save the current version as a constant somewhere.

hexrdgui/masking/mask_manager.py
```diff
@@ -0,0 +1,390 @@
...
+            self.export_masks_to_file.emit(d)
+
+    def save_state(self, h5py_group):
+        if 'masks' not in h5py_group:
+            h5py_group.create_group('masks')
+
+        self.write_all_masks(h5py_group['masks'])
+
+    def load_masks(self, h5py_group):
+        items = load_masks_v1_to_v2(h5py_group)
```
psavery, line 282: We can just call a `load_masks()` function here, and that function should figure out whether it should convert from `v1` to `v2` based upon the version number. (see previous comment about `load_masks()`.

hexrdgui/masking/mask_manager.py
```diff
@@ -0,0 +1,390 @@
...
+            self.masked_arrays = create_polar_mask_from_raw(self._raw)
+
+    def serialize(self):
+        data = {
+            'name': self.name,
+            'mtype': self.type,
+            'visible': self.visible,
+        }
+        for i, (det, values) in enumerate(self._raw):
+            data.setdefault(det, {})[str(i)] = values
```
psavery, line 82: Can we put this under a `data` key instead, so it looks like this:

```python
data = {
    'name': self.name,
    'mtype': self.type,
    'visible': self.visible,
    'data': {
        'detector_1': ...,
        'detector_2': ...,
    },
}
```

Although it isn't likely to happen, this will help us ensure that the user's detector names will not clash with the keywords `visible`, `mtype`, etc. That would be a difficult-to-track bug if it were to occur.

psavery (commented):
Looking really nice so far! We really needed this re-organization! I left a few comments about ways we could improve things further.

hexrdgui/calibration/auto/powder_runner.py
```diff
@@ -66,7 +66,7 @@ def _run(self):
             fwhm_estimate = options['initial_fwhm']
 
         # Get an intensity-corrected masked dict of the images
-        img_dict = HexrdConfig().masked_images_dict
+        img_dict = MaskManager().masked_images_dict
```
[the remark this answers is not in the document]
bnmajor: Okay, that makes sense! I've moved the `raw_masks_dict` and `masked_images_dict` properties as well as the `create_masked_images_dict` function back to `HexrdConfig`.

hexrdgui/masking/mask_compatability.py
```diff
@@ -0,0 +1,36 @@
+from hexrd.utils.compatibility import h5py_read_string
+
+
+def load_masks_v1_to_v2(h5py_group):
```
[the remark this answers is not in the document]
bnmajor: Got it :+1:

hexrdgui/calibration/polarview.py
```diff
@@ -410,15 +411,16 @@ def apply_snip(self, img):
 
     def apply_masks(self, img):
         # Apply user-specified masks if they are present
+        from hexrdgui.masking.mask_manager import MaskManager
```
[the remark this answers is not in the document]
bnmajor: This is in order to avoid a circular import

hexrdgui/masking/constants.py
```diff
@@ -0,0 +1,7 @@
+class MaskType:
+    region = 'region'    # Rectangle or Ellipse mask
+    polygon = 'polygon'  # Hand drawn mask
+    threshold = 'threshold'
+    powder = 'powder'
+    laue = 'laue'
+    pinhole = 'pinhole'
```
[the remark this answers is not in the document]
bnmajor: We need to know the type of mask for the mask boundaries feature. I chose to be explicit about each type in case we have need for these details in future feature requests, but if we want to keep things simpler we could just have something like `threshold`, `computed` (`powder`, `laue`, `pinhole`) and `drawn` (`region`, `polygon`)? I'm open to better names of course (`threshold` could arguably be included in `computed` as well).

hexrdgui/masking/mask_manager_dialog.py
```diff
@@ -0,0 +1,218 @@
...
+from hexrdgui.ui_loader import UiLoader
+from hexrdgui.utils.dialog import add_help_url
+
+import matplotlib.pyplot as plt
+
+
+class MaskManagerDialog(QObject):
+
+    # Emitted when masks are removed or visibility is toggled
+    update_masks = Signal()
```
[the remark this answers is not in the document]
bnmajor: This is actually no longer even used so I've just removed it.

hexrdgui/masking/mask_manager.py
bnmajor: No, I can't. Updated to be property getter/setter instead

hexrdgui/masking/mask_manager_dialog.py
```diff
@@ -0,0 +1,218 @@
...
+        self.ui.export_masks.clicked.connect(MaskManager().write_all_masks)
+        self.ui.import_masks.clicked.connect(self.import_masks)
+        self.ui.panel_buffer.clicked.connect(self.masks_to_panel_buffer)
+        self.ui.view_masks.clicked.connect(self.show_masks)
+        self.ui.hide_all_masks.clicked.connect(self.hide_all_masks)
+        self.ui.show_all_masks.clicked.connect(self.show_all_masks)
+        MaskManager().mask_mgr_dialog_update.connect(self.setup_table)
+        MaskManager().export_masks_to_file.connect(self.export_masks_to_file)
+
+    def setup_table(self):
```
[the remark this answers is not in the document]
bnmajor: Sure thing!

hexrdgui/masking/mask_manager_dialog.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def get_old_name(self, row, column):
+        if column != 0:
+            return
+
+        self.old_name = self.ui.masks_table.item(row, 0).text()
+
+    def update_mask_name(self, row):
+        if not hasattr(self, 'old_name') or self.old_name is None:
+            return
```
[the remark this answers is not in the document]
bnmajor: Yup, that can be changed now.

hexrdgui/masking/constants.py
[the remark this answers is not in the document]
psavery: We can keep these for now! Thanks Brianna.

hexrdgui/masking/mask_manager_dialog.py
```diff
@@ -0,0 +1,218 @@
...
+        fig.canvas.draw_idle()
+        fig.show()
+
+    def update_visibility_checkboxes(self):
+        with block_signals(self.ui.masks_table):
+            for i, key in enumerate(MaskManager().mask_names):
+                cb = self.ui.masks_table.cellWidget(i, 1)
+                status = key in MaskManager().visible_masks
+                cb.setChecked(status)
+        MaskManager().masks_changed()
```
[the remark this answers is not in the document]
bnmajor: Yes, the visibility of masks are changed here. The UI is first updated to reflect the changes and then the function is called so that the appropriate signal is emitted and the images are updated. This function is only called from `hide_all_masks` and `show_all_masks` so the logic here is what was redundant between the two. If you'd prefer though I can move `MaskManager().masks_changed` into `hide_all_masks` and `show_all_masks` instead.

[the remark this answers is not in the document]
psavery: Yeah, can you put it in those functions instead? I think clarity is better in this case than redundancy...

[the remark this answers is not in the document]
bnmajor: Sounds good :+1:

hexrdgui/main_window.py
psavery: I do think it's a better approach to perform the update lazily. Lazy updates might save us some computation time (since we might sometimes currently be calling `update_masked_arrays()` when we don't need to, but if we call that function right before retrieving the masked arrays, we are guaranteed to perform it only when we need it).

And then, with the lazy updates, we would need some way to invalidate the `masked_arrays`, probably just by setting it to `None` (or to make it clearer, have a function called `invalidate_masked_arrays()` that does just that). This would force it to be re-computed the next time it is retrieved. And we would do things like that when we, for instance, change the polar pixel resolution size.

I don't think the `MaskManager().view_mode != image_mode` approach would work because the individual masks aren't keeping track of which image mode was used to create its `masked_array`. The mask manager's view mode might not match the view mode that was used to create the mask's `masked_array`. I think those could get out of sync easily.

So I think it would be nice to call a `get_masked_arrays(image_mode)` function that basically checks if its masked array was generated already using that image mode, and just return it if so, or re-generate it if the image mode doesn't match. But if this is too much work for this PR, we can wait on it until sometime in the future...

bnmajor: My original question was mostly because I was misunderstanding what you meant at first - you weren't asking to sync the mode on each mask every time we switch image modes, just need to track the mode we were in when we last generated the masks. I think what you've explained makes total sense and I do not think that it's too much for this PR!

hexrdgui/masking/mask_manager.py
```diff
@@ -28,6 +28,16 @@ def __init__(self, name='', mtype='', visible=True):
         self.name = name
         self.visible = visible
         self.masked_arrays = None
+        self.last_view_mode = ViewType.raw
```
psavery, line 31: This might be better named `masked_arrays_view_mode`, since it is being used specifically to keep track of the current view mode of the `masked_arrays`.

hexrdgui/masking/mask_manager.py
```diff
@@ -28,6 +28,16 @@ def __init__(self, name='', mtype='', visible=True):
...
+
+    def get_masked_arrays(self, image_mode=ViewType.raw):
+        if self.masked_arrays is None or self.last_view_mode != image_mode:
+            self.update_masked_arrays(image_mode)
```
psavery, line 35: Nice!

hexrdgui/masking/mask_manager.py
```diff
@@ -113,6 +124,7 @@ def data(self, values):
         self.update_masked_arrays()
 
     def update_masked_arrays(self, view=ViewType.raw):
+        self.last_view_mode = view
```
psavery, line 127: Since the threshold masks don't actually use the view mode, maybe `get_masked_arrays` should be overloaded for it, and skip the view mode check?

hexrdgui/masking/mask_manager.py
```diff
@@ -249,7 +261,7 @@ def load_masks(self, h5py_group):
             else:
                 new_mask = RegionMask.deserialize(data)
             self.masks[key] = new_mask
-            new_mask.update_masked_arrays()
+            new_mask.update_masked_arrays(self.view_mode)
```
psavery, line 264: We probably don't need this line of code, right? The mask will get generated when the `get_` method is called.

psavery:
Looks like there are still a bunch of calls for `mask.update_masked_arrays()` in places. I think we should be able to remove them all (in some places we need to replace them with `invalidate_masked_arrays()`), since we now only compute the masked arrays right before they are needed.

hexrdgui/masking/mask_manager.py
bnmajor: Got it, changed :+1:

hexrdgui/masking/mask_manager.py
bnmajor: Sounds good!

hexrdgui/masking/mask_manager.py
bnmajor: Yes, there were quite a few redundant `update_` calls... Those have been removed now and in relevant places the they have been replaced with `invalidate_` calls.

psavery (approved):
I made a few small fixes, and tested nearly everything. It's working well! Great job, @bnmajor! Please look over my changes, and if they look good, then this is ready to merge!

PR #1631, opened 2023-12-01
Animate overlay artists through blitting

psavery:
This adds a blit manager to manage overlay artists

The blit manager was modified from this example: https://matplotlib.org/stable/users/explain/animations/blitting.html

It is being used to automatically manage things like updating the background image, updating the blit artists, etc.

We are now using it to manage the overlay artists, including drawing them and removing them.

Blitting the overlay artists provides a *significant* performance boost when we modify parameters that only require the overlay artists to be redrawn. Especially when we have a high resolution in any of the views (other than raw, which doesn't have a resolution).

For example, the sliders for the Laue parameters are significantly more interactive now. And this provides a great opportunity for us to boost performance for the pressure sliders too.

### DIFF
--- a/hexrdgui/blit_manager.py
+++ b/hexrdgui/blit_manager.py
@@ -1,0 +1,103 @@
+from collections.abc import Sequence, ValuesView
+
+from matplotlib.artist import Artist
+
+
+class BlitManager:
+    def __init__(self, canvas):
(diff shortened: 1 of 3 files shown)


bnmajor (approved):
This is fantastic!!

PR #1634, opened 2023-12-06
Make the default working dir the current directory

psavery:
This fixes an issue where in many places in the code, we assume the working_dir to be a valid path. But if it is `None`, an exception is raised.


bnmajor (approved):
Awesome!

Issue #1446, opened 2023-05-16
Look into adding ROI support to the GUI similar to how you can use it in HEXRD


psavery:
This goes along with #1347

We will hopefully get an example soon.

psavery:
More info:

HEXRD supports an ROI parameter in the detector parameters of the instrument config. This defines a "region of interest", so that the detector only uses a region of pixels from the image, like so:

```yaml
detectors:
  ff1_0_0:
    pixels:
      roi:
      - 0
      - 0
  # ...
  ff1_0_1:
    pixels:
      roi:
      - 0
      - 1536
```

(those are start pixels for row/column, and then the number of rows/columns in the detector settings is used to determine where to stop)

Then in the script, multiple detectors are specified for a single image file, like so:

```yaml
image_series:
  format: frame-cache
  data:
    - file: mruby-1104-1_7_ff1_000011-cachefile.npz
      panel:  # must match detector keys
        - ff1_0_0
        - ff1_0_1
        # ...
```

The question now is: should we add support for this in the GUI?

My opinion so far: I think we need to discuss the pros/cons and alternative solutions before going down this route. Having ROI support in the command line interface makes sense and keeps it flexible (although it may better belong in the image series options rather than in the detector options). But adding this to the GUI exactly as it is may be complicated, and it may be somewhat non-intuitive.

In my limited testing, it looks like there are many places in the GUI where we would have to add logic to handle it, and thus could be a fair amount of work to add. It would also add an extra degree of maintenance (if we make any changes, we'd have to ensure it doesn't break ROI support), which could make maintenance more difficult over time. But we can certainly still do it anyways if it is important.

If we could alternatively add an option in the GUI to break apart detectors into multiple sub-detectors (which would also break up the image files into separate image files), that may help alleviate the issue. Perhaps also we could allow instrument configs with ROI to be loaded in, then immediately have the user select image files, and then immediately break up the image files into separate files and eliminate the ROI section in the config. That would be easy to add, and easy to maintain. Maybe this would only pose a problem if the file sizes were very large.

So to summarize: we could add ROI support (where an `roi` parameter is maintained in the instrument config), but it may be a fair amount of work, and I think there are some good alternatives that may be easier to support and maintain (which could include loading in an instrument file with an `roi` parameter and then converting it to something easier to work with). What do you think, @donald-e-boyce?

PR #1636, opened 2023-12-12
Roi

bnmajor:
Relies on HEXRD/hexrd#591

Fixes: #1446

### DIFF
--- a/hexrdgui/hexrd_config.py
+++ b/hexrdgui/hexrd_config.py
@@ -1555,8 +1555,13 @@ class HexrdConfig(QObject, metaclass=QSingleton):
     def is_roi_instrument_config(self):
         for det in HexrdConfig().detectors.values():
             if det.get('pixels', {}).get('roi', {}).get('value'):
-                 return True
+                return True
         return False
+
+    def detector_roi_group(self, detector_name):
+        if self.is_roi_instrument_config:
+            det = self.detector(detector_name)
+            return det.get('group', {}).get('value')
 
     def detector_pixel_size(self, detector_name):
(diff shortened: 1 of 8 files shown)


hexrdgui/hexrd_config.py
```diff
@@ -1555,9 +1555,14 @@ def default_detector(self):
...
         for det in HexrdConfig().detectors.values():
             if det.get('pixels', {}).get('roi', {}).get('value'):
-                 return True
+                return True
         return False
 
+    def detector_roi_group(self, detector_name):
+        if self.is_roi_instrument_config:
+            det = self.detector(detector_name)
+            return det.get('group', {}).get('value')
```
psavery, line 1564: I think a detector "group" might be applicable in other settings other than ROI. For example, for a chunked instrument (which won't have ROI), we might still want to constrain distances between detectors in the same group for calibration. 

So we might want to just have `detector_group()` and remove the `self.is_roi_instrument_config` check.

bnmajor: Makes sense!

bnmajor: Just pushed the fix for the calls that I missed - should be all fixed now.

> and in this places where the group name is checked in the file name, we should only do that if it is roi as well.

So for something like the chunked instrument example you mentioned above it may have a group key but we'd expect that we wouldn't have fewer images than detectors?

psavery: Right! It would have a group key, but the number of images will match the number of detectors.

bnmajor: Got it!

PR #1637, opened 2023-12-12
Prevent image series slider size adjustment

psavery:
It is non-intuitive that the image series slider would change size while you are sliding it, but that would happen because the omega text label would change widths.

This fixes the issue.


bnmajor (approved):
Love this!

PR #1639, opened 2023-12-15
Add constant chi overlays

psavery:
These overlays help the user to visualize constant chi values.

The material associated with a constant chi overlay is used to generate the suggested chi values (after the user selects a fiber).

The chi value information also now appears on mouse hover.

To do:

- [x] I need to verify that the tree view editor in a few different places is not broken by any of these changes.


saransh13 (approved):
I have tested with this the TARDIS instrument. The results are as expected. All options in the feature was working and it does not break any other feature in the GUI.

Issue #593, opened 2020-09-25
Panel buffer config not working


joelvbernier:
I am trying to apply a npy file as a buffer (couldn't use simple threshold to do it), and first it complains about a size mismatch (there isn't one).  Then, when I try to reopen the 'Configure Panel Buffer' dialog after canceling it, it get the following `KeyError`:
```
Traceback (most recent call last):
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/tree_views/value_column_delegate.py", line 44, in _clicked
    dialog = PanelBufferDialog(detector, self)
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/calibration/panel_buffer_dialog.py", line 34, in __init__
    self.update_gui()
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/calibration/panel_buffer_dialog.py", line 131, in update_gui
    buffer = detector_config['buffer']['value']
KeyError: 'value'
```
It also hoses something in the config; attempting to restart `hexrdgui` produces:
```
Traceback (most recent call last):
  File "/Users/joel/opt/miniconda3/envs/hexrdgui/bin/hexrdgui", line 33, in <module>
    sys.exit(load_entry_point('hexrdgui', 'gui_scripts', 'hexrdgui')())
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/main.py", line 29, in main
    window = MainWindow()
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/main_window.py", line 121, in __init__
    ImageFileManager().load_dummy_images(True)
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/image_file_manager.py", line 42, in load_dummy_images
    iconfig = HexrdConfig().instrument_config
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/hexrd_config.py", line 267, in instrument_config
    return self.filter_instrument_config(
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/hexrd_config.py", line 478, in filter_instrument_config
    self.remove_status(default['instrument'])
  File "/Users/joel/Documents/GitHub/hexrdgui/hexrd/ui/hexrd_config.py", line 519, in remove_status
    self.remove_status(value, current, key)
[...]

PR #1640, opened 2023-12-19
Add dialog for pressure and temperature calculations

psavery:
This adds a dialog where lattice parameters of a material can be computed using different temperatures and pressures. A button to make it appear is now located in the "Properties" tab of the materials panel.

This includes the option to load in PT parameters from a JCPDS file. 

image

Depends on: #593


saransh13 (approved):
Tested and ready to go.

PR #1646, opened 2024-01-05
Azimuthal offset

### LINKED ISSUE #1563: Ability to add a constant background to the azimuthal overlay
Sometimes the baseline of the azimuthal integration is above zero and using the azimuthal overlay doesn't look very nice. Can we add an option to add a constant offset to the simulated values so the entire plot shifts up (or down if the offset is negative)?

bnmajor:
Fixes #1563

### DIFF
--- a/hexrdgui/azimuthal_overlay_editor.py
+++ b/hexrdgui/azimuthal_overlay_editor.py
@@ -37,12 +37,12 @@ class AzimuthalOverlayEditor:
 
     def update_fwhm(self, value):
         self.selected_overlay['fwhm'] = value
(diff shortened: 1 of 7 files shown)


psavery (approved):
After changing `azimuthal_plot_modified` to `azimuthal_options_modified`, this looks great!

psavery:
@saransh13 Do you want to give this a try before we merge it?

saransh13:
I was initially thinking of shifting the overlays, instead of the actual data. But this was works too, and I believe is more flexible and useful. LGTM!

PR #1650, opened 2024-01-10
Add an option to load pre-configured instrument files

### LINKED ISSUE #1586: Add a set of pre-configured instruments that can be loaded
We should add somewhere in HEXRDGUI the option to load pre-configured instruments (such as "Dual Dexelas").

We can ship HEXRD with a set of these pre-configured instruments, and allow the user to save their own pre-configured instruments to this list. So the users may need a list editor (look at the [materials list editor](https://github.com/HEXRD/hexrdgui/blob/master/hexrd/ui/material_list_editor.py) for inspiration).

bnmajor:
Also updates the naming from "Configuration" to "Instrument" in the "Open" and "Save" menus.

Relies on HEXRD/hexrd#602.

Fixes #1586

### DIFF
--- a/hexrdgui/main_window.py
+++ b/hexrdgui/main_window.py
@@ -1590,12 +1590,20 @@ class MainWindow(QObject):
         self.load_state_file(path)
 
     def on_action_open_preconfigured_instrument_file_triggered(self):
+        # Should we put this in HEXRD?
+        aliases = {
+            'dual_dexelas.yml': 'Dual Dexelas',
+        }
+
         # Create a dict of options for loading an instrument, mapping file
         # name to instrument config
         options = {}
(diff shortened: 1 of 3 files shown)


bnmajor:
@psavery Should we immediately support saving files to the pre-configured list? Or start like detector templates and limit it to options in the HEXRD repo for now?

psavery:
> @psavery Should we immediately support saving files to the pre-configured list? Or start like detector templates and limit it to options in the HEXRD repo for now?

I think what we have now is fine to start with!

psavery (approved):
I added some aliases. I think this is working great! Thanks @bnmajor :slightly_smiling_face:

Issue #1627, opened 2023-11-28
Add option to stitch ROI images together in raw view


psavery:
When you have an ROI detector setup, there should be a checkbox in the raw view as to whether the ROI subpanel images should be stitched together or not.

For example, for dexelas, where there are 4 subpanels per detector, and two detectors, there would by default be 8 images shown in the raw view. Checking this box will stitch the subpanel images together (perhaps just taking them from the original imageseries exactly as they are), so there will only be two images in the raw view.

PR #1655, opened 2024-01-25
Add option to stitch ROI images in raw view

psavery:
If the instrument config contains both the `roi` and `group` key in each detector, an additional checkbox now appears in the raw view settings that allows users to stitch the ROI images together.

Each subpanel is still treated separately. Overlays are also generated separately for each subpanel. However, the final images are stitched together into one large panel.

Fixes: #1627


bnmajor (approved):
Looks great!!

Issue #1628, opened 2023-11-28
For ROI, add option to lock subpanel transforms in the slider view


psavery:
For ROI image setup, we should add a checkbox to the slider view so that the translation/tilt differences between subpanels are essentially "locked". If checked, and if the user moves the slider for one subpanel, all of the other subpanels will move too.

PR #1658, opened 2024-02-10
Add option to lock ROI group detector transforms

psavery:
If an ROI instrument config is loaded, a new checkbox appears in the 
instrument slider view with the label "Lock ROI Group Transformations".

If checked, then when a detector is transformed (via translation or tilt),
all other detectors in the same group are transformed similarly.

For translation, all other detectors in the same group are translated by
the same amount, and in the same direction.

For tilt, all detectors in the group are rotated about the mean of their
centers by the tilt amount (which means their translation is affected too).

See below video as an example.

https://github.com/HEXRD/hexrdgui/assets/9558430/b90e2774-c5e9-49da-9427-3119f243fcfe

Fixes: #1628

### DIFF
--- a/hexrdgui/blit_manager.py
+++ b/hexrdgui/blit_manager.py
@@ -43,6 +43,10 @@ class BlitManager:
                 )
                 raise RuntimeError(msg)
 
+        if not cv.figure.axes:
+            # Ignore the request, as there are currently no axes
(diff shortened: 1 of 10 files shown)


bnmajor (approved):
Very nice!

Issue #1664, opened 2024-03-01
Add HEDM intensity column to reflections table


psavery:
The reflections table currently has intensity information for powder via Ip (powder intensity), however, there is no HEDM equivalent to work with.  Adding a column which combines structure factor, polarization factor, and Lorentz factor together (specifically (|F|^2)*(polarization)*(Lorentz)).  This is already taken care of in the powder intensity calculation (along with other factors).  Such a value would also be a more useful value when filtering HKLs in fitgrains.  Thoughts?

_Originally posted by @gustafsonsven in https://github.com/HEXRD/hexrdgui/discussions/1662_

Yes! Let's add another column. Powder intensity is currently computed as follows:

```python
structure_factor*multiplicity*lorentz*polarization
```

But we will create a new column for HEDM that is computed as follows:

```python
structure_factor*lorentz*polarization
```

PR #1665, opened 2024-03-04
Add reflections table column for HEDM intensity

psavery:
This is computed nearly the same way as the powder intensity, except that it does not include multiplicity. It is particularly useful for rotation series data.

It is `Iₛ` in this table:

image

Fixes: #1664
Depends on: hexrd/hexrd#619


bnmajor (approved):
Other than the failing test this looks good to me!

psavery:
Just needed to re-run with updated HEXRD :slightly_smiling_face:

PR #1669, opened 2024-03-22
Only setup signals if masking is in progress

### LINKED ISSUE #1651: Masking issue
The point selector for `polygon` mask is still on while I'm in the `ellipse/rectangular` masking option.
Applied ellipse/rectangular masks
Applied polygon masks
The went back to apply ellipse/rectangular masks.

Screenshot of the test case
Screenshot 2024-01-12 at 12 33 50 PM

bnmajor:
Fixes #1651


psavery (approved):
I didn't try it, but the changes look good to me.

saransh13 (approved):
Can no longer see the bug with this change. LGTM

PR #1674, opened 2024-03-26
Use masked images for color map settings

psavery:
For Eiger, invalid pixels have a value of 4294967295. This really messes up the color map percentile defaults.

Instead of using the raw data for creating the default color map settings, we should use masked data instead, so that these invalid pixels are not included in the percentile setup.


psavery:
@saransh13 Is this change okay with you?

bnmajor (approved):
LGTM

PR #1676, opened 2024-03-26
Add an option to color invalid pixels

psavery:
If checked, pixels that are `nan` are displayed with a specified color.

First, a QColorDialog appears asking the user to select a color. After a color is selected, `nan` pixels will be displayed with this color.


psavery:
Here's an example:

eiger_invalid

bnmajor (approved):
Nice!

PR #1679, opened 2024-03-28
Performance improvements to raw view loading

psavery:
In order to make scrolling through images in the raw view faster, this PR provides several speed improvements to raw view loading. On my computer, it is now about 3.2x faster if there are no display masks and no overlays. Here are the performance improvements:

1. The images dict was accidentally being created twice, which caused a performance hit. It is now only being created once.
2. If there are no masks with a border-only setting, we avoid creating separate dictionaries for display images and computed images (we can re-use the display images *as* the computed images).
3. If there are no overlays, we don't update the blit manager, which can be time-consuming.

### DIFF
--- a/hexrdgui/hexrd_config.py
+++ b/hexrdgui/hexrd_config.py
@@ -888,13 +888,13 @@ class HexrdConfig(QObject, metaclass=QSingleton):
 
     @property
     def raw_masks_dict(self):
-        return self.create_raw_masks_dict(display=False)
+        return self.create_raw_masks_dict(self.images_dict, display=False)
 
-    def create_raw_masks_dict(self, display=False):
+    def create_raw_masks_dict(self, images_dict, display=False):
         """Get a masks dict"""
         from hexrdgui.masking.mask_manager import MaskManager
         masks_dict = {}
-        for name, img in self.images_dict.items():
+        for name, img in images_dict.items():
             final_mask = np.ones(img.shape, dtype=bool)
             for mask in MaskManager().masks.values():
                 if display and not mask.visible:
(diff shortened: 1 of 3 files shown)


hexrdgui/hexrd_config.py
```diff
@@ -940,22 +940,23 @@ def create_masked_images_dict(self, fill_value=0, display=False):
...
-        raw_masks_dict = self.create_raw_masks_dict(display=display)
+        raw_masks_dict = self.create_raw_masks_dict(images_dict,
+                                                    display=display)
         for det, mask in raw_masks_dict.items():
+            img = images_dict[det]
             if has_panel_buffers:
                 panel = instr.detectors[det]
                 utils.convert_panel_buffer_to_2d_array(panel)
 
-            for name, img in images_dict.items():
```
psavery, line 949: @bnmajor I got rid of this extra loop, because it appears to be looping over detector names as well (same as the outer loop). Please let me know if I was incorrect to do so.

hexrdgui/hexrd_config.py
```diff
@@ -888,13 +888,13 @@ def images_dict(self):
...
     def raw_masks_dict(self):
-        return self.create_raw_masks_dict(display=False)
+        return self.create_raw_masks_dict(self.images_dict, display=False)
 
-    def create_raw_masks_dict(self, display=False):
+    def create_raw_masks_dict(self, images_dict, display=False):
         """Get a masks dict"""
         from hexrdgui.masking.mask_manager import MaskManager
         masks_dict = {}
-        for name, img in self.images_dict.items():
```
psavery, line 897: We are avoiding calling `self.images_dict` here because it is a computed property and takes some time to generate. Instead, the already-created `images_dict` is passed in as a function argument.

hexrdgui/hexrd_config.py
bnmajor: I think that you are correct, this was redundant code that should have been removed in a previous refactor of this function... LGTM!

bnmajor (approved):
This seems to be working well for me! Awesome!

Issue #1682, opened 2024-03-29
Speed up toggling masks


bnmajor:
After the Mask Manager refactor there appears to be a slow down in updating the image when masks are toggled on/off. Need to investigate and improve speeds.

PR #1683, opened 2024-03-29
Performance improvements for toggling mask display

psavery:
On one example, it was taking ~6 seconds to change a single mask display type. Now it takes ~0.37 seconds for that same example.

Fixes: #1682


saransh13 (approved):
Really impressive improvement in masking performance! Tested it on some usual cases (PXRDIP and TARDIS). Ready to merge.

PR #1686, opened 2024-04-09
Add and integrate generic calibration dialog

psavery:
A generic calibration dialog was added that re-uses the same dialog
for various calibration workflows. This allows users to utilize
the new lmfit parameters (from hexrd/hexrd#628) as well as more
generally control the calibration procedure (including "undo", or
changing refinable parameters and running again), after points have
been picked.

The "Fast Powder" and "Composite (Laue and Powder)" have been updated
to utilize this generic calibration dialog.

Depends on: hexrd/hexrd#628


saransh13 (approved):
Passed all tests. LGTM

Issue #1432, opened 2023-05-03
Polar lineout powder lines are repeated


psavery:
Right now, the mean (for x) of each detector's powder line is drawn in the lineout. This isn't usually noticeable because most of the time, each detector has the same mean value for their powder line. But if the overlays are distorted, and some masking or panel buffers are applied, you can end up with different means per detector, which results in extra powder lines in the lineout. Instead, we should concatenate points across all detectors, take the mean, and draw that once in the lineout.

PR #1687, opened 2024-04-11
Aggregate HKL powder lines in the polar lineout

psavery:
Before this PR, each detector would draw the mean of its own powder line data in the polar view azimuthal lineout. This isn't noticeable when the powder lines are completely straight. However, when the powder lines are not completely straight, the duplicate lines would be visible as separate means.

To avoid this, concatenate all lines that belong to the same HKL from each detector, and take their mean together. Use this mean to draw a single line in the azimuthal lineout. Repeat this kind of thing for the ranges, too.

# Before
before

# After
after

Fixes: #1432


saransh13 (approved):
changes look good.

PR #1688, opened 2024-04-11
Improve performance of generating polar masks

psavery:
For a large number of polar masks, `create_hedm_instrument()` was taking a lot of cumulative time, since it was being called repeatedly. Provide the option to pass it in as an argument for performance.

Also, an intensity-corrected images dict was being repeatedly created. Now, we skip that step if the PolarView is initialized with an instrument that is `None`.

This resulted in a ~50% performance improvement for generating the polar view with an example that had 42 masks.

### DIFF
--- a/hexrdgui/calibration/polarview.py
+++ b/hexrdgui/calibration/polarview.py
@@ -48,12 +48,16 @@ class PolarView:
 
         self.distortion_instr = distortion_instrument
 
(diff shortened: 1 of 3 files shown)


saransh13 (approved):
Noticeable performance bump for lower resolution dewdrops. LGTM

PR #1692, opened 2024-04-19
Add pinhole radius to sample layer distortion

psavery:
This is now an option for sample layer distortion. It is used to compute the critical beta angle, and invalidate points outside of that angle.

Depends on: hexrd/hexrd#635


saransh13 (approved):
LGTM

Issue #1690, opened 2024-04-18
Intensities not syncing after B&C adjustment


saransh13:
Steps:
1. Started the structureless calibration in logs-log-sqrt intensity scaling
2. Tried to edit points in two-click mode
3. Intensity in the zoom-in box was using scaling None instead of the log-log-sqrt

Example image below

![Screenshot 2024-04-18 at 11 28 03 AM](https://github.com/HEXRD/hexrdgui/assets/15834451/318dd72b-0705-4fcb-a9ac-8ebbabcf58a7)

psavery:
This should be fixed in #1693

PR #1693, opened 2024-04-19
Masking fixes

bnmajor:
- Make sure cartesian coordinates are clipped before converting to/from raw coords
- Apply tth distortion (if needed) to mask coords when converted from raw to polar

Fixes: #1690

### DIFF
--- a/hexrdgui/calibration/polarview.py
+++ b/hexrdgui/calibration/polarview.py
@@ -407,7 +407,7 @@ class PolarView:
 
         # Both "visible" and "boundary" masks are applied to the
         # computational image
-        comp_img = self.apply_boundary_masks(img)
+        comp_img = self.apply_boundary_masks(disp_img)
         self.computation_img = comp_img
 
     def apply_snip(self, img):
@@ -489,7 +489,7 @@ class PolarView:
 
         # Both "visible" and "boundary" masks are applied to the
         # computational image
-        comp_img = self.apply_boundary_masks(img)
+        comp_img = self.apply_boundary_masks(disp_img)
         self.computation_img = comp_img
 
     @property
--- a/hexrdgui/calibration/polarview.py
+++ b/hexrdgui/calibration/polarview.py
@@ -401,14 +401,13 @@ class PolarView:
         # masks should also be distorted as well.
 
         # We only apply "visible" masks to the display image
-        img = self.apply_visible_masks(img)
-        disp_img = self.apply_tth_distortion(img)
+        img = self.apply_tth_distortion(img)
+        disp_img = self.apply_visible_masks(img)
         self.display_image = disp_img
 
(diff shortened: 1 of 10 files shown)


hexrdgui/calibration/polarview.py
```diff
@@ -491,7 +491,6 @@ def reapply_masks(self):
         # Both "visible" and "boundary" masks are applied to the
         # computational image
         comp_img = self.apply_boundary_masks(img)
-        comp_img = self.apply_tth_distortion(comp_img)
```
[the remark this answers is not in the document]
bnmajor: Yes and no - this commit appears completely out of sync with my local branch so I'm going to re-push those changes. The deletion is intentional but only because we're swapping the order of operations so we don't need to apply the tth distortion twice.

hexrdgui/masking/create_raw_mask.py
```diff
@@ -55,10 +53,74 @@ def convert_polar_to_raw(line_data, reverse_tth_distortion=True):
...
+                    # The correct intersection should be the closest to coords1
+                    distances = np.sqrt(
+                        ((coords1 - intersections)**2).sum(axis=1)
+                    )
+                    new_coords = intersections[np.argmin(distances)]
+
+                    add_coords.append(new_coords)
+                    add_indices.append(new_idx)
+
+                raw = np.insert(raw, add_indices, add_coords, axis=0)
```
psavery, line 122: @saransh13 Can you review this code section? Let me know if you think of a simpler or better way to do anything.

hexrdgui/calibration/polarview.py
```diff
@@ -401,14 +401,13 @@ def apply_image_processing(self):
...
         # We only apply "visible" masks to the display image
-        img = self.apply_visible_masks(img)
-        disp_img = self.apply_tth_distortion(img)
+        img = self.apply_tth_distortion(img)
+        disp_img = self.apply_visible_masks(img)
         self.display_image = disp_img
 
         # Both "visible" and "boundary" masks are applied to the
         # computational image
         comp_img = self.apply_boundary_masks(img)
```
psavery, line 410: @bnmajor For the `comp_img`, we want to apply *both* the visible masks and the boundary masks, right? So should the input to this function be the `disp_img` (where the visible masks have been applied)?

psavery: Same for `reapply_masks`

psavery: I believe this will fix #1690

bnmajor: Sounds good! Fixed now.

saransh13 (approved):
Nice job with the major improvement in performance, and fixing many bugs with the mask boundaries. LGTM

Issue #1694, opened 2024-04-19
polar view recalculation triggered after save state


saransh13:
After saving state file, the polar view seems to be calculated again. Can we switch this trigger off?

psavery:
To add further detail, I think this only happens when you save over a currently opened state file. I think this is happening because we have to re-create the imageseries so that they point to the new state file (the imageseries can no longer refer to the old file since it was overwritten, and new imageseries have to be made). We might still be able to avoid the recalculation, though.

PR #1696, opened 2024-04-25
Do not rerender when overwriting loaded state file

psavery:
We were previously rerendering, because overwriting a loaded state file requires us to reload the imageseries from the new state file, and we wanted to make sure that any changes would be reflected in the UI.

However, there really shouldn't be any changes, and rerendering sometimes takes quite a while (for example, in high resolution polar view). So just skip rerendering, and do not allow the colormap ranges to reset either.

Fixes: #1694


saransh13 (approved):
LGTM

PR #1706, opened 2024-06-11
Add option to set lmfit boundaries as a delta

psavery:
lmfit does not allow a delta to be used, but instead a min/max must be specified. We can, however, allow a delta to be used in the GUI, and then set the min/max on the lmfit parameters based upon this delta. That is what is being done in this PR.

Using delta for boundaries is more intuitive for some workflows.

### DIFF
--- a/hexrdgui/calibration/calibration_dialog.py
+++ b/hexrdgui/calibration/calibration_dialog.py
@@ -20,6 +20,7 @@ from hexrdgui.tree_views.multi_column_dict_tree_view import (
 )
 from hexrdgui.ui_loader import UiLoader
 from hexrdgui.utils.dialog import add_help_url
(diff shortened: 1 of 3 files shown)


hexrdgui/calibration/calibration_dialog.py
```diff
@@ -395,15 +462,31 @@ def initialize_tree_view(self):
...
+        # Keep the same scroll position
+        scrollbar = self.tree_view.verticalScrollBar()
+        scroll_value = scrollbar.value()
+
+        self.ui.tree_view_layout.removeWidget(self.tree_view)
+        self.tree_view.deleteLater()
+        del self.tree_view
+        self.initialize_tree_view()
+
+        # Restore scroll bar position
```
bnmajor, line 487: Nice! :)

bnmajor (approved):
LGTM!

PR #1708, opened 2024-06-24
Filter down engineering constraint types

psavery:
We are currently identifying TARDIS and PXRDIP both when `guess_instrument_type()` is used. However, only TARDIS has engineering constraints, and an exception is raised if we try to set engineering constraints using PXRDIP.

When guessing the engineering constraints, we should filter down (ignore) any instrument types for which we do not yet have engineering constraints defined. This prevents errors from occuring later.


saransh13 (approved):
LGTM

PR #1710, opened 2024-07-02
Physics package

bnmajor:
Relies on the HEXRD [absorption-correction](https://github.com/HEXRD/hexrd/tree/absorption-correction) branch.


bnmajor:
@psavery @saransh13 Some additional questions:

1. We currently expect there to be a `sample_materials` and `window_materials` file like the `pinhole_materials` which we do not have - do we need to create these files (and if so which materials should be included)? Or should we just use the `pinhole_materials` file?
    1. Same question for filter/coating/phosphor for the absorption correction dialog
2. We discussed automatically applying pinhole corrections and that the pinhole should be set to `None` if it is not needed but that is not currently an option.
    1. Do we want to change [this condition](https://github.com/HEXRD/hexrd/pull/649/files#diff-a3083700d9fbb1a60ea45abdc4cdbea90e7355d11af434834a9e89b5b300e927R621) to allow `None` as well? If so, is it safe to assume that we will never call [`calc_transmission`](https://github.com/HEXRD/hexrd/pull/649/files#diff-829d10c3bc618d49be61bb99870c7d45e5b9edabcbb5b2903db883b4e635e47bR2089) (which calls [`calc_effective_pinhole_area`](https://github.com/HEXRD/hexrd/pull/649/files#diff-a3083700d9fbb1a60ea45abdc4cdbea90e7355d11af434834a9e89b5b300e927R1691)) without a pinhole?
    2. Would it be simpler to say that if this is a `TARDIS` or `PXRDIP` (and eventually `BBXRD`) instrument we just automatically apply pinhole corrections?
3. How should we handle inferring instrument type? Currently we set that automatically if users come from the LLNL import tool, but how should we handle this if they do not? We have engineering constraints to determine if it is `TARDIS`, but no clear solution for `PXRDIP`.

psavery:
My thoughts are:

1. Yes we probably need to make those files (but I think `filter/coating/phosphor` will just be a dict within HEXRD)
2. I think we probably want to allow the transmission to be calculated without a pinhole (make the pinhole optional in transmission calculation)
3. You can just use [this function](https://github.com/HEXRD/hexrdgui/blob/master/hexrdgui/utils/guess_instrument_type.py), Brianna. That detects both TARDIS and PXRDIP automatically (and only those two so far) based upon the detector names.
4. It looks like we are finding the max transmission across all detectors and normalizing based upon that. I think that is the behavior we planned. Maybe something else is missing, @saransh13?

bnmajor:
> 3. You can just use [this function](https://github.com/HEXRD/hexrdgui/blob/master/hexrdgui/utils/guess_instrument_type.py), Brianna. That detects both TARDIS and PXRDIP automatically (and only those two so far) based upon the detector names.

Looks like this only can guess TARDIS, or am I missing something? If not I can add PXRDIP

hexrdgui/pinhole_correction_editor.py
```diff
@@ -475,18 +438,15 @@ def apply_panel_buffer_visible(self, b):
...
+        # If not default to "Enter Manually"
         for i in range(1, w.count()):
             name = w.itemText(i)
             if not name:
                 # It's a separator
                 continue
-
-            absorption_length = self.get_rygg_absorption_length(i)
-            if np.isclose(value, absorption_length):
+            if name == HexrdConfig().pinhole_package.material:
```
psavery, line 449: Are we still letting users manually edit the absorption correction length in the Rygg pinhole correction editor? If not, I don't think we need this, right?

saransh13 (approved):

PR #1712, opened 2024-07-03
Invalidate past critical beta

psavery:
Depends on: hexrd/hexrd#668


saransh13 (approved):
Works as expected. LGTM!

PR #1713, opened 2024-07-05
Do not set a parent for OverlayStylePicker

psavery:
Calling `exec()` on the overlay style picker appears to be hiding its parent dialog, which we don't want.

It's okay (arguably preferable) for this picker to be application-modal, so remove the parent.


saransh13 (approved):
LGTM

PR #1714, opened 2024-07-05
Do not apply pinhole distortion to pinhole mask

psavery:
It should not be affected by the pinhole distortion settings.


saransh13 (approved):
Fixed the issue with the pinhole mask. LGTM.

Issue #1716, opened 2024-07-09
Feature Request: Save a grains.out from the indexing step before fit-grains


gustafsonsven:
Hi all,

This one is a feature request.  Would it be possible to have an option to save an unfit grains.out from the indexing step before going into fitting?  The CLI drops an accepted orientations file (I think?  Maybe it now drops a grains.out) - perhaps it should also drop a grains.out.  

Thoughts?

Thanks - Sven

psavery:
I think if you run `hexrd find-orientations` in the CLI, it does drop an unfit grains.out file.

Maybe we should add a button to save the grains table in the "Indexing Results Dialog" or in the "Fit Grain Options" dialog? Or both?

gustafsonsven:
I agree - saving a unfit grains.out and grains.npz is ideal.  I think a button in the "Indexing Results Dialog" would be ideal.  I don't see a strong reason to not add it to the "Fit Grains Options" dialog besides clutter on that page.

PR #1718, opened 2024-07-23
Add ability to export grains from indexing results

psavery:
This adds an "Export Grains" button to the indexing results dialog where the user can write out the unfit grains.

If the file is saved with a `.npz` extension, the new NPZ grains file format is used. Otherwise, the old `.out` format is used.

Fixes: #1716


bnmajor (approved):
LGTM!

Issue #1724, opened 2024-08-20
Add "Mirror First Detector" button to calibration workflow dialog


psavery:
This will get all detector `vary` statuses (tilts/translations) to match that of the first detector. This would be a major improvement in cases with many detectors (32-panel Eiger, 6-panel FIDDLE, etc.).

PR #1725, opened 2024-08-20
Add button to mirror "Vary" from first detector in Calibration Dialog

psavery:
This is significantly helpful in cases like the 32 subpanel Eiger system,
so that users do not need to click "Vary" checkboxes for every single detector,
if they are changing them in the same way.

https://github.com/user-attachments/assets/9099d09f-adaf-4e21-b438-a941615616bb

Fixes: #1724


saransh13 (approved):
LGTM

PR #1726, opened 2024-08-30
Significantly speed up calibration tree view load

psavery:
It takes a *very* long time to manually expand every row in Qt tree views. It is *significantly* faster to use Qt's function to recursively expand all rows. Perhaps this is because a redraw of the table might be performed on every expand?

This changes the code so that all rows are recursively expanded at first, and then only needed rows are collapsed afterward. In my testing, even collapsing every row does not have a significant impact on performance, so I think this is the best way to go.

For loading a state file with the 32 subpanel eiger, the load time went from like 106 seconds to 5.7 seconds on my system, so this is a big speedup.


bnmajor (approved):
Awesome!

PR #1732, opened 2024-09-30
Fix several physics package issues

psavery:
This fixes several places to ensure a physics package is present if it is required.


saransh13 (approved):
LGTM!

PR #1734, opened 2024-09-30
Add support for multiple x-ray sources 

psavery:
This includes support for having overlays with different x-ray sources, and calibrating according to those overlays with different x-ray sources.


saransh13 (approved):
LGTM!

PR #1738, opened 2024-10-04
Only pad for stereo if needed

psavery:
Check if the start/stop values are already present, and if they are, do not add them.


saransh13 (approved):
LGTM

PR #1748, opened 2024-10-30
Remove comma so it is not interpreted as a tuple

psavery:
This was causing issues since it was interpreted as a tuple...


psavery:
This change got merged as a part of #1750

Issue #1741, opened 2024-10-15
Feature Request: HEDM Calibration Revision


kenygren:
Users are having trouble understanding a. _how_ the HEDM calibration works / what it's doing and b. how _to run_ the HEDM calibration. These two points are interrelated. 

Sven and I are proposing a few major changes: 
- Allow the user to select/view all refinements and the rotationseries/grain data needed for the optimization FROM the HEDM calibration landing page
- Change the wording of the selection options away from "clobbering" to be "refinement choices" and the bullet actually describes what is happening from an experimental point of view (see picture). The term clobber has caused a lot of confusion for users (sorry Joel and Darren) but should show up in the hover over. 
- _Auto populate_ the refinement pane based on what is selected or un selected in the optimization pane (with the option to override). 
![image](https://github.com/user-attachments/assets/fed0facd-12df-4f69-8e53-59f9c3afdf55)

PR #1750, opened 2024-11-02
Refactor HEDM calibration for short-term changes

psavery:
We eventually want to use features from our main calibration workflows, such as the "undo" button, and the relative constraints. However, we need to make short-term updates to the HEDM calibration workflow.

This makes several changes, including allowing users to see how the refinement options get changed with different refinement settings, allowing users to select grains to be used, and allowing users to see and set all parameters used as input for `pull_spots()`.

Fixes: #1741


psavery:
Merging so it is available for testing in the prerelease while I'm away...

PR #1754, opened 2024-11-08
Physics package fixes

saransh13:
The thicknesses of each layer kept increasing upon repeated selection of pinhole correction. The problem was due too using a dictionary by reference. Fixed by making a copy and editing it instead of original dictionary.

### DIFF
--- a/hexrdgui/physics_package_manager_dialog.py
+++ b/hexrdgui/physics_package_manager_dialog.py
@@ -142,7 +142,7 @@ class PhysicsPackageManagerDialog:
         self.ui.pinhole_density.setValue(physics.pinhole_density)
         if self.instrument_type == 'PXRDIP':
             self.ui.pinhole_thickness.setValue(70)
-            self.ui.pinhole_diameter.setValue(130)
+            self.ui.pinhole_diameter.setValue(300)
         else:
             self.ui.pinhole_thickness.setValue(physics.pinhole_thickness)
             self.ui.pinhole_diameter.setValue(physics.pinhole_diameter)
--- a/hexrdgui/overlays/powder_overlay.py
+++ b/hexrdgui/overlays/powder_overlay.py
@@ -654,7 +654,7 @@ class PowderOverlay(Overlay, PolarDistortionObject):
         kwargs = self.tth_distortion_kwargs.copy()
         if self.pinhole_distortion_type == 'RyggPinholeDistortion':
             # Add our absorption length
-            kwargs['absorption_length'] = HexrdConfig().absorption_length()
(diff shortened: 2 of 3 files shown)


hexrdgui/image_canvas.py
```diff
@@ -1067,15 +1102,66 @@ def finish_show_polar(self, iviewer):
...
+                # add grid lines parallel to y-axis
+                kwargs['which'] = 'both'
+                kwargs['axis'] = 'x'
+                axis.grid(**kwargs)
+
                 # Set our custom tick locators as well
-                self.axis.xaxis.set_major_locator(PolarXAxisTickLocator(self))
-                axis.xaxis.set_major_locator(PolarXAxisTickLocator(self))
+                # self.axis.xaxis.set_major_locator(PolarXAxisTickLocator(self))
+                # axis.xaxis.set_major_locator(PolarXAxisTickLocator(self))
```
psavery, line 1158: So, I think we were using these before so that when you set the x-axis to use Q instead, tick markers show up at `1, 2, 3, 4, ...`. Can you verify that still happens with the above changes? If not, we might need to fix that before merging.

saransh13: Yes, looks like switching to `Q` messes up the `xticklabels`. We'd have to redo this step when the axis is updated to use `Q`.

Screenshot 2024-11-11 at 2 57 17 PM
Screenshot 2024-11-11 at 2 57 32 PM

psavery: So actually, those `PolarXAxisTickLocator` classes inherit `AutoLocator`. I think that in the code above, where you have `yaxis.set_major_locator(AutoLocator())`, we might be able to just change that `AutoLocator` to `PolarXAxisTickLocator`.

However, we probably also need to change the minor locator, too.

hexrdgui/physics_package_manager_dialog.py
```diff
@@ -142,7 +142,7 @@ def setup_form(self):
         self.ui.pinhole_density.setValue(physics.pinhole_density)
         if self.instrument_type == 'PXRDIP':
             self.ui.pinhole_thickness.setValue(70)
-            self.ui.pinhole_diameter.setValue(130)
+            self.ui.pinhole_diameter.setValue(300)
```
psavery, line 145: Maybe these should be set by the physics package too?

psavery (approved):
I removed the polar view style changes. If this looks good to you, feel free to merge.

hexrdgui/physics_package_manager_dialog.py
saransh13: This was to account for the different default pinhole dimensions in PXRDIP and TARDIS. I think it's still editable in the physics package.

PR #1761, opened 2024-11-27
Fiddle instrument

bnmajor:
- [x] New instrument option in LLNL Import: `FIDDLE`
- [x] Two modes: `With Image PLate` and `Without Image Plate`
- [x] With no image plate = load image per detector w/ dark subtraction for each image
- [x] With image plate = images + dark + IP w/ template & masking
- [x] Try to automatically find remaining images + dark after first image is loaded
- [x] Support manually loading images + dark if automatic find fails
- [x] Add hdf5 path as known path so users do not need to select it
- [x] Need correct default config file  (See HEXRD/hexrd#736)
- [x] Need template boundary(ies) file  (See HEXRD/hexrd#736)
- [x] Imageseries should say `frame` or `timestep` instead of `omega`
- [x] Image + dark = 4 frames per image file/dark file
    - [x] Subtract each dark frame from each image frame
    - [x] Create 4 frame imageseries from results
- [x] All 4 frames are in single hdf5 file. Need to handle multiple hdf5 paths for this new case
- [x] ~Need 4th config option: ATLAS coordinates - Saransh will provide script and example data~ (Will put up in separate PR)
- [x] Add new intensity correction option: median w/ kernel (default to 7; user can change)
- [x] On load for FIDDLE instrument pop up dialog w/ option to apply new median filter and set kernel size before Physics Package dialog


hexrdgui/main_window.py
```diff
@@ -5,6 +5,7 @@
 import tempfile
 
 import h5py
+from hexrdgui.median_filter_dialog import MedianFilterDialog
```
psavery, line 8: The imports are sorted like this:

* system imports (imports included in Python's standard library, like functools, os, etc.)
* Most third-party modules (unless noted below)
* Qt modules
* hexrd modules
* hexrdgui modules

They're also all alphabetized (or they're supposed to be, at least).

Can you put this new import with the other `hexrdgui` imports, in alphabetical order?

hexrdgui/hexrd_config.py
```diff
@@ -10,6 +10,7 @@
 import matplotlib
 import numpy as np
 import yaml
+from scipy.signal import medfilt2d
```
psavery, line 13: Can you alphabetize this import?

hexrdgui/hexrd_config.py
```diff
@@ -1002,6 +1009,13 @@ def intensity_corrected_images_dict(self):
             for name, img in images_dict.items():
                 images_dict[name] = img - minimum
 
+        if HexrdConfig().apply_median_filter_correction:
+            for name, img in images_dict.items():
+                images_dict[name] = medfilt2d_memoized(
+                    img,
+                    kernel_size=HexrdConfig().median_filter_kernel_size
+                )
```
psavery, line 1017: I'll need to remember to move this with the other intensity corrections in my other PR...

hexrdgui/median_filter_dialog.py
```diff
@@ -0,0 +1,40 @@
...
+        self.ui.kernel_size.setValue(self.kernel_size)
+
+        self.update_gui()
+        self.setup_connections()
+
+    def update_gui(self):
+        # Cannot exceed array dimensions
+        img = HexrdConfig().image(HexrdConfig().detector_names[0], 0)
+        upper_bound = min(img.shape[0], img.shape[1])
+        self.ui.kernel_size.setMaximum(upper_bound)
```
psavery, line 23: Do we need to check the shape of all images? Or just the image for the first detector?

I could imagine this filter being used for other instrument types too.

hexrdgui/resources/calibration/default_fiddle_config.yml
```diff
@@ -0,0 +1,15 @@
...
+    pixels:
+      columns: 512
+      rows: 1024
+      size: [0.025, 0.025]
+    transform:
+      tilt: [1.8496392739293273, 2.5380251768660202, -0.01561236535824728]
+      translation: [-46.44506775, -1.82675053, 56.74914575]
+oscillation_stage:
+  chi: 0.0
+  translation: [0.0, 0.0, 0.0]
```
psavery, line 15: Is this file needed, considering we have one with all of the detectors in `hexrd`? If we still need this, it's fine. I'm just checking...

hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -1,18 +1,22 @@
+import glob
 import os
+import re
+from hexrdgui.median_filter_dialog import MedianFilterDialog
```
psavery, line 4: Can you put this with the other `hexrdgui` imports below? It might also be nice to sort all of the imports :slightly_smiling_face:

hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -145,56 +197,99 @@ def get_instrument_defaults(self):
...
         if self.ui.instrument.isEnabled():
             self.get_instrument_defaults()
 
+    def instrument_settings_changed(self, index):
+        has_ip = index == 0
+        self.set_widget_visibility(
+            self.ui.image_plate_raw_image,
+            self.ui.template_instructions,
+            self.ui.outline_appearance,
+        visible=has_ip)
```
psavery, line 237:
```suggestion
            visible=has_ip)
```

hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -297,23 +409,149 @@ def update_bbox_width(self, val):
...
+        img = [[]] * FIDDLE_FRAMES
+        first, last = '/'.join(FIDDLE_HDF5_PATH).rsplit('0', 1)
+        for frame in range(FIDDLE_FRAMES):
+            # Use known path to find data 0/1/2/3 paths
+            path = f'{first}{frame}{last}'
+            # Load in detector data and dark data for this frame
+            with h5py.File(data_file) as data:
+                data_raw = np.array(data[path])
+            with h5py.File(dark_file) as dark:
+                dark_raw = np.array(dark[path])
```
psavery, line 436:
```suggestion
            with h5py.File(dark_file) as dark:
                dark_raw = np.array(dark[path])
```

Looks like these lines were duplicated

hexrdgui/median_filter_dialog.py
bnmajor: Actually, this should either be a per-image setting or use the smallest height/width as the max to prevent the kernel size from exceeding the shape of any of the images...

Also this value is a scalar based on this exact use case, but it could also be [a list of two](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.medfilt2d.html#scipy.signal.medfilt2d).

Maybe we should just use the overall minimum for now and leave a comment as a reminder on how this can be extended in the future if there is a use for it?

hexrdgui/resources/calibration/default_fiddle_config.yml
bnmajor: This approach could certainly be re-thought a little and corrected to use the `hexrd` files, but it would require some changes to the logic. I think it requires enough changes that it shouldn't be done on this branch, but I'd be happy to open this as an issue and put up a new PR for it!

hexrdgui/llnl_import_tool_dialog.py
bnmajor: Actually one is for the data file, the other for the dark file

psavery: Ah I see! :slightly_smiling_face:

hexrdgui/median_filter_dialog.py
psavery: Sure! Let's just use the overall minimum for now.

Maybe you can change it to something like this?

```python
upper_bound = np.iinfo(np.int32).max
for det_key in HexrdConfig().detector_names:
    img = HexrdConfig().image(det_key, 0)
    upper_bound = min(upper_bound, *img.shape)

Comment (bnmajor) [reply hexrdgui/median_filter_dialog.py]: I had something a [little different](https://github.com/HEXRD/hexrdgui/pull/1761/commits/1c94d553f0b925306aaaa49b8577bd6c05ebaefd#diff-6f2c5e63a9f95f46c3c62e41aa54137dee64b0ee2e3866ce446ce1f78c0835d3L19-L23) stashed in case you were okay with the overall min approach - let me know if you'd prefer I use your suggested changes instead though!

Comment (psavery) [reply hexrdgui/median_filter_dialog.py]: Looks fine to me!

File: hexrdgui/median_filter_dialog.py [hunk 9]
```diff
@@ -0,0 +1,51 @@
...
+    def exec(self):
+        return self.ui.exec()
+
+    @property
+    def kernel_size(self):
+        return HexrdConfig().median_filter_kernel_size
+
+    def accept_changes(self):
+        value = self.ui.kernel_size.value()
+        if value % 2 != 0:
```
psavery, line 44:
```suggestion
        if value % 2 == 0:
```

Should it be this instead?

bnmajor: Oof, yes it should, thanks!

psavery (approved):
LGTM!

PR #1762, opened 2024-12-11
HEDM Calibration: restore grain settings

psavery:
In the HEDM calibration, if the user does things like unchecking whether the strain is fixed to identity, restore the previous setting that the strain had.

This way, modifying options won't result in a permanent loss of the previous settings. It also enables us to fix the strain by default (which was recommended).


bnmajor (approved):
LGTM!

PR #1763, opened 2024-12-11
Upgrade macos to 13 for packaging

psavery:
macos 12 is deprecated and will not work soon.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -22,7 +22,7 @@ jobs:
           }
         - {
             name: "MacOSX",
-            os: macos-12,
+            os: macos-13,
             package: 'dmg'
           }
         - {


psavery:
After doing some research, it seems that macOS 13 has more [spurious failures](https://github.com/actions/runner-images/issues/7522) when creating DMGs, due to a race condition with `XProtect`. So even though the CI here just passed, it may fail spuriously in the future. The solution appears to just essentially try repeatedly, such as is done [here](https://github.com/fredowski/osxbundler/commit/67d668c579d76426e780370eaf6d02858be89180).

We may need to replace the `CPACK_COMMAND_HDIUTIL` environment variable with a custom script that tries `hdiutil` repeatedly, 10 times, before actually failing. I'll do that in a separate PR.

bnmajor (approved):

PR #1768, opened 2024-12-17
Repeat "hdiutil create" commands to avoid failures

psavery:
Race conditions with XProtect on macos-13 can cause spurious failures
for `hdiutil create` commands.

As a work-around, try the `hdiutil create` command several times (up 
to 10), and only fail if the command fails each time.


bnmajor (approved):
Nice!

PR #1769, opened 2024-12-19
fix scaling when dealing with masked arrays

saransh13:
The cartesian view is a `numpy.ma.masked_array`. This was not correctly handled by the polar plot intensity scaling functions.


psavery (approved):
I made it a little cleaner

PR #1773, opened 2025-01-08
Separate physics package and detector filters

bnmajor:
All instruments should have the option to apply the physics package, the absorption correction, or both. This PR:
- Removes restrictions around which instruments have the option to add a physics package
- Removes restrictions around which instruments have the option to apply absorption correction
- The physics package is no longer a prerequisite for absorption correction

Relies on HEXRD/HEXRD#741

### DIFF
--- a/hexrdgui/main_window.py
+++ b/hexrdgui/main_window.py
@@ -482,8 +482,8 @@ class MainWindow(QObject):
             # Turn off absorption correction
             self.ui.action_apply_absorption_correction.setChecked(False)
 
-            # Set the physics package to None
-            HexrdConfig().update_physics_package()
+            # Remove the physics package
+            HexrdConfig().physics_package = None
 
             # Turn off all detector coatings
             HexrdConfig().detector_coatings_dictified = {}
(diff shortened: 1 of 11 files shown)


hexrdgui/main_window.py
```diff
@@ -1562,7 +1560,12 @@ def on_action_hedm_import_tool_triggered(self):
         self.simple_image_series_dialog.show()
 
     def on_action_llnl_import_tool_triggered(self):
-        self.llnl_import_tool_dialog.show()
+        dialog = self.llnl_import_tool_dialog
+        dialog.show()
+        # Always assume Physics Package is needed for LLNL import
+        dialog.ui.complete.clicked.connect(
+            lambda: self.on_action_include_physics_package_toggled(True)
```
psavery, line 1567: This would create a new connection every time the LLNL import tool is shown. I'll move this to `setup_connections()` instead. Otherwise, this could get triggered multiple times.

psavery (approved):
LGTM in the things I've tested so far.

PR #1776, opened 2025-01-14
Apply all intensity corrections after SNIP in polar

psavery:


saransh13:
There are fixes in Hexrd branch https://github.com/HEXRD/hexrd/pull/793 that are needed to be merged before this branch works. Otherwise LGTM.

PR #1789, opened 2025-02-14
Export masked lineout intensities as nans

psavery:
Previously, the masked lineout intensities were being filled with zeros accidentally (through a conversion to a numpy array). Instead, we want to fill the masked intensities with nan before exporting.

This fixes the issue.


saransh13 (changes requested):
Can you filter out the two theta points with `nan` values from the azimuthal average? I don't these points should be included in the output.

psavery:
@saransh13 Fixed. Ready for review again.

saransh13 (approved):
LGTM

PR #1796, opened 2025-02-28
Modify several settings to increase canvas size

psavery:
This reduces the required widths of the image mode widget and the calibration widget in order to allow the main canvas to be larger.

The users can still increase the sizes of these widgets if they prefer, but on some computers, like Mac laptops with lower resolution settings, it is very helpful to allow the widgets to be smaller, and the canvas to be larger.

We typically should prefer having a very large canvas, as that contains the most important thing to see: the data.


saransh13 (approved):
LGTM

PR #1797, opened 2025-03-03
Use buttons in toolbar

bnmajor:
This branch also needs an okay from UDRI folks since this change would affect all user workflows (although hopefully as an improvement for all).
Feedback and remaining changes:
- [x] Make sure the frame can still be input manually
- [x] Investigate omega to make sure that the toolbar is not being re-sized/shifting around (Note: There is no fix in this PR but I was unable to reproduce re-sizing/shifting)

### DIFF
--- a/hexrdgui/image_series_toolbar.py
+++ b/hexrdgui/image_series_toolbar.py
@@ -1,11 +1,14 @@
 from pathlib import Path
 from PySide6.QtCore import QCoreApplication, Qt
-from PySide6.QtWidgets import QGridLayout, QLabel, QSlider, QSpinBox, QWidget
+from PySide6.QtWidgets import (
+    QGridLayout, QLabel, QLineEdit, QPushButton, QSlider, QWidget
+)
 from PySide6.QtGui import QFontMetrics, QPixmap
 from hexrdgui import resource_loader
 
 import hexrdgui.resources.icons
 from hexrdgui.hexrd_config import HexrdConfig
+from hexrdgui.utils import block_signals
 
 
 class ImageSeriesInfoToolbar(QWidget):
@@ -62,6 +65,8 @@ class ImageSeriesToolbar(QWidget):
         self.ims = ims
         self.slider = None
         self.frame = None
+        self.back_button = None
+        self.forward_button = None
         self.layout = None
         self.widget = None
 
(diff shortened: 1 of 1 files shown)


hexrdgui/image_series_toolbar.py
```diff
@@ -161,6 +177,17 @@ def update_omega_label_text(self):
 
         self.omega_label.setText(omega_label_text(*ome_range))
 
+    def update_back_forward_buttons(self, val):
+        self.back_button.setEnabled(self.slider.minimum() != val)
+        self.forward_button.setEnabled(self.slider.maximum() != val)
+
+    def change_frame(self, value):
```
psavery, line 184:
```suggestion
    def shift_frame(self, value):
```
Can you rename this so it's clear that we are shifting by the value, rather than changing *to* the value?

hexrdgui/image_series_toolbar.py
```diff
@@ -74,14 +79,19 @@ def __init__(self, ims, parent=None):
 
     def setup_connections(self):
         self.slider.valueChanged.connect(self.val_changed)
-        self.slider.valueChanged.connect(self.frame.setValue)
-        self.frame.valueChanged.connect(
-            self.slider.setSliderPosition)
+        self.slider.valueChanged.connect(lambda i: self.frame.setText(str(i)))
+        self.frame.textChanged.connect(
+            lambda i: self.slider.setSliderPosition(int(i)))
```
psavery, line 84:
```suggestion
        self.frame.editingFinished.connect(self.on_frame_edited)
```

I think we should only trigger an update after the user has finished editing. Otherwise, if they type in `11`, for example, it triggers an update with every keystroke (so first loads frame `1` and then loads frame `11`).

But `editingFinished` also does not supply the new text as an argument, so you'd also have to check it with `self.frame.text()`.

Also, we need to do some validation for the text that the user enters. I can type in invalid text like `f` and it produces an exception. Instead, we should do something like this:

```python
    try:
        val = int(text)
    except ValueError:
        # Not a valid integer. Restore the previous value.
        val = prev_val

    # Clip the value to the min/max range
    val = max(min_val, val)
    val = min(max_val, val)
```

Notice also we should clip the value to the min/max range.

psavery:
We need to increase the width of the line edit based upon the max number of digits in the range. Notice that if the max value is 4 digits, when we get to the 4 digit numbers, part of the number is cut off. The below number is `1351`.
image

For FIDDLE, it can be just wide enough to hold 1 digit, since that's all it will ever have.

Maybe you can do something like this: `num_digits = len(str(max_frame_index))`.

psavery (approved):

PR #1800, opened 2025-03-07
Refactor HEDM calibration to use new workflow

psavery:
This refactors the HEDM calibration to use the new calibration workflow we have been using for powder and Laue. It includes the use of lmfit for boundary constraints, support for relative constraints between detectors (including for subpanels), and many highly desirable features such as undo/redo.

This refactoring completely eliminates the need for the old calibration workflow, and thus a lot of old code is being removed as well, including the ability to set refinements in the instrument tree view and the refinements editor.

Although this still needs to be further tested, it is overall a major improvement for HEDM calibration.

Depends on: hexrd/hexrd#779


psavery:
Since this is somewhat urgent, we'll get it in and perform any necessary cleanup later...

PR #1802, opened 2025-03-12
Fix issue with panel buffer dialog losing focus

psavery:
This fixes an issue where `setData()` would be called with the `enable` status of the panel buffer button when the panel buffer button loses focus! This was highly unexpected behavior, and resulted in a lot of errors and issues...


psavery:
Merging, as this is needed somewhat urgently...

PR #1805, opened 2025-03-14
Fix the application name on macos

psavery:
Unfortunately, this requires a third-party library, since macos requires these things to be edited in Objective C (which Python cannot natively do).

However, the third-party library appears reliable.

Here is what the app name appeared as previously:

Screenshot 2025-03-13 at 7 46 18 PM

But now it displays "HEXRD" instead of "Python" in both places.


saransh13 (approved):
Fixed the issue. LGTM

PR #1806, opened 2025-03-14
Atlas config

bnmajor:
DRAFT: Depends on #1761 and HEXRD/hexrd#736


bnmajor:
@psavery Let me know if you need data to test this branch with

bnmajor:
@saransh13 This is ready for feedback when you have the chance!

saransh13 (approved):
Tested this and the results look correct to me. LGTM

hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -39,6 +39,65 @@
 from hexrdgui.utils.dialog import add_help_url
 
 
+class AtlasConfig:
```
psavery, line 42: I think it'd be nicer if `AtlasConfig` was on its own separate file, such as `atlas_config.py`. But we can leave it for this PR.

psavery (approved):

PR #1810, opened 2025-03-17
Add option to save images in fch5 format

psavery:
This also "fixes" the issue where, when you have multiple detectors, the dialog asking for the threshold would pop up once for every detector.

Now it just pops up once and uses the same threshold for all detectors.


ChristosT (approved):
Looks good. I verified that it works locally.

PR #1811, opened 2025-03-18
Calibration distortion fixes

psavery:
This implements several fixes for using detector distortion parameters during HEDM calibration.


psavery:
Merged as this is somewhat urgent.

PR #1813, opened 2025-03-20
Add ability to create waterfall plots

psavery:
If the loaded data is an imageseries with 2 to 10 frames per detector, a new button appears in the polar view settings named "Waterfall Plot".

If clicked, an azimuthal lineout is computed for every frame in the imageseries, and the results are displayed in an interactive waterfall plot.

Since the azimuthal lineout intensities are arbitrarily scaled and offset, the plots can be vertically stacked on top of one another with some spacing between. Clicking and dragging allows one to move a plot up/down. Shift-clicking and dragging also allows the plot to move left/right (which is typically not advised since two theta should not change). Using the mouse wheel while hovering over a lineout causes that lineout to scale up/down in relative intensity.

In summary, this allows for easy comparison of the different azimuthal lineouts for each frame in the imageseries.

image

### DIFF
--- a/hexrdgui/calibration/polarview.py
+++ b/hexrdgui/calibration/polarview.py
@@ -55,9 +55,6 @@ class PolarView:
             # Use an image dict with the panel buffers applied.
             # This keeps invalid pixels from bleeding out in the polar view
             self.images_dict = HexrdConfig().images_dict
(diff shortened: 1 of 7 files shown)


saransh13 (approved):
LGTM. Ready to merge!

PR #1814, opened 2025-03-21
Fix median filter menu checkbox

bnmajor:
When the median filter is set immediately after the LLNL import workflow is complete the menu option still needs to be toggled "on" if the filter is applied. Make sure that it is.

### DIFF
--- a/hexrdgui/llnl_import_tool_dialog.py
+++ b/hexrdgui/llnl_import_tool_dialog.py
@@ -46,7 +46,7 @@ class LLNLImportToolDialog(QObject):
 
     cancel_workflow = Signal()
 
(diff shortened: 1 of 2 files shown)


hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -46,7 +46,7 @@ class LLNLImportToolDialog(QObject):
 
     cancel_workflow = Signal()
 
-    complete_workflow = Signal()
+    complete_workflow = Signal(bool)
```
psavery, line 49: Can you leave a comment above this line as well that indicates what the `bool` is for?

Issue #1791, opened 2025-02-26
Improve Masking


bnmajor:
- [x] Offer on-hover feedback that indicates which mask is being used in a current area
- [x] Use "apply" instead of immediate changes to prevent excess re-rendering
- [x] Make sure masks in one view are disabled when switching view
    - [x] Masks created in XRS1 are disabled/turned off when switching to XRS2
    - [x] Masks created in raw are exempt regardless of which x-ray source they belong to
- [x] Move from table view to tree view for mask manager dialog
    - [x] Masks should be nested under the view they we created in (`raw`, `polar`, `polar/XRS1`, `polar/XRS2`, etc.)

bnmajor:
@psavery @saransh13 Please feel free to add to this if more items come up!

psavery:
@bnmajor Did all of these get resolved?

bnmajor:
Yes! This issue was resolved when #1815 was merged in.

PR #1815, opened 2025-03-26
New masking features

bnmajor:
- [x] Only apply changes to mask presentation on button press, not automatically
- [x] Remember which image mode a mask was created in
- [x] Remember which x-ray source was selected when the mask was created (if applicable)
- [x] Update mask manager dialog to use a tree view instead of a table view
- [x] Organize masks by view mode + x-ray source in tree
- [x] Keep backwards compatibility with old mask file versions
- [x] Automatically hide masks that were not created in the current view mode with the current x-ray source
- [x] Collapse mask categories that do not match current x-ray source
- [x] ~Allow users to skip this automatic mask hiding functionality if desired~ Skipping for now
- [x] Make sure that masks are sorted properly based on name then value
- [x] Remove old print statement in the hand drawn mask dialog
- [x] Add new top-level category that is not view mode specific ("global" or something similar)
- [x] Move info and "apply changes" button to bottom of dialog
- [x] Test threshold masking in all image modes

Addresses #1791

### DIFF
--- a/hexrdgui/image_canvas.py
+++ b/hexrdgui/image_canvas.py
@@ -2012,6 +2012,7 @@ class ImageCanvas(FigureCanvas):
         all_verts = []
         for name in MaskManager().visible_boundaries:
             mask = MaskManager().masks[name]
+            verts = None
             if self.mode == ViewType.raw:
                 if HexrdConfig().stitch_raw_roi_images:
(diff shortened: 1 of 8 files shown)


psavery:
This was just rebased on master...

PR #1816, opened 2025-03-31
Verify calibrators exist before updating them

psavery:
This fixes the "Undo" button for the structureless calibrator, which does not have a list of calibrators like the instrument calibrator does.


saransh13 (approved):
LGTM

PR #1818, opened 2025-04-02
Fiddle image names

bnmajor:
As requested by @saransh13 :

Previously when users changed the detector or image plate that had already been loaded and/or cropped the text was not updated and the image was not re-loaded. Now the previously selected image is loaded, the text label is updated, and if
it is an image plate the completed template is drawn in the preciously selected position.

Additionally, the completed detectors list now also contains a reference to the file associated with the completed detectors and/or image plates.

### DIFF
--- a/hexrdgui/llnl_import_tool_dialog.py
+++ b/hexrdgui/llnl_import_tool_dialog.py
@@ -133,7 +133,7 @@ class LLNLImportToolDialog(QObject):
         self.current_image_selection = None
         self.defaults = {}
         self.import_in_progress = False
-        self.loaded_images = []
+        self.loaded_images = {}
         self.canvas = parent.image_tab_widget.active_canvas
         self.detector_images = {}
         self.atlas_coords = None
@@ -162,6 +162,7 @@ class LLNLImportToolDialog(QObject):
             self.ui.instr_settings,
             self.ui.load_atlas,
             self.ui.atlas_label,
+            self.ui.completed_dets_and_ips,
             visible=False)
 
         self.update_config_settings()
@@ -210,9 +211,6 @@ class LLNLImportToolDialog(QObject):
         for w in widgets:
             w.setVisible(visible)
             w.setEnabled(visible)
-        # Resize on show/hide widgets to prevent large empty spaces
-        self.ui.setMinimumHeight(315)
-        QTimer.singleShot(0, lambda: self.ui.adjustSize())
 
     def set_default_color(self):
         self.outline_color = '#00ffff'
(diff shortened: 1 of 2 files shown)


hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -751,8 +789,19 @@ def complete_current_selection(self):
...
+            self.ui.completed_dets_and_ips,
+            visible=True)
+        if self.instrument == 'PXRDIP':
+            shared_file = next(iter(self.loaded_images.values()))
+            self.loaded_images[self.current_image_selection] = shared_file
+        loaded = [
+            f'{d}: {Path(self.loaded_images[d]).stem}'
+            for d in sorted(self.loaded_images)
+        ]
+        text = '\n'.join(set(loaded))
```
psavery, line 802: Is the conversion to a `set` done to remove any duplicates?

hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -540,20 +575,22 @@ def load_detector_images(self):
                 self.ui.accept_detector,
                 enabled=bool(selected['data'] and selected['dark']))
 
-    def load_images(self):
+    def load_images(self, checked=False, selected_file=None):
```
psavery, line 578: Is this `checked` argument used for anything?

hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -376,6 +374,19 @@ def detector_selected(self, selected):
             self.ui.instrument.setDisabled(selected)
             self.detector = self.ui.detectors.currentText()
             self.current_image_selection = self.detector
+            # If we have already loaded this detector's image, reload it
+            if self.detector in self.completed:
+                self.current_image_selection = self.detector
+                # Find the data and dark files for this detector
+                data_file = self.loaded_images[f'{self.detector} data']
+                dark_file = self.loaded_images[f'{self.detector} dark']
```
psavery, line 382: Does this work for all instrument types (TARDIS, PXRDIP, and FIDDLE)?

hexrdgui/llnl_import_tool_dialog.py
bnmajor: This is also connected to the [self.ui.image_plate_load.clicked](https://github.com/HEXRD/hexrdgui/pull/1818/files#diff-ac087ac85102dd9829f7addaa727a770a24b4a510dd23ec4b6faf36cfe9fccc8R114) signal, which passes in a `checked` argument.

hexrdgui/llnl_import_tool_dialog.py
bnmajor: Yes, although PXRDIP uses the same image for all image plates

hexrdgui/llnl_import_tool_dialog.py
bnmajor: Yes, but it unnecessary - I've removed it now.

psavery (approved):

PR #1819, opened 2025-04-02
Support drag and drop on main window

bnmajor:
If a file or files are dropped on the main window we will attempt to automatically load them as a state file, instrument config, materials file, or as images.

### DIFF
--- a/hexrdgui/main_window.py
+++ b/hexrdgui/main_window.py
@@ -1055,7 +1055,7 @@ class MainWindow(QObject):
                         return True
                 # If the main window is closing, save the config settings
                 HexrdConfig().save_settings()
-            elif event.type() == QEvent.DragEnter or event.type() == QEvent.Drop:
+            elif event.type() in (QEvent.DragEnter, QEvent.Drop):
                 self.validateDragDropEvent(event)
                 return True
 
--- a/hexrdgui/main_window.py
+++ b/hexrdgui/main_window.py
@@ -541,12 +541,13 @@ class MainWindow(QObject):
 
         return selected_file
 
-    def open_image_files(self):
-        # Get the most recent images dir
-        images_dir = HexrdConfig().images_dir
+    def open_image_files(self, checked=False, selected_files=None):
+        if selected_files is None:
+            # Get the most recent images dir
+            images_dir = HexrdConfig().images_dir
 
-        selected_files, selected_filter = QFileDialog.getOpenFileNames(
-            self.ui, dir=images_dir)
+            selected_files, selected_filter = QFileDialog.getOpenFileNames(
+                self.ui, dir=images_dir)
 
         if selected_files:
             # Save the chosen dir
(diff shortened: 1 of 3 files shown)


hexrdgui/main_window.py
```diff
@@ -1730,3 +1735,55 @@ def action_apply_median_filter_toggled(self, b):
...
+
+        # Make sure the selction(s) are files with a supported file extensions
+        paths = [url.toLocalFile() for url in mime_data.urls()]
+        supported_extensions = [
+            '.hexrd',
+            '.yml', '.yaml',
+            '.h5', '.hdf5',
+            '.tif', '.tiff',
+            '.png',
+            '.jpg', '.jpeg'
```
psavery, line 1753: So, for images, we actually support any type of image file that fabio supports (which is a lot of different kinds).  If a file comes in with an extension we don't recognize, maybe we should just try to open it as an image file and see what happens?

We also support frame cache formats in `.fch5` and `.npz` extensions. See [here](https://github.com/HEXRD/hexrdgui/blob/ff267ba3d0becf870d988b891bfadf7647558dec/hexrdgui/image_file_manager.py#L97-L100). We also support [hdf4 files too, right](https://github.com/HEXRD/hexrdgui/blob/ff267ba3d0becf870d988b891bfadf7647558dec/hexrdgui/image_file_manager.py#L64-L68)?

hexrdgui/main_window.py
```diff
@@ -1730,3 +1735,55 @@ def action_apply_median_filter_toggled(self, b):
...
+                    except:
+                        # If that fails, try loading it as a materials file
+                        HexrdConfig().load_materials(paths[0])
+                elif ext in ('.hexrd', '.yml', '.yaml'):
+                    try:
+                        # Try loading it as an instrument config first
+                        HexrdConfig().load_instrument_config(paths[0])
+                    except:
+                        # If that fails, try loading it as an image file
+                        self.open_image_files(selected_files=paths)
```
psavery, line 1783:
```suggestion
                    # Try loading it as an instrument config first
                    HexrdConfig().load_instrument_config(paths[0])
```

There's already a `try/except` block outside of this that tries to open as image files.

hexrdgui/main_window.py
```diff
@@ -1730,3 +1735,55 @@ def action_apply_median_filter_toggled(self, b):
...
+        if len(paths) == 1:
+            try:
+                # if hdf5 try as state, then materials, then images
+                if ext in ('.h5', '.hdf5'):
+                    try:
+                        # Try loading it as a state file first
+                        self.load_state_file(paths[0])
+                    except:
+                        # If that fails, try loading it as a materials file
+                        HexrdConfig().load_materials(paths[0])
```
psavery, line 1776: I think it's true that we normally raise exceptions when we attempt to load a file that isn't the right kind. Since we are now relying on it, now we should definitely make sure we do that :slightly_smiling_face:.

hexrdgui/main_window.py
bnmajor: Thanks! I actually thought I left a comment about this exact thing, must not have hit "comment" :upside_down_face: 

Makes sense though, my fear was that I was being too restrictive with the list which is in fact the case. I've dropped the list of expected file extensions and fall back to trying to open as an image.

hexrdgui/main_window.py
bnmajor: I've made some improvements here and we do get an error when trying to load non-image file(s):
```
Failed to process imageseries.

zero length imageseries.
```

If you try to load an invalid image file you will still have to go through the steps of selecting the path (for `.h5`, `.hdf5`) and associating the image(s) with the correct detectors before you get the error, both of which can be cancelled if you realize this was a mistake.

hexrdgui/main_window.py
```diff
@@ -1779,4 +1779,8 @@ def dropEvent(self, paths):
             # extension is not in known list
             self.open_image_files(selected_files=paths)
         except Exception as e:
-            print(f'Error loading images: {e}')
+            error_message = (
+                'Unable to guess file type (state, instrument, materials, or '
+                'image). Please use File menu to load.'
+            )
```
psavery, line 1785:
```suggestion
            )
            print(error_message)
```
Can you also print this out, Brianna?

psavery: Oh actually, scratch this. I think there's something better to do.

psavery (approved):

PR #1820, opened 2025-04-02
Support zoom by scroll wheel

bnmajor:


psavery:
I'd love to include this, but I do think it's a little slow (not your fault). We are probably going to have to figure out some ways to make matplotlib update the image faster when zooming to make this worth it. `draw()` and `draw_idle()` are just not fast enough.

PR #1823, opened 2025-04-07
Don't use `i` and `j` for polar distortion

psavery:
We were accidentally re-assigning `i` and `j` within that loop. Use a different variable name instead.


saransh13 (approved):
Nice work tracking this down! LGTM

PR #1826, opened 2025-04-09
Fix filter options for absorption correction

bnmajor:
Make sure that we manage filter inputs separately for each detector.

### DIFF
--- a/hexrdgui/absorption_correction_options_dialog.py
+++ b/hexrdgui/absorption_correction_options_dialog.py
@@ -112,7 +112,10 @@ class AbsorptionCorrectionOptionsDialog:
     def setup_connections(self):
         for k, w in self.material_selectors.items():
             w.currentIndexChanged.connect(
-                lambda index, k=k: self.material_changed(index, k))
+                lambda index, k=k: self.material_changed(k, index=index))
+        for k, w in self.material_inputs.items():
+            w.textChanged.connect(
+                lambda text, k=k: self.material_changed(k, text=text))
         self.ui.filter_density.valueChanged.connect(self.filter_info_changed)
         self.ui.filter_thickness.valueChanged.connect(self.filter_info_changed)
         self.ui.detectors.currentTextChanged.connect(self.detector_changed)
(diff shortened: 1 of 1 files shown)


hexrdgui/absorption_correction_options_dialog.py
```diff
@@ -10,7 +11,14 @@ def __init__(self, parent=None):
                                    parent)
 
         self.additional_materials = {}
-        self.filters = {n: {} for n in HexrdConfig().detector_names}
+        self.filters = {
+            det: {
+                'material': HexrdConfig().detector_filter(det).material,
+                'density': HexrdConfig().detector_filter(det).density,
+                'thickness': HexrdConfig().detector_filter(det).thickness
+            }
```
psavery, line 19: Does the above work too if there is no detector filter for that detector?

hexrdgui/absorption_correction_options_dialog.py
```diff
@@ -10,7 +11,19 @@ def __init__(self, parent=None):
...
 
         self.additional_materials = {}
-        self.filters = {n: {} for n in HexrdConfig().detector_names}
+        # Get filter defaults for each detector
+        self.filters = {}
+        for det in HexrdConfig().detector_names:
+            filter = HexrdConfig().detector_filter(det)
+            if filter is None:
+                HexrdConfig().update_detector_filter(det)
+                filter = HexrdConfig().detector_filter(det)
```
psavery, line 20: So, if there isn't a filter, we also wouldn't want to add one unless the user clicks "OK", right? I think we shouldn't modify any settings unless the user clicks "OK".

In this case, if the user clicks "Cancel", would there now be a filter present for that detector?

saransh13:
I still see the filter material going back to the default one when cycling between detectors. There's also a stray print statement somewhere generating this text

```config filter IMAGE-PLATE-D: Cu 2.7 12.5
config filter IMAGE-PLATE-L: Cu 2.7 12.5
config filter IMAGE-PLATE-R: Cu 2.7 12.5
config filter IMAGE-PLATE-U: Cu 2.7 12.5```

https://github.com/user-attachments/assets/17707563-5d63-45c2-a09e-fb7c67bf0238

saransh13 (approved):
Ok, LGTM ready to merge now.

Issue #1821, opened 2025-04-07
Remove detector resets state


saransh13:
When a detector is removed using the form view, the images associated with all detectors are deleted and the configuration is reset. The images associated with the other detectors should not be affected when one detector is removed.

PR #1828, opened 2025-04-10
Keep valid images after detectors added/removed

psavery:
Previously, if detectors were added/removed from the instrument form view, we would automatically clear all the imageseries. We can instead, however, keep the imageseries that are still valid. If one particular detector does not contain good images, this will allow, for example, that detector to just be removed.

The imageseries that are considered "valid" and are kept are only ones whose detector name still exists in the list of detectors, and whose shape matches the detector shape.

Depends on: https://github.com/HEXRD/hexrd/pull/790

Fixes: #1821

### DIFF
--- a/hexrdgui/hexrd_config.py
+++ b/hexrdgui/hexrd_config.py
@@ -869,17 +869,19 @@ class HexrdConfig(QObject, metaclass=QSingleton):
         return len(next(iter(self.imageseries_dict.values())))
 
     @property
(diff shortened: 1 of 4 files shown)


psavery:
@bnmajor Can you let me know if everything (especially the load panel and recent images stuff) looks good to you?

bnmajor (approved):
LGTM!

PR #1829, opened 2025-04-10
Enable pinhole correction for the Cartesian view

psavery:
The panel needs to have the source distance specified on it to use with pinhole correction.

We've never had to use pinhole correction in the Cartesian view (TARDIS and PXRDIP cannot be used with the Cartesian view), so we did not encounter this issue until now.


saransh13 (approved):
LGTM.

Issue #1822, opened 2025-04-07
XRS2 overlay in XRS1


saransh13:
Currently, when in XRS1 view and both XRS1 XRS2 overlays are visible, then the XRS2 overlay is also drawn in the azimuthal average area and vice versa. The overlay for the other view should only be shown in the 2D polar plot and omitted from the azimuthal average plot.

PR #1830, opened 2025-04-16
Only draw lineout powder lines if XRS matches

psavery:
If a powder overlay is only for a specific x-ray source, only draw its mean lines in the lineout if the active x-ray source matches.

Fixes: #1822


saransh13 (approved):
LGTM

PR #1832, opened 2025-04-24
For single XRS, skip XRS-disabling logic

psavery:
When we have a single XRS, that XRS still has a name, and there may be some cases where the name of the single XRS doesn't match the name in the picks.

For this reason, we should always skip any multi-xrs logic when we only have a single xrs. In this specific case, this fixes a bug where the only pick path would be disabled if the name did not match.


saransh13 (approved):
LGTM

Issue #1772, opened 2025-01-08
Issue removing artists in matplotlib 3.10


psavery:
There's some new issue with removing several of our artists starting in matplotlib 3.10 (including overlays, detector borders, the beam marker, and potentially more). It is very common that when `remove()` is called on the artists, that a `NotImplementedError` is raised. The error is as follows:

```python
.../matplotlib/artist.py", line 256, in remove
    raise NotImplementedError('cannot remove artist')
NotImplementedError: cannot remove artist
```

Perhaps this is an issue that will get fixed in matplotlib soon. If not, we'll need to figure out an alternative way to remove the artists.

For now, we are fixing the matplotlib version to be less than 3.10 in #1771

patquem:
Hello,
I experienced the same issue (and fixed it). 
In contrast to earlier versions of Matplotlib, where calling `artist.remove()` after an `ax.clear()` did not raise any problems, recent versions now trigger an error message similar to the one reported here.
To avoid this, ensure that `remove()` is not called after `ax.clear()`.

psavery:
Thanks for the tips, @patquem! Very helpful! I'll try to fix this soon 🙂

PR #1833, opened 2025-04-24
Fix issues we encountered in matplotlib 3.10

psavery:
This will also unpin the matplotlib version as a result.

I did not immediately see any further issues, but if we encounter more in the future, we can hopefully fix them quickly.

Thanks @patquem for the pointer in #1772!

Fixes: #1772

### DIFF
--- a/hexrdgui/blit_manager.py
+++ b/hexrdgui/blit_manager.py
@@ -2,6 +2,8 @@ from collections.abc import Sequence, ValuesView
 
 from matplotlib.artist import Artist
 
+from hexrdgui.utils.matplotlib import remove_artist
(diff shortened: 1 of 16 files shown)


psavery:
@saransh13 @bnmajor Can either of you review? You will need to upgrade matplotlib on your computer.

If we run into more issues, hopefully we can fix them pretty quickly. The changes matplotlib make did result in errors in numerous places.

psavery:
This has been revived. I think we are ready to upgrade matplotlib

saransh13 (approved):
didn't encounter any issues. LGTM

PR #1837, opened 2025-04-25
Fix invalid table selection model

psavery:
I have no idea why, but for *this* specific UI file, the table's selection model's model starts out in an invalid state. I haven't seen it be invalid for other uses of QTableWidget. But if we call `self.ui.table.selectionModel().model()`, which we must do in some functions, we get the following error:

```python
RuntimeError: Internal C++ object (PySide6.QtCore.QAbstractTableModel) already deleted.
```

I tried resetting all QTableWidget settings to default in the ui file, but the model still starts out in an invalid state! So I have no idea why it is happening.

But one fix I found is that if we set the table's selection model on the table (which you'd think wouldn't change anything), the model on the selection model gets fixed. So we are doing that.

### DIFF
--- a/hexrdgui/list_editor.py
+++ b/hexrdgui/list_editor.py
@@ -33,6 +33,22 @@ class ListEditor(QObject):
 
         self.ui = UiLoader().load_file('list_editor.ui', parent)
 
+        # I have no idea why, but for *this* specific UI file, the table's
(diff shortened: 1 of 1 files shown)


saransh13:
I can't reproduce the issue on my Mac :/

saransh13 (approved):
LGTM

PR #1838, opened 2025-04-29
Fix resetting the parameters for WPPF

psavery:
We need to set the `param_dict` object onto the WPPF parameters, not the params itself.


saransh13 (approved):
Thanks for the quick fix. LGTM

Issue #1792, opened 2025-02-26
Add more options to Physics Package


bnmajor:
Include pusher, heatshield, and ablator. They all need the same inputs as the sample.

PR #1842, opened 2025-04-30
Add new layers to physics package

psavery:
This adds layers for the ablator, heatshield, and pusher. The "Sample Layer" pinhole correction was now also generalized to a "Layer" pinhole correction, and the layer type is selected from a dropdown menu. The layer types include the ablator, heatshield, pusher, sample, and window. The layer standoff for each layer is correctly computed so that they may be used for two theta distortion.

This also fixes a bug where if a layer (such as window/pinhole) were disabled in the physics package, it would have an effect on the UI, but it wouldn't have an effect on the internal calculations. Now, if a layer is disabled, its thickness is set to 0 as well.

Depends on: hexrd/hexrd#795

Fixes: #1792


saransh13 (approved):
working great so far! LGTM

Issue #1824, opened 2025-04-08
Filter, phosphor and coating from material list


saransh13:
Currently, the filter, phosphor and detector coatings are only supported as formulas and density. However, the GUI offers the option to specify these as materials from the list of materials. This causes an error during the absorption correction calculation. 

![Image](https://github.com/user-attachments/assets/8ead9dd6-b76a-4b28-9dc2-e5b46875fd12)

The options are to:

1. Only support use of formula and density input
2. Support the use of material input in `hexrd.instrument.detector_coatings.AbstractLayer`

2 is more work and gives more flexibility but 1 is an easy fix and is used in vast majority of the case. I'm leaning toward 1 but would love to hear some thoughts.

PR #1847, opened 2025-05-09
Provide chemical formula for absorption correction

psavery:
For all of the layers (both those on the detector and in the physics package), provide the chemical formula, if possible, to be used for absorption correction.

This enables users to select materials whose name is not a chemical formula, such as "diamond". We extract the chemical formula from the material and provide that.

Depends on: hexrd/hexrd#800

Fixes: #1824


saransh13 (approved):
LGTM

PR #1848, opened 2025-05-09
Allow changing line style and width for mask borders

### LINKED ISSUE #1845: Mask Manager Styling
Extend border options to include border color, line style, and line width for masks.

bnmajor:
Adds options to set mask border line style as well as thickness.

image

Fixes #1845

### DIFF
--- a/hexrdgui/masking/mask_border_style_picker.py
+++ b/hexrdgui/masking/mask_border_style_picker.py
@@ -1,0 +1,60 @@
+from PySide6.QtCore import QObject
+from PySide6.QtGui import QColor
+from PySide6.QtWidgets import QColorDialog
+
+from hexrdgui.ui_loader import UiLoader
+
+
+class MaskBorderStylePicker(QObject):
+
+    def __init__(self, original_color, original_style, original_width, parent=None):
+        super().__init__(parent)
+
+        loader = UiLoader()
+        self.ui = loader.load_file('mask_border_style_picker.ui', parent)
+        self.original_color = original_color
+        self.original_style = original_style
+        self.original_width = original_width
+        self.color = original_color
+        self.style = original_style
+        self.width = original_width
+
+        self.setup_ui()
+        self.setup_connections()
+
+    def setup_ui(self):
+        self.ui.border_color.setText(self.original_color)
+        self.ui.border_color.setStyleSheet('QPushButton {background-color: %s}' % self.original_color)
+        self.ui.border_style.setCurrentText(self.original_style)
+        self.ui.border_size.setValue(self.original_width)
+
+    def exec(self):
+        self.ui.adjustSize()
+        return self.ui.exec()
+
+    def setup_connections(self):
+        self.ui.border_color.clicked.connect(self.pick_color)
+        self.ui.border_style.currentIndexChanged.connect(self.update_style)
+        self.ui.border_size.valueChanged.connect(self.update_width)
+        self.ui.button_box.rejected.connect(self.reject)
+
+    def pick_color(self):
+        dialog = QColorDialog(QColor(self.original_color), self.ui)
+        if dialog.exec():
+            self.color = dialog.selectedColor().name()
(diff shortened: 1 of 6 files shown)


hexrdgui/masking/mask_border_style_picker.py
```diff
@@ -0,0 +1,60 @@
...
+        self.color = original_color
+        self.style = original_style
+        self.width = original_width
+
+        self.setup_ui()
+        self.setup_connections()
+
+    def setup_ui(self):
+        self.ui.border_color.setText(self.original_color)
+        self.ui.border_color.setStyleSheet('QPushButton {background-color: %s}' % self.original_color)
```
psavery, line 27:
```suggestion
        self.ui.border_color.setStyleSheet(f'QPushButton {{background-color: {self.original_color}}}')
```

No need to change this, but in case you were curious, this is how you'd do this with an f-string (double braces get transformed into a literal single brace).

hexrdgui/masking/mask_border_style_picker.py
```diff
@@ -0,0 +1,60 @@
...
+        if dialog.exec():
+            self.color = dialog.selectedColor().name()
+            self.ui.border_color.setText(self.color)
+            self.ui.border_color.setStyleSheet('QPushButton {background-color: %s}' % self.color)
+
+    def update_style(self):
+        self.style = self.ui.border_style.currentText()
+
+    def update_width(self):
+        self.width = self.ui.border_size.value()
```
psavery, line 52: Instead of having extra variables like `self.width` and `self.style`, you could always make them properties that just reflect the UI state. For example:

```python
@property
def width(self) -> float:
    return self.ui.border_size.value()
```

Then the variables can be eliminated. `reject()` could call `self.setup_ui()` to restore them to their original states.

Also, for the caller code, I do think the pattern:

```python
if dialog.exec():
    # Do whatever is needed on acceptance
```

is a nicer pattern. But you don't necessarily need to change anything.

hexrdgui/masking/mask_manager.py
```diff
@@ -385,8 +392,8 @@ def load_masks(self, h5py_group):
         actual_view_mode = self.view_mode
         self.view_mode = ViewType.raw
         for key, data in items.items():
-            if key == '__boundary_color':
-                self.boundary_color = data
+            if key.startswith('__boundary_'):
+                setattr(self, key.split('__', 1)[1], data)
```
psavery, line 396: Nice!

psavery (approved):
LGTM

hexrdgui/masking/mask_border_style_picker.py
bnmajor: > Instead of having extra variables like `self.width` and `self.style`, you could always make them properties that just reflect the UI state.

Very nice, much cleaner! Changes made :+1: 

> Also, for the caller code, I do think the pattern:
> 
> ```python
> if dialog.exec():
>     # Do whatever is needed on acceptance
> ```
> 
> is a nicer pattern. But you don't necessarily need to change anything.

Easy fix, cleaned up now!

hexrdgui/masking/mask_manager_dialog.py
```diff
@@ -408,10 +409,17 @@ def show_all_boundaries(self):
...
+            MaskManager().boundary_color,
+            MaskManager().boundary_style,
+            MaskManager().boundary_width
+        )
         if dialog.exec():
-            MaskManager().boundary_color = dialog.selectedColor().name()
+            color, style, width = dialog.result()
+            MaskManager().boundary_color = color
+            MaskManager().boundary_style = style
+            MaskManager().boundary_width = width
```
psavery, line 422: So actually, since you now only set these if the dialog is accepted (the return of `dialog.exec()` is `True`), I think you don't even need to do the internal connection to the `reject` function since the values won't be modified if it is rejected. But it's not really an important change.

PR #1849, opened 2025-05-09
Multi select masks

### LINKED ISSUE #1846: Allow multi-select in mask manager
- [ ] Users should be able to select some subset of masks (currently restricted to one selection)
- [ ] Shortcut to select all masks under a current group (relies on #1815)
- [ ] Action performed on any selected mask (export, change presentation, etc) should be applied to all currently selected masks
- [ ] Add keyboard shortcuts (ctrl+a to select all, etc)

bnmajor:
This PR:
- Rearranges the buttons to try to provide a more intuitive layout
- Add support for multi-select in the tree view (CTRL + left-click to select individual items, SHIFT + left-click at start and SHIFT + left-click at end to select all items in between)
- Can use CTRL + a to select all
- Allows changing the presentation of multiple masks at once
    - If more than one row is selected the combobox on the right will be enabled
    - If one or more of the selected masks are `powder`, `laue`, or `threshold` the options will be limited to `None` or `Visible`. Otherwise the `Boundary only` and `Visible + Boundary` will also be options.
    - The changes are applied to all selected masks immediately
- Adds a button to export selected masks
    - This is functionally the same as selecting multiple rows and using the context menu to export, just provides the option of a button if that is preferred over right-click
    - The button is only enabled when more than one row is selected
- Adds the option to remove multiple masks at a time

image

- [x] Relies on #1815, #1848

Fixes #1846

### DIFF
--- a/hexrdgui/masking/mask_manager_dialog.py
+++ b/hexrdgui/masking/mask_manager_dialog.py
@@ -53,7 +53,7 @@ class MaskManagerDialog(QObject):
         self.ui.masks_tree.itemChanged.connect(self.update_mask_name)
         self.ui.masks_tree.customContextMenuRequested.connect(
             self.context_menu_event)
-        self.ui.export_masks.clicked.connect(MaskManager().write_all_masks)
+        self.ui.export_masks.clicked.connect(MaskManager().write_masks)
         self.ui.import_masks.clicked.connect(self.import_masks)
         self.ui.panel_buffer.clicked.connect(self.masks_to_panel_buffer)
         self.ui.view_masks.clicked.connect(self.show_masks)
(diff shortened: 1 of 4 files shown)


hexrdgui/masking/mask_manager_dialog.py
```diff
@@ -427,3 +435,49 @@ def apply_changes(self):
             self.change_mask_presentation(index, name)
         self.changed_masks = {}
         self.ui.apply_changes.setEnabled(False)
+
+    def selected_changed(self):
+        selected = self.ui.masks_tree.selectedItems()
+        self.ui.presentation_selector.setEnabled(len(selected) > 1)
+        self.ui.export_selected.setEnabled(len(selected) > 1)
+        self.ui.remove_selected.setEnabled(len(selected) > 1)
```
psavery, line 443: I would typically do something like this:

```python
        enable = len(selected) > 1
        widgets = [
            self.ui.presentation_selector,
            self.ui.export_selected,
            self.ui.remove_selected,
        ]
        for w in widgets:
            w.setEnabled(enable)
```

But you do not have to change it!

psavery (commented):
I wonder if we can just get rid of the "Remove Mask" buttons on each row, and rely on the "Remove Selected" instead. What do you think, @saransh13?

Overall, the code changes look good to me! Let us know what you think, Saransh.

saransh13 (approved):
I'm approving the PR but will note that the `mask to panel buffer` is not working as expected. Since this is also a problem in master, we can fix that issue in a follow up PR.

PR #1852, opened 2025-05-14
Use raw ATLAS coordinates for FIDDLE import

saransh13:
Instead of importing the corners of the sensors, use the raw ATLAS coordinates to determine the coordinate transforms connecting the CMM and TCC frames. Rest of the workflow remains the same.


hexrdgui/llnl_import_tool_dialog.py
```diff
@@ -38,17 +42,83 @@
...
+        result = res.minimize(method='least_squares', params=params)
+
+        if result.chisqr > 1:
+            '''please include a message box here with this message
+            '''
+            msg = (f'least-squares did not find a good solution. '
+                    f'Double check data to make sure input is in order.')
+            print(msg)
+
+        self.result = result
```
bnmajor, line 100:
```suggestion
            msg = (
                f'least-squares did not find a good solution. '
                f'Double check data to make sure input is in order.'
            )
            QMessageBox.warning(None, 'HEXRD', msg)

        self.result = result
```

bnmajor (approved):
LGTM, appears to work well!

PR #1853, opened 2025-05-14
Utilize instrument rigid body params for sliders

psavery:
For locking relative tilts/translations in an "Instrument Rigid Body" fashion, we used to show the parameters for the currently selected detector and update them.

Now, we utilize the internally stored instrument rigid body parameters (the same as those used in the calibration workflows). This means we can now switch back and forth between the calibration workflow and the slider widget and see the same parameters when the transformations all have an "Instrument Rigid Body" lock.


saransh13 (approved):
LGTM

Issue #1794, opened 2025-02-27
Enable widths automatically for composite calibration


psavery:
In the fast powder calibration, widths are enabled automatically if they are disabled.

We should likewise enable widths (both for powder and laue) automatically if they are disabled in composite calibration.

PR #1855, opened 2025-05-16
Automatically enable widths for calibration

psavery:
Similar to fast powder, in the composite calibration, automatically enable widths if there are any overlays which do not have widths enabled.

Fixes: #1794


saransh13 (approved):
LGTM

Issue #1719, opened 2024-07-29
Detector borders.


saransh13:
Detector borders do not move when the instrument `tvec` is changed. See video below:

https://github.com/user-attachments/assets/eb267571-f9f3-4881-984a-e29e0531f825

PR #1856, opened 2025-05-16
Take into account sample tvec for detector borders

psavery:
For the polar and stereo views, take into account the sample tvec when generating their detector borders.

This change is not necessary for the raw view (no detector borders) and the Cartesian view (where the detector borders should not change).

Fixes: #1719


saransh13 (commented):
LGTM

saransh13 (approved):

PR #1857, opened 2025-05-20
Memoize projection to polar view

psavery:
This function was memoized in the past, and appears to have been "un-memoized" by accident. After the polar view has been generated one time, this memoization substantially improves the performance of re-generating the same polar view using the same settings (for example, switching to a different image mode and then coming back).


saransh13 (approved):
LGTM. However, I'm not noticing any performance difference while using the GUI

psavery:
Only when you switch tabs away from the polar view and come back. And probably only if it's a high resolution one.

Issue #1861, opened 2025-05-22
Get error message when "Hand picking" points


saransh13:
When editing points in the calibration menu, and `Hand picking` an additional point, the following error message is generated. The new picked point does show up on the canvas and in the picked points tree view though.

```TypeError: dict_modified() only accepts 0 argument(s), 1 given!
TypeError: dict_modified() only accepts 0 argument(s), 1 given!```

PR #1862, opened 2025-05-26
Update `dict_modified()` arguments

psavery:
These were changed recently to use a `QModelIndex` argument in their signal.

Fixes: #1861


saransh13 (approved):
LGTM

PR #1863, opened 2025-05-27
Use a vertical scroll area for the polar tab

psavery:
The polar tab is becoming a bit too large vertically for Mac computers at low resolution. It is causing the main window to extend below the display.

Use a vertical scroll area for the polar tab to fix this.

### DIFF
--- a/hexrdgui/image_mode_widget.py
+++ b/hexrdgui/image_mode_widget.py
@@ -2,7 +2,8 @@ from functools import partial
 import multiprocessing
 import numpy as np
 
-from PySide6.QtCore import QObject, QTimer, Signal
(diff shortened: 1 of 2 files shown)


psavery:
@saransh13 Let me know how this looks for you. This made the main window fit within my Mac laptop screen (so it's better for laptops I think)

saransh13:
I think it looks good for the laptop, but on a big screen, the window size is artificially cut, even though there is plenty of gray space (see below). It is possible to resize the window size automatically to use the empty gray space?

Screenshot 2025-06-05 at 11 19 04 AM

psavery:
Yes there's probably a way. I'll look into this!

psavery:
@saransh13 Can you try again? I think this is ready.

Issue #1878, opened 2025-06-06
Issue with log-log-root colorscale


akrygier:
The log-log-root colorscale is changing (the image is displaying different) when I toggle "apply erosion" on or off.  It should not do this.

Here is an example where the only think I did was toggle apply erosion.

apply erosion on:

![Image](https://github.com/user-attachments/assets/d85cfb25-7b8c-4b0f-b315-71937d26a3eb)

apply erosion off:

![Image](https://github.com/user-attachments/assets/101cafec-350f-475c-9faf-87cdc218ca4c)

Also, I am not sure it's implemented correctly...

here is my python code:
def llr(m):
    return np.log(np.log(np.sqrt(m+1)+1)+1)

when m = 0, llr(0) = 0.526... but I don't think that's consistent with what is shown (though it seems to compress the image colorscale pretty effectively)

akrygier:
Patrick, I don't think resetting the colorscale this way is a good approach since it means lineouts from different shots or even on a single shot with different settings cannot be compared with this colorscale. In practice, it also has the effect of changing the colorscale around everytime the calibration is changed or snip is changed, etc. 

Maybe it's worth just keeping the lineout as a linear scale when plotting the log log root and nan-ing anything negative?

Also, I think you're missing + 1 inside the sqrt.  The extra +1 takes care of some of the negatives that you are trying to avoid by subtracting the min. Here's a screenshot from one the Morhac papers that came from:

<img width="305" alt="Image" src="https://github.com/user-attachments/assets/2f39b161-aae6-4414-93f9-e0ff339dc43d" />

psavery:
I think nan-ing negative numbers sounds fine to me! What do you think, @saransh13?

Yeah, we could add that `+ 1` in there too.

psavery:
@akrygier Yeah, there's a checkbox in the polar view settings for applying the scaling to the lineout. Otherwise, it only affects the 2D image.

PR #1882, opened 2025-06-11
Remove min subtraction in scaling

psavery:
The min subtraction was causing some inconsistencies in coloring within the same dataset, because if someone, for example, masks out (or performs something like an erosion on) the data which eliminates the old minimum in the data, it would appear that all of the data values are rescaled.

We also should include a `+ 1` for the square root cases that will eliminate most negative values we'd encounter.

The only case where the data may be adversely affected by this change would be cases where there are some very negative values in the data. Perhaps this happens for some datasets with dark background subtraction.

Fixes: #1878


saransh13 (approved):
LGTM

PR #1885, opened 2025-06-20
Add beam energy correction to grains

psavery:
This allows a slope (in eV/mm), intercept (in mm), and axis ('x' or 'y') to be specified for a beam energy correction according to grain positions.

This is important to take into consideration gradients in the beam energy profile, such that some grains may "experience" a different beam energy than others.

The option to add the HEDM beam energy correction was added to the instrument form view. If added, it also goes into the tree view, and the slope and the intercept will appear in the HEDM calibration as well as refinable parameters.

Depends on: hexrd/hexrd#819


bnmajor (approved):
LGTM!

psavery:
Here's an example video of this in action:

https://github.com/user-attachments/assets/4c86ef54-45fd-412b-9b98-f7f5aa4686bd

Issue #1881, opened 2025-06-11
WPPF Liquid Diffraction


psavery:
Add options in WPPF to utilize the new liquid (amorphous) diffraction analysis.

There will be an option to choose between a split Gaussian fit, loading an experimental file, or (in the future) a pseudo-voigt fit. New parameters will appear in the lmfit parameters table.

"Degree of Crystallinity" is a single value that should be shown in the WPPF options dialog when this is enabled. This is a property on the WPPF classes.

PR #1889, opened 2025-07-11
Add amorphous model support to WPPF

psavery:
I haven't fully tested everything in here. I think we need a nice test case for the GUI. But it seems to work in my testing so far.

Depends on: hexrd/hexrd#823

Fixes: #1881


saransh13 (approved):
Approving pending passing checks after the `hexrd/wppf-amorphous-fixes` branch is merged.

Issue #1892, opened 2025-07-23
Add option in WPPF dialog to plot the background, and the amorphous model (if present) in the lineouts


saransh13:
Add check boxes to plot the 

1. background (green curve below)
2. amorphous model (red curve below)

in addition to the overall calculated spectrum.

The background is present in `LeBail.background` or `Rietveld.background`. The amorphous lineout data is present in `LeBail.amorphous_model.amorphous_lineout` or `Rietveld.amorphous_model.amorphous_lineout`.

<img width="1090" height="552" alt="Image" src="https://github.com/user-attachments/assets/3db14683-cd40-4c04-97d8-ff7ea0548b15" />

PR #1895, opened 2025-08-07
Add ability to plot WPPF background and amorphous

psavery:
This adds a couple of checkboxes to the WPPF options dialog to plot the background and amorphous on the azimuthal lineout in the polar view.

There are also options to edit the styles of these plots within "Edit Plot Style".

Fixes: #1892


saransh13 (changes requested):
Do you mind changing the default marker to `x` and the default face color to something other than white (maybe blue)? For the `x` marker, the white color is the same as the background so the markers look like they disappear. Looks good otherwise.

Screenshot 2025-08-11 at 9 40 18 AM

PR #1898, opened 2025-08-11
Allow multiple amorphous phases in WPPF

psavery:
This adds the ability to use multiple amorphous phases in WPPF.

If more than one amorphous phase is specified, they will appear in the tree view with their phase names, such as `c1`, `c2`, etc.

Multiple experimental files are also supported to use with the multiple phases, and file dialogs appear for each phase when "Experimental" is selected.

This PR also fixes the issue where, after running WPPF, the parameter tree view scroll bar would reset its position.


saransh13 (approved):
LGTM

PR #1911, opened 2025-09-04
Invalidate polar masks on detector transform

psavery:
The polar masks need to be regenerated when the detectors are transformed. This adds functionality that does so.


bnmajor (approved):
LGTM!

PR #1912, opened 2025-09-04
Keep waterfall plot in front of the application

psavery:
Also make a small improvements to appearance of physics package manager layout.


saransh13 (approved):
Approving for new release

PR #1915, opened 2025-09-18
Move blit manager update out of mask highlights

psavery:
This is so that we can better control it and ensure it is only called once, when needed.

This fixes an issue where the old view would "flash" on the screen when you switch to any view except raw.


saransh13 (approved):
For new release

Issue #1926, opened 2025-10-27
WPPF Improvements


psavery:
1. Make an advanced option, allowing us to change the parameters to the intensity extraction function
2. Always apply masks to binned/simulated polar view 2D images
3. Make a way to reset all texture parameters to zero

saransh13:
Another one is to use units of Angstrom (A) in the WPPF module. For legacy reasons, internally, everything is in units of nm. However, we should expose the numbers in units of A to the user. Relevant parameters where this would show up:
1. Lattice constants
2. `S_hkl` (anisotropic peak broadening parameters)

saransh13:
- [ ] And finally, since the package here is based on the GSAS convention, instead of displaying `X`, '`Y`, and `U` values, we can display an equivalent size and microstrain. See below for the conversion. These numbers are much more physically meaningful than what is displayed currently. However, internally we could use the same code.

<img width="730" height="166" alt="Image" src="https://github.com/user-attachments/assets/da7a5ab1-6dd9-424d-b88f-a892c112738e" />
<img width="536" height="435" alt="Image" src="https://github.com/user-attachments/assets/0e2054ad-4f13-4ab6-90b8-dc15afd56065" />

psavery:
Comments about the equations above:

1. We use `K = 0.91` instead of `K = 1`
2. We will only use the `S` on the left, not the one on the right
3. `LX` maps to `material_X`, `LY` maps to `material_Y`, and `GP` is `material_P`.
4. `S` is `Microstrain`, the first `p` is `Lorentzian Particle Size`, and the second `p` is `Gaussian Particle Size`.

psavery:
Add units in parentheses.

PR #1933, opened 2025-11-07
Add units and conversion capabilities to WPPF

psavery:
This adds the basic capabilities for units and unit conversion to the WPPF module, and a few initial examples, including for lattice parameters.

The lattice parameter lengths are automatically converted from nanometers to Angstroms for display (and back when the user modifies them).

It also adds Angstrom units to the lattice lengths, and degree units to the lattice angles.

We'll use this same infrastructure for other unit conversions.

Addresses part of: https://github.com/HEXRD/hexrdgui/issues/1926


saransh13:
the last change is to use scientific notation for the `Shkl` parameters. Other than that all numbers look good to me.

PR #1935, opened 2025-12-01
Autoscale WPPF diff y-axis if percent is set

psavery:
If a percentage difference is being used, autoscale the WPPF difference plot y-axis.


saransh13 (approved):
LGTM

PR #1936, opened 2025-12-02
Allow cmap scaling changes when image not ready

psavery:
Previously, error messages would appear if the cmap scaling was changed while the image was still being generated (such as the polar view being generated).

Now, error messages are not generated, and the cmap scaling change is handled gracefully.


saransh13 (approved):
lgtm

PR #1941, opened 2025-12-05
Ensure all nan-slice warnings are caught

psavery:
They were not always being caught with the previous setup, but this catches them all.


psavery:
Please let me know if this issue is now gone after all of the latest updates. You can try both with this branch and without this branch. If it still persists... I'll look into it more.

PR #1946, opened 2025-12-10
Cif export

bnmajor:


hexrdgui/main_window.py
```diff
@@ -604,10 +606,14 @@ def images_loaded(self, enabled=True):
     def on_action_open_materials_triggered(self):
         selected_file, selected_filter = QFileDialog.getOpenFileName(
             self.ui, 'Load Materials File', HexrdConfig().working_dir,
-            'HDF5 files (*.h5 *.hdf5)')
+            'HDF5 files (*.h5 *.hdf5), CIF (*.cif)')
+        if not selected_file:
```
bnmajor: @psavery Currently only one CIF file can be opened at a time through the file menu since we only allow one HDF5 file to be selected at a time. Multiple CIF files can be imported at once from the material list editor. Let me know if we should adjust this behavior.

psavery:
Code looks good to me. Can you review, @saransh13?

saransh13:
The other thing that should be fixed is that the apple export menu says `open` instead of `save`. 

image

hexrdgui/material_list_editor.py
```diff
@@ -67,6 +68,30 @@ def item_added(self, new_name):
...
         HexrdConfig().add_material(new_name, material)
 
+    def export_to_cif(self):
+        selections = self.editor.selected_items
+        if len(selections) == 1:
+            caption = 'Save CIF file as'
+            selected_file = QFileDialog.getSaveFileName(
+                self.ui, caption, HexrdConfig().working_dir,
+                'CIF files (*.cif)')
+            selected_dir = str(Path(selected_file).parent)
```
psavery: This looks like it could "trick" users a little bit. Like, they could specify a cif file name like "my_new_mat.cif", but then we don't end up using their cif file name, but just use the directory and name the file `f'{mat_name}.cif`.

I think we should either have them choose a directory (and NOT choose the CIF file name) so it is clear that they don't get to pick the CIF file name. Or, use whatever name they pick, and don't rename it.

Feel free to do the first option if that's easier for now.

hexrdgui/hexrd_config.py
```diff
@@ -1289,9 +1289,16 @@ def load_materials(self, f):
         beam_energy = valWUnit('beam', 'energy', self.beam_energy, 'keV')
         self.materials = load_materials_hdf5(f, kev=beam_energy)
 
-    def save_materials(self, f, path=None):
+    def save_materials_hdf5(self, f, path=None):
         save_materials_hdf5(f, self.materials, path)
 
+    def save_material_cif(self, material, directory=None, filename=None):
```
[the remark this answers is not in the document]
psavery: I thought the second option was better too :slightly_smiling_face:

saransh13:
@bnmajor There is one change and one enhancement in this PR. The enhancement can be a separate PR.
For the change, there is still `All blocks output.` printed to the terminal and in the message box. Can you suppress that. I will approve and you can merge after this change.

For the enhancement (future PR), we can 

 - get rid of the indent here:

image

- include the `_atom_site_U_iso` or `_atom_site_U_aniso` key in the loop. This will write the mean-squared atom displacement or the Debye-Waller factor.

psavery (approved):

Issue #1950, opened 2026-01-10
on import of instrument files, only .yml files are shown, not .yaml


donald-e-boyce:
When opening an instrument file, the window only allows the user to select files with a `.yml` suffix. The `.yaml` suffix is also used and slightly preferred, according to the [yaml.org FAQ](https://yaml.org/faq.html).

PR #1951, opened 2026-01-12
Support .yaml suffix when loading instrument files

psavery:
The .yaml suffix is actually preferred over the .yml suffix. Both should be supported.

Fixes: #1950


donald-e-boyce (approved):
Looks good, and thanks. 

By the way and FYI, I made (on osx) symbolic links to the `.yaml` files but with the `.yml` suffices, but I still wasn't able to select them.

PR #1980, opened 2026-03-10
Add dialog for HEDM calibration diagnostics

psavery:
This dialog provides several plots indicating how good of a fit the calibration is.

It is a new "Spot Diagnostics" button in the HEDM calibration dialog.

It is also available in the HEDM workflow, after fit-grains has finished.

The dialog appears like the following:

image

Depends on: hexrd/hexrd#905


bnmajor (approved):
Looks great to me!

PR #1986, opened 2026-03-24
Skip multiplication for layer type

psavery:
The layer type is a string and should not be multiplied.

Originally, the multiplication worked fine, because the multiplier was an integer of 1, and doing things like `'sample' * 1` results in `'sample'` - just the string.

However, we changed the multiplier to a float in the type annotations PR, and we can't multiply a float with a string, so we would get an error.

This fixes the code so the layer type string does not get multiplied at all.


saransh13 (approved):
LGTM

PR #1988, opened 2026-03-25
Perform gap detection for stereo view

psavery:
We were previously getting long overlay lines connecting one end of a powder overlay with the other end. These lines shouldn't exist.

We needed to perform gap detection in the stereo view that is similar to the gap detection we've been performing in the polar view. This fixes the issue.


saransh13 (approved):
lgtm

PR #1989, opened 2026-03-26
Restore panel subset properly when loading state

psavery:
We were accidentally erasing the panel subset setting while loading the state file.

Don't do that so that the panel subset gets saved/restored properly.


psavery:
Received verbal approval from @saransh13

PR #1991, opened 2026-03-27
Fix several issues related to imageseries

### LINKED ISSUE #1990: Main plot window resizing when using Simple Image series
The main plot window can not be resized freely to a smaller size when a simple image series is loaded. This behavior is unlike the regular plot window which can be freely resized

psavery:
1. The image series slider at the bottom no longer has a fixed minimum width. Previously, this could make the GUI take up more horizontal space than is available on a monitor, and would not allow you to resize. It is better if it just takes up the amount of available width, and allow the window to be resized, rather than forcing a minimum width.

2. The omage label now only takes up the maximum amount of space that it could take up for the particular imageseries that was loaded. Previously, it would take up potentially more space than would be needed. Now, we iterate through the possibilities and determine the max space.

3. Render flash in the raw view when changing image series frames is now fixed. It was only present for the raw view because the raw view is loaded synchronously in the main thread (as opposed to the other views which are loaded in a background thread).

4. Remove the `new_images_loaded` connection in the rotation series overlays and instead call the relevant function from the main window. Since the rotation series overlays are not QObjects, they would not disconnect properly when they were deleted. Iterating through them in the main window fixes this issue.

Fixes: #1990

### DIFF
--- a/hexrdgui/image_series_toolbar.py
+++ b/hexrdgui/image_series_toolbar.py
@@ -126,13 +126,6 @@ class ImageSeriesToolbar(QWidget):
         self.omega_label = QLabel(self._parent_widget())
         self.omega_label.setVisible(False)
 
-        # Compute the text width for the maximum size label we will have
-        # with the current font, and set the label width to be fixed at
-        # this width. This will prevent the slider from shifting around
-        # while we are sliding.
-        example_label_text = omega_label_text(359.999, 359.999)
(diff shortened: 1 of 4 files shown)


psavery:
@saransh13 I'm pretty sure this should fix #1990. Can you verify?

psavery:
This is important for fixing some CI issues in other PRs. Merging now...

PR #1998, opened 2026-04-09
Subtract minimum before all scaling operations

psavery:
We are reverting back the change from #1882, but we are instead
keeping a record of the minimum of the image before masking, and
always using that minimum in the subtraction.

The scaling looks better, and the images look nicer, after this
change. Keeping the recorded minimum also prevents the "jumping"
in scaling that occurred when masking out the minimum pixel. So
the issue that #1882 was trying to fix is now resolved too.


saransh13 (approved):
The contrast looks much nicer and no saturation or scaling issues when brightest spots are masked. LGTM!

PR #2000, opened 2026-04-13
Fix mypy errors from the CI

psavery:
I think these are just caused by a recent update to mypy...


psavery (commented):
These are just typing fixes for the linter. Merging.

PR #2003, opened 2026-04-16
Minimize required updates when raw tabs modified

psavery:
Previously, every raw tab would be re-rendered when a change was made, such as drawing a mask or switching detectors. This could be very slow for many detectors.

Now, only the stale raw tab (the one modified) gets updated. This is a faster update path when a rerender from scratch is not necessary.


saransh13 (commented):
lgtm

PR #2004, opened 2026-04-16
Add option to combine panel buffers

psavery:
When a new panel buffer is set on a detector, if an old panel buffer already existed, the user now has the option to:

1. Replace the old panel buffer entirely
2. Combine the two panel buffers (as an "OR" of the invalid regions)
3. Combine the two panel buffers (as an "AND" of the invalid regions)

This allows the user to, for example, take an existing panel buffer, and then add a border to it.


saransh13 (approved):
LGTM!

PR #2005, opened 2026-04-20
Speed up loading significantly in raw tabbed view

psavery:
Before, all detector images would be loaded, and all overlays computed, for every single tab in the tabbed view.

Now, all detector images are only loaded once, and all overlays are computed only once, for all tabs. It uses a new caching parameter to do this.


saransh13 (approved):
Great improvement in rendering speed. LGTM!

PR #2009, opened 2026-05-06
Fix a few small errors

psavery:
* Create coatings if missing during deserialization. If they are missing, we just just automatically create them. This makes for more robust deserialization.
* Add a shape guard when reapplying masks. If the shape doesn't match, then a full update is about to be performed anyways, and we should bail before we encounter an error.


saransh13 (approved):
cleared the errors for me! LGTM

PR #2010, opened 2026-05-13
Fix numba warning on Linux for using TBB

psavery:
This fixes the following numba warning on Linux:

```
Numba: Attempted to fork from a non-main thread, the TBB library may be in an invalid state in the child process.
```

The main idea is that forked subprocesses (which we indeed do, in order to take advantage of copy-on-write mechanics and let subprocesses share data), which only happen on Linux (Windows and Mac do "spawn"), would always print this numba warning, even though we were never actually using TBB in an invalid way.

There's no way to suppress this warning, so we switch to OpenMP on Linux, which is just as performant.

In fact, it has one better feature too: it will crash if we try to run numba in parallel mode on a forked subprocess. TBB was just undefined, but the OpenMP implementation will actually crash, which prevents developers from ever doing that.

### DIFF
--- a/hexrdgui/main.py
+++ b/hexrdgui/main.py
@@ -14,6 +14,19 @@ if sys.platform.startswith('darwin'):
     # Prevent crashing when using OpenBLAS
     os.environ['OPENBLAS_NUM_THREADS'] = '1'
 
+if sys.platform == 'linux':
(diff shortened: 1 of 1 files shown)


saransh13 (approved):
LGTM

PR #2011, opened 2026-05-13
Hide harmless loky warning

psavery:
Here is the warning, which would occur when closing the application:

```python
/usr/local/lib/python3.14/multiprocessing/resource_tracker.py:396: UserWarning: resource_tracker: There appear to be 2 leaked semaphore objects to clean up at shutdown: {'/loky-853470-5i9c59nk', '/loky-853470-v_q7ann3'}
```

This also moves the logic for ignoring warnings into an `ignore_warnings.py` file.


saransh13 (approved):
LGTM.

PR #2012, opened 2026-05-18
Fix polar mask conversion near beam center

psavery:
The old approach converted polygon perimeter vertices from raw to polar coordinates and rasterized in polar space. When a mask edge crossed near the beam center, the eta coordinate swung ~180 degrees, inflating the polygon into a huge triangle. The new approach rasterizes the polygon in raw pixel space first, then warps the binary mask to polar using the same cached coordinate mapping as image warping. Also scales the project_on_detector memoize cache to the number of detectors.

This new method may be very slightly slower, but will produce accurate results around the beam center.


saransh13 (approved):
Fixes the bug in the test case. LGTM!

PR #2015, opened 2026-05-20
Suppress omp_set_nested routine deprecated warning

psavery:
There is no way to remove this warning until it is fixed upstream in numba.


saransh13 (approved):
lgtm

PR #2020, opened 2026-06-02
Use aspect='auto' for redraw after removing mask

### LINKED ISSUE #2019: Fix main canvas resizing
When a mask is removed, the main canvas gets automatically resized. 

https://github.com/user-attachments/assets/ca3df5e6-bf2d-4a2a-8fd2-a2af24701d70

psavery:
The previous behavior was messing up the image extents

Fixes: #2019

### DIFF
--- a/hexrdgui/blit_manager.py
+++ b/hexrdgui/blit_manager.py
@@ -2,6 +2,7 @@ from collections.abc import Generator, Sequence, ValuesView
 from typing import Any
 
 from matplotlib.artist import Artist
+from matplotlib.image import AxesImage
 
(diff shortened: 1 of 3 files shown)


saransh13 (changes requested):
The shaded mask region should be flushed after the mask is deselected from the list. Other than that, the resizing problem is fixed. should be ready to merge after the suggested change.

https://github.com/user-attachments/assets/e5f400e7-f691-4fee-adf0-b6f6c346a4d3

saransh13 (approved):
All comments are addresses. LGTM!

PR #2021, opened 2026-06-02
Update simulate_laue function call

psavery:
The path was updated in hexrd in: https://github.com/HEXRD/hexrd/pull/911

Update it in hexrdgui as well to reflect the new path.


saransh13 (approved):
LGTM

PR #2024, opened 2026-06-04
Apply threshold on raw images before polar warp

### LINKED ISSUE #1689: Apply threshold mask before warping to polar view
We should apply the threshold mask before warping to the polar view.

We should start off by trying out `np.nan`. But we need to see what this does to the SNIP algorithms, to see if it is okay.

psavery:
Before, the threshold mask for the polar view was applied *after* the bilinear interpolation in the polar warp. This meant that invalid values could easily bleed into other polar values and not be masked out.

Now, the threshold mask is applied on the raw images, *before* the bilinear interpolation. Thus, users can select raw values to threshold out, and reason about the thresholded values much more easily. Any polar pixels affected by a thresholded pixel will be nan, now, so the masking is more aggressive.

This also added many thorough tests to verify the behavior is as intended. In the tests, we generate a polar view, create a few masks (including hand-drawn masks and a threshold mask), and apply different SNIP background methods. In every test step, we verify the polar view appears as expected.

Fixes: #1689


saransh13 (approved):
changes look good to me.

PR #2025, opened 2026-06-10
Add new pvheating peakshape options to GUI

psavery:
This corresponds to the newly added options in hexrd/hexrd#944


saransh13 (approved):
LGTM. Lets run the tests once the herd branch is merged before merging these changes.

psavery:
I have to fix something fundamental to the hexrdgui testing before I can get the CI to pass (there are issues with teardown/setup in between tests that do not matter for a real user running the app), but I have a good strategy.

psavery:
We had to add some automatic disconnect logic for certain classes in order for the setup/teardown in between tests to work properly. That's in place now so it should hopefully work, and this will hopefully make the GUI tests more robust overall.

PR #2027, opened 2026-06-13
Pin matplotlib to be less than 3.11 temporarily

psavery:
Until the upstream issue is resolved that is creating some kind of deadlock with Qt.


psavery:
This is urgent for the CI to pass on other PRs...

PR #2028, opened 2026-06-13
Resize eta omega maps dialog to fit display

### LINKED ISSUE #2026: eta-omega map window size issues
I've been seeing this manifest on and off in various versions of the GUI, and sometimes differently depending on how the GUI is being rendered (e.g. over NoMachine remote desktop connection to a Linux session versus on a local machine), but it can sometimes be very disruptive:

The window that pops up during the first stage of find-orientations in the GUI does not have the same flexibility for resizing the window as other parts of HEXRDGUI do. Especially over remote connections, this window almost always needs to be resized as it pops up larger than the available display size. Unlike every other window in the GUI, it is sometimes not possible to resize this window by grabbing and dragging the top or size edges of the window (marked in red here):

Image

It was possible to right-click near the top of the window, select "move", and drag the window so the bottom edge is visible. Sometimes, I am able to resize the window by grabbing and dragging the lower right-hand corner, but that isn't an option in my current install + current session, for whatever reason. Additionally, if I accidentally move the window so that only the lower left corner is visible (below), I have no way to resize or move the window again:

Image

Wondering if something can be changed in the GUI source code to ensure that this window is reliably resizable.

psavery:
If the eta omega maps dialog is larger than the display, resize it automatically so it fits within the display.

This was important because in some cases, especially remote desktops, the dialog would be too large, and the resize handle would be unreachable, which made it impossible to use.

This also adds a few tests for the relevant resize testing code.

Fixes: #2026

### DIFF
--- a/hexrdgui/indexing/ome_maps_viewer_dialog.py
+++ b/hexrdgui/indexing/ome_maps_viewer_dialog.py
@@ -39,7 +39,7 @@ from hexrdgui.scaling import SCALING_OPTIONS
 from hexrdgui.select_items_widget import SelectItemsWidget
 from hexrdgui.ui_loader import UiLoader
 from hexrdgui.utils import block_signals
-from hexrdgui.utils.dialog import add_help_url
+from hexrdgui.utils.dialog import add_help_url, fit_window_to_screen
 from hexrdgui.utils.matplotlib import remove_artist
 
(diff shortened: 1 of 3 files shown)


psavery:
Merging, as this is somewhat urgent...

PR #2029, opened 2026-06-15
Invalidate polar cache for panel distortion params

psavery:
Previously, the polar cache would not be updated when panel distortion parameters were changed (such as is done during calibration refinement). This meant the polar view would not be fully updated.

We need to add the distortion parameters as kwargs so that the polar view cache would be invalidated automatically.

This also triggers a rerender when users change distortion parameters.


saransh13 (approved):
Fixes the issue. LGTM

PR #2032, opened 2026-06-15
Allow fast powder asymmetry params manual set

psavery:
There is also a button to automatically load these from WPPF if available.

Depends on: hexrd/hexrd#947


saransh13 (approved):
LGTM

PR #2034, opened 2026-06-19
Fix stitched raw view overlays issue

psavery:
Fix an error that would occur when viewing overlays in the stitched raw view.

It now defaults to -1 if nothing was accumulated yet.


psavery:
Merging, as this is somewhat urgent...

PR #2042, opened 2026-07-02
Pin away from conda 26.5.3 in the packaging env

psavery:
conda 26.5.3 ships bin/conda with a "#!/usr/bin/env python" shebang. When package.py runs conda-build, the host env (no conda) shadows python on PATH, so conda runs under the wrong interpreter and fails with "No module named 'conda'", failing the Linux CPack build.


psavery:
Merging as this is somewhat urgent to fix the packaging.

PR #2045, opened 2026-07-08
Prevent recursion crash on NaN image data

psavery:
Images that are fully NaN (e.g. dark subtraction with a NaN dark file) made the color map bounds NaN, and NaN spin box values recurse infinitely since NaN != NaN re-emits valueChanged on every clamp. Guard the spin boxes, fall back to a finite range, and keep one bad detector from turning the global scaling minimum into NaN.


saransh13 (approved):
I've lost the stack trace and the recipe to reproduce this so unable to test this. I'm sure that the `np.nan` and `np.inf` checks won't break anything so approving

PR #2049, opened 2026-08-31
Add GE_41RT_newton to the distortion options

psavery:
Exposes the "_newton" GE_41RT variant in the detector distortion combo box, and treats it as a 6-parameter function like GE_41RT.

This is really necessary for some GE detectors that appear to exhibit pincushion effects, rather than barrel distortion effects. The new closed form method cannot handle the pincushion modeling.


saransh13 (approved):
LGTM

saransh13 (approved):
LGTM

psavery:
Wow, double-approved! :slightly_smiling_face:

saransh13:
a PR so nice, you have to approve it twice! 😄

PR #2050, opened 2026-09-04
Apply and show the space-group setting from the settings combo box

psavery:
The combo box offered the origin choices but the selection never reached material.sgsetting, so picking the second origin still built the structure in the first, and the widget was hidden because of it. Map the ":1" and ":2" suffixes onto the setting, select the matching entry when showing a material, and unhide the widget.

Depends on: hexrd/hexrd#965


psavery:
This will involve more extensive changes to the symmetry operations in hexrd, so putting this in a draft til we get those in.

PR #2051, opened 2026-09-10
Remove the JHE pinhole option

psavery:
Follows the hexrd change that deletes the JHE model: the correction is
now offered simply as Pinhole, and its JHE tab and widgets are gone.
State files written with either older name are migrated on load, with
JHE settings dropped because they carry no absorption length.

xref: https://github.com/HEXRD/hexrd/pull/968


saransh13 (approved):
LGTM

PR #2054, opened 2026-09-17
Preserve amorphous WPPF state

### LINKED ISSUE #2052: state file does not save amorphous peak information in wppf
When a state file is saved with a wppf analysis, and the wppf model included an amorphous peak, then on reload of the state file, the information is not saved. If amorphous peak is included again, then default values are used instead of the values from last analysis.

psavery:
Save the amorphous model setup so its fitted parameters reload.

Fixes: #2052


saransh13 (approved):
LGTM

